PR #6, opened 2017-06-24
Show a cancel button when adding a new account

BasThomas:
This adds a "cancel" button when adding new account. Without this, the user would get stuck when adding a new account and not being able to finish the flow.

### DIFF
--- a/Classes/Login/LoginViewController.swift
+++ b/Classes/Login/LoginViewController.swift
@@ -113,6 +113,15 @@ final class LoginViewController: UITableViewController, UITextFieldDelegate {
             navigationItem.rightBarButtonItem = nil
         }
     }
+    
+    override func accessibilityPerformEscape() -> Bool {
+        if isInitialLogin {
+            return false
+        } else {
+            handleResult(.cancelled, login: "")
+            return true
+        }
+    }
 
     // MARK: UITextFieldDelegate
 
--- a/Classes/Login/LoginViewController.swift
+++ b/Classes/Login/LoginViewController.swift
@@ -12,6 +12,7 @@ import OnePasswordExtension
 final class LoginViewController: UITableViewController, UITextFieldDelegate {
 
     var client: GithubClient!
+    var isInitialLogin: Bool = true
 
     let signinCellIndexPath = IndexPath(item: 0, section: 2)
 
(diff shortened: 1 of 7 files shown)


Classes/Login/LoginViewController.swift
```diff
@@ -53,6 +57,11 @@ final class LoginViewController: UITableViewController, UITextFieldDelegate {
             }
         }
     }
+    
+    @IBAction func onCancel(_ sender: Any) {
+        guard !isInitialLogin else { fatalError("Should not be able to cancel during initial login") }
+        handleResult(.cancelled, login: "")
```
BasThomas: Not sure about this - we don't use (or have a) "login" when cancelling, but just passing `""` feels wrong. Making it optional is a hassle, but might better reflect the possible state?

Classes/Login/LoginViewController.swift
```diff
@@ -12,12 +12,14 @@ import OnePasswordExtension
...
 
     var client: GithubClient!
+    var isInitialLogin: Bool = true
 
     let signinCellIndexPath = IndexPath(item: 0, section: 2)
 
     @IBOutlet weak var usernameTextField: UITextField!
     @IBOutlet weak var passwordTextField: UITextField!
     @IBOutlet weak var onepasswordButton: UIButton!
+    @IBOutlet var cancelButton: UIBarButtonItem!
```
BasThomas: This `@IBOutlet` is not `weak`, as it would be released when setting `rightBarButtonItem` to `nil`, thus not re-adding when calling `showCancelButton(true)`.

rnystrom:
This is awesome! Quick question, why not just dismiss the presented VC on cancel instead of mucking with login states and client systems?

BasThomas:
I did that because I found an issue when just doing a dismiss with `dismiss(animated: true)` on the VC. Because of [this](https://github.com/rnystrom/Freetime/blob/master/Classes/Systems/RootNavigationManager.swift#L34) line, just dismissing and then trying to reopen the login flow fails.

I am not really sure why you check this, though? If it is to not show multiple LoginVCs, we might want to check in another way?

rnystrom:
Ya, it's to prevent multiple logins when background and foregrounding the app. I use `applicationDidBecomeActive` to display the login VC b/c at that point the VC hierarchy is assembled/displayed (`didFinishLaunchingWithOptions` is too early).

This gave me an idea and I refactored how I'm doing settings entirely. Now I'm presenting a settings VC (contained in a nav) and _pushing_ the login flow, so you can just hit "Back" to exit. Good call on this.

I should add a warning to the README that I'm going to be changing stuff _very quickly_ so YMMV on whether I'll get to a PR in time... 😅

BasThomas:
Haha, no worries!

PR #32, opened 2017-06-29
Add clearer accessibility labels and traits

BasThomas:
What these changes do:

- Make the "more options" button in an issue / PR be called "More options" instead of "bullet"
- Make the "add reaction" button in a comment be called "Add reaction" instead of "plus"
- Add a button trait to both `SettingsUserCell` and `ButtonCell` to indicate these are buttons
- Make the user cell in settings make clear this is an account, and tell us when it is selected (or not)
- Make the SettingsVC dismissible with a two-finger Z-swipe in VoiceOver mode.
- Make the "settings" button be called "Settings" instead of "bullets hollow"
- Remove the button trait from buttons used as labels (eg. the "merged" / "open" / "closed" images)


rnystrom:
omfg, amazing

PR #82, opened 2017-07-08
Disable mark all button if all visible notifications are read

gtchance:
https://github.com/rnystrom/Freetime/issues/44

I’m interested in contributing to the project. Just trying to get my feet wet with this one.

### DIFF
--- a/Classes/Notifications/NotificationsViewController.swift
+++ b/Classes/Notifications/NotificationsViewController.swift
@@ -67,13 +67,7 @@ NotificationNextPageSectionControllerDelegate {
     }
     
     private func updateMarkAllEnabled() {
-        var allRead = true
-        for notification in filteredNotifications {
-            if !notification.read {
-                allRead = false
-                break
-            }
-        }
+        let allRead = !filteredNotifications.contains(where: { $0.read == false })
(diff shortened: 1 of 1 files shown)


Classes/Notifications/NotificationsViewController.swift
```diff
@@ -63,6 +63,18 @@ NotificationNextPageSectionControllerDelegate {
             target: self,
             action: #selector(NotificationsViewController.onMarkAll(sender:))
         )
+        updateMarkAllEnabled()
+    }
+    
+    private func updateMarkAllEnabled() {
+        var allRead = true
+        for notification in filteredNotifications {
```
Sherlouk: Maybe a bit simpler to just do a `.contains` on `filteredNotifications`. Example:

```
let allRead = !filteredNotifications.contains(where: { $0.read == false })
```

(Not sure if that logic is the right way round, but for example it'll do!

gtchance: You're right, that's much cleaner. I made the change.

rnystrom:
This is great, thanks for helping out @gtchance!

PR #93, opened 2017-07-10
[minor] Update alert text

dkhamsing:
I found the "Mark All" alert a bit strange.. this pull requests changes that to "OK" instead

From https://developer.apple.com/ios/human-interface-guidelines/ui-views/alerts/
> Give alert buttons succinct, logical titles. The best button titles consist of one or two words that describe the result of selecting the button. As with all button titles, use title-style capitalization and no ending punctuation. To the extent possible, use verbs and verb phrases that relate directly to the alert title and message—for example, View All, Reply, or Ignore. Use OK for simple acceptance. 

Also this changes "Ok" to "OK" in other places

Thx for the app!

### DIFF
--- a/Classes/Notifications/NotificationsViewController.swift
+++ b/Classes/Notifications/NotificationsViewController.swift
@@ -91,7 +91,7 @@ NotificationNextPageSectionControllerDelegate {
             preferredStyle: .alert
         )
         let markAll = UIAlertAction(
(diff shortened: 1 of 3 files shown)


dkhamsing:
Would you accept "Mark All Read" ?

rnystrom:
@dkhamsing sure! I like that.

dkhamsing:
@rnystrom updated

PR #101, opened 2017-07-10
Update to use Apple's ISO8601 Date Formatter

### LINKED ISSUE #22: Bug: Relative time is incorrect
iOS 11 beta 2; 6S+

The relative time on actual GitHub says 5 hours ago, and the one in the app says 17 hours ago!

One of them must be wrong 🤔

### LINKED ISSUE #64: Half of issue is missing
Weird issue but heres the article I'm trying to view:
https://github.com/rnystrom/Freetime/issues/47

That issue was opened by Bas, and I commented in between Ryan's posts - yet in app I see this:

https://www.dropbox.com/s/96ftt2y5z35m1co/Photo%2006-07-2017%2C%2010%2055%2005%20am.png?dl=0

iOS 11 beta 2

Sherlouk:
Fixes #64
I also think this may fix #22

So originally I was just going to add the locale and timeZone properties as per Apple's documentation [here](https://developer.apple.com/documentation/foundation/dateformatter#2528261). But then I realised we only support iOS 10+, and that gives us access to the new [`ISO8601DateFormatter`](https://developer.apple.com/documentation/foundation/iso8601dateformatter)

I gave the app a cursory glance in both 12hr and 24hr modes and it does appear to be working?

The alternative (if we wanted iOS 9 support, or if someone finds further issues with this) is, for reference:
``` swift
dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ssZ"
dateFormatter.locale = Locale(identifier: "en_US_POSIX")
dateFormatter.timeZone = TimeZone(secondsFromGMT: 0)
dateFormatter.calendar = Calendar(identifier: .iso8601)
```

### DIFF
--- a/Classes/Systems/GithubAPIDateFormatter.swift
+++ b/Classes/Systems/GithubAPIDateFormatter.swift
@@ -8,9 +8,8 @@
 
 import Foundation
 
-private let dateFormatter = DateFormatter()
-func GithubAPIDateFormatter() -> DateFormatter {
+private let dateFormatter = ISO8601DateFormatter()
+func GithubAPIDateFormatter() -> ISO8601DateFormatter {
     // https://developer.github.com/v3/#schema
-    dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ssZ"
     return dateFormatter
 }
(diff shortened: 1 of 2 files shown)


rnystrom:
Yaaaay!

Mouvedia:
@Sherlouk which app would you recommended for someone stuck on iOS 9?

PR #109, opened 2017-07-11
Fixed retain cycle related to AttributedStringView

gtchance:
Most notably caused IssuesViewController to be retained after
dismissing.


rnystrom:
😱

Whoops! Thank you!!

Issue #118, opened 2017-07-12
Improve accessibility


BasThomas:
What can still be improved:

- [ ] Issue contents (text, images, tables etc.) are not accessible at all (!!!)
- [ ] In an issue, the username is tappable (at the top of a comment cell), but is not specified as a button
- [ ] "Labels" cell is a button, but is not specified as such
- [ ] Expanded labels are also (collapsing) buttons, but are not specified as such
- [ ] The tappable username should have an accessibility hint (what does it do? -> open profile in Safari)
- [ ] In "All", it is not specified if an issue / PR is unread or not
- [ ] Issue / PR status only says its status (eg: "open"), not the fact that it is a status
- [ ] Reactions are accessible in reversed order (?): first reactions, then "Add reactions"
- [ ] Reactions are not specified as buttons
- [ ] Reactions should have an accessibility hint explaining that tapping them +1s the reaction
- [ ] Specify if a comment is self-authored
- [ ] Issue statuses should be read as a whole (now accessible per "part", such as username, label, time) -> This means we might want to remove the option to tap a username to open it in SafariVC?

Sherlouk:
@BasThomas Not sure if this issue is intentionally all VoiceOver, or can cover other accessibility things? But also think we should tackle

- [ ] Dynamic type (Font Sizes)
- [ ] Reduce view controller animations if "reduce motion" is enabled
- [ ] [Prevent inverting](https://developer.apple.com/documentation/uikit/uiview/2865843-accessibilityignoresinvertcolors) some images (such as profile images)

Sherlouk:
Another thing is the notification cells currently read out each of the subviews (which is definitely an improvement!) but maybe we could make it a bit more conversational and provide a bit more information? An example:

`rnystrom Freetime, {Issue/Pull Request} Number 118, Improve Accessibility, Last Updated {an hour ago}`

As opposed to:
`rnystrom/Freetime #118, an hour ago, Improve Accessibility, Issue`

BasThomas:
Of course accessibility is more than just VoiceOver, so definitely!

On improved labels: we can do that, but that also means it is easier to break (eg. not updating the label when something changes in the future.

BasThomas:
Different tab icons for selected / deselected state

rnystrom:
YES, lets do it. Should be easy.

BasThomas:
https://github.com/rnystrom/GitHawk/issues/279#issuecomment-326995965

[...]

PR #119, opened 2017-07-12
Improve accessibility

BasThomas:
Ref #118 

This PR:

- [x] Makes the More Options button in the issue overview say "More Options" instead of "bullets-hollow"
- [x] Makes the notification cell accessible as a whole, as well as mentioning its type (issue, pr or commit)


Classes/Notifications/NotificationCell.swift
```diff
@@ -98,6 +100,10 @@ final class NotificationCell: SwipeSelectableCell {
         textLabel.attributedText = viewModel.title.attributedText
         dateLabel.setText(date: viewModel.date)
         reasonImageView.image = viewModel.type.icon?.withRenderingMode(.alwaysTemplate)
+        accessibilityLabel = contentView.subviews
+            .flatMap { $0.accessibilityLabel }
+            .reduce("", { $0 + ".\n" + $1 })
```
BasThomas: the `./n`-s makes VoiceOver read these as sentences instead of just one, long sentence, making the different labels / elements separated.

Classes/Views/SelectableCell.swift
```diff
@@ -47,6 +57,9 @@ class SelectableCell: UICollectionViewCell {
 
     override func layoutSubviews() {
         super.layoutSubviews()
+        accessibilityLabel = contentView.subviews
+            .flatMap { $0.accessibilityLabel }
+            .reduce("", { "\($0 ?? "").\n\($1)" })
```
rnystrom: Mind moving this to an override of `accessibilityLabel`? Then it only executes when necessary instead of every layout pass.

Classes/Notifications/NotificationCell.swift
rnystrom: Same note about moving this to override?

rnystrom:
SO GOOD, I love it

Classes/Views/SelectableCell.swift
BasThomas: That would be a little restricting though - we can't access the `viewModel` in the `NotificationCell`, which I used to provide the type (Issue / PR / commit).

Classes/Notifications/NotificationCell.swift
rnystrom: What about this one?

BasThomas: See the comment on the other one:

> That would be a little restricting though - we can't access the `viewModel` in the `NotificationCell`, which I used to provide the type (Issue / PR / commit).

PR #134, opened 2017-07-15
Implement login with personal access token

sberrevoets:
This implements the functionality to log in with a personal access token. I'm not entirely sure what you think the UI should look like for this; for now I've just added a text field on the login screen. When the field is not empty, the contents will be assumed to be the PAT and the OAuth flow will be skipped in that case.

simulator screen shot jul 14 2017 11 37 43 pm

Errors aren't handled, as I'm not sure how that should be surfaced. I tested this functionality on my own account and it worked after creating a token for myself.

Fixes: https://github.com/rnystrom/Freetime/issues/111


rnystrom:
I think I messed something up by rewriting history to purge some of my client keys from the repo. My b.

I really want to do this, but design wise would put it in another VC. I think I'll close this and take your work but just move it to another VC. Sound good?

sberrevoets:
Yep, works for me. If you won't be able to get around to it soon and have pointers as to what you'd want from a UI like that instead, let me know and I can look into that as well. Currently this is the only thing that prevents me from actually using the app, so looking forward to being able to use this 🙌

rnystromtest:
bump for testing

rnystrom:
@sberrevoets if you want to take this on still, I'd love it. Mucked around w/ some design thoughts and came up w/ something simple:

- A `UIButton` with title "Personal Token" in the top-right corner of the login VC
- Tapping button displays a `UIAlertController` w/ text field
- And let's do error handling

That's basically it. What do you think?

sberrevoets:
Sounds good to me! The only problem I foresee is error handling. Since the `UIAlertController` API is pretty limited in what it can do (no animation, no text colors, etc.), what do you think should happen when the user taps "Authorize" (or whatever) on the alert but pasted an invalid/wrong token?

We can make the alert disappear while we do a network request to validate the token, but then what do we do while that network request is in-flight? We can keep the alert up and disable the buttons, but really the only thing you can do then is update the alert's message, which isn't very clear.

Any thoughts/opinions?

rnystrom:
@sberrevoets how about we dismiss the alert and show the same [loading spinner](https://github.com/rnystrom/Freetime/blob/master/Classes/Login/LoginSplashViewController.swift#L32) while logging in. On failure, just call [handleError](https://github.com/rnystrom/Freetime/blob/master/Classes/Login/LoginSplashViewController.swift#L65-L75). That way the experience is exactly the same as oauth flow, and changes are extremely minimal.

sberrevoets:
Awesome! I'll take a look at that this weekend.

sberrevoets:
Found some time tonight actually 🙂 https://github.com/rnystrom/Freetime/pull/194

PR #135, opened 2017-07-15
Add version number to settings page also

### LINKED ISSUE #130: Show app version info in settings
Would be awesome to prefill that in bug reports, but not sure if that's possible.

Sherlouk:
Closes #130 

Includes some minor tweaks to cell spacing to make it look nice, and a new LabelCell

Moved the version string stuff into it's own extension so it could be reused (and added localization to it because Version in Spanish is Versión and they are totally different words!)

Looks like this:
Ooh fancy

### DIFF
--- a/Classes/Settings/SettingsVersionInfoSectionController.swift
+++ b/Classes/Settings/SettingsVersionInfoSectionController.swift
@@ -1,0 +1,33 @@
+//
+//  SettingsVersionInfoSectionController.swift
+//  Freetime
+//
+//  Created by Sherlock, James on 15/07/2017.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import UIKit
+import IGListKit
+
+final class SettingsVersionInfoSectionController: ListSectionController {
+    
+    override init() {
+        super.init()
+        inset = UIEdgeInsets(top: 0, left: 0, bottom: Styles.Sizes.tableSectionSpacing, right: 0)
+    }
+    
+    override func sizeForItem(at index: Int) -> CGSize {
+        guard let width = collectionContext?.containerSize.width else { fatalError("Collection context must be set") }
+        return CGSize(width: width, height: Styles.Sizes.tableCellHeight)
+    }
+    
+    override func cellForItem(at index: Int) -> UICollectionViewCell {
+        guard let cell = collectionContext?.dequeueReusableCell(of: LabelCell.self, for: self, at: index) as? LabelCell
+            else { fatalError("Collection context must be set or cell incorrect type") }
+        cell.label.text = Bundle.main.prettyVersionString
+        return cell
+    }
+    
+    override func didSelectItem(at index: Int) {}
+    
+}
--- a/Classes/Settings/SettingsVersionInfoSectionController.swift
+++ b/Classes/Settings/SettingsVersionInfoSectionController.swift
@@ -13,7 +13,7 @@ final class SettingsVersionInfoSectionController: ListSectionController {
     
     override init() {
         super.init()
-        inset = UIEdgeInsets(top: 0, left: 0, bottom: Styles.Sizes.tableSectionSpacing, right: 0)
+        inset = .zero
     }
     
     override func sizeForItem(at index: Int) -> CGSize {
@@ -27,7 +27,5 @@ final class SettingsVersionInfoSectionController: ListSectionController {
         cell.label.text = Bundle.main.prettyVersionString
         return cell
     }
-    
-    override func didSelectItem(at index: Int) {}
     
 }
--- a/Classes/Utility/Bundle+Version.swift
+++ b/Classes/Utility/Bundle+Version.swift
@@ -19,7 +19,7 @@ extension Bundle {
     }
     
     var prettyVersionString: String {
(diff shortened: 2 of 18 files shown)


Classes/Views/LabelCell.swift
```diff
@@ -0,0 +1,37 @@
...
+
+final class LabelCell: UICollectionViewCell {
+    
+    let label = UILabel()
+    
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        
+        contentView.backgroundColor = .clear
+        accessibilityTraits |= UIAccessibilityTraitStaticText
```
BasThomas: 😍

Sherlouk: Admittedly wouldn't have thought to do it if it wasn't for the fact I copied it from the button cell 😅 😅

BasThomas: That works too, right? :)

Classes/Utility/Bundle+Version.swift
```diff
@@ -0,0 +1,25 @@
...
+    
+    var buildNumber: String? {
+        return infoDictionary?["CFBundleVersion"] as? String
+    }
+    
+    var prettyVersionString: String {
+        return NSLocalizedString("Version", comment: "") + " \(versionNumber ?? "Unknown") (\(buildNumber ?? "0"))"
+    }
+    
+}
```
rnystrom: Lovely :)

Classes/Utility/Bundle+Version.swift
```diff
@@ -0,0 +1,25 @@
...
+    var versionNumber: String? {
+        return infoDictionary?["CFBundleShortVersionString"] as? String
+    }
+    
+    var buildNumber: String? {
+        return infoDictionary?["CFBundleVersion"] as? String
+    }
+    
+    var prettyVersionString: String {
+        return NSLocalizedString("Version", comment: "") + " \(versionNumber ?? "Unknown") (\(buildNumber ?? "0"))"
```
rnystrom: Let's use `Strings.unknown` so its using common strings + localized

Classes/Views/LabelCell.swift
```diff
@@ -0,0 +1,37 @@
...
+//  Created by Sherlock, James on 15/07/2017.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import UIKit
+import SnapKit
+
+final class LabelCell: UICollectionViewCell {
+    
+    let label = UILabel()
```
rnystrom: `private`

Classes/Views/Styles.swift
```diff
@@ -19,7 +19,7 @@ enum Styles {
         static let rowSpacing: CGFloat = 8
         static let cellSpacing: CGFloat = 15
         static let tableCellHeight: CGFloat = 44
-        static let tableSectionSpacing: CGFloat = 35
+        static let tableSectionSpacing: CGFloat = 17
```
rnystrom: This also effects the "no notifications" (:tada:) view. Double check that doesn't look too tight?

Classes/Settings/SettingsVersionInfoSectionController.swift
```diff
@@ -0,0 +1,33 @@
...
+//
+
+import UIKit
+import IGListKit
+
+final class SettingsVersionInfoSectionController: ListSectionController {
+    
+    override init() {
+        super.init()
+        inset = UIEdgeInsets(top: 0, left: 0, bottom: Styles.Sizes.tableSectionSpacing, right: 0)
```
rnystrom: All the others have the inset at the `top`, why bottom here?

Classes/Settings/SettingsVersionInfoSectionController.swift
```diff
@@ -0,0 +1,33 @@
...
+    }
+    
+    override func cellForItem(at index: Int) -> UICollectionViewCell {
+        guard let cell = collectionContext?.dequeueReusableCell(of: LabelCell.self, for: self, at: index) as? LabelCell
+            else { fatalError("Collection context must be set or cell incorrect type") }
+        cell.label.text = Bundle.main.prettyVersionString
+        return cell
+    }
+    
+    override func didSelectItem(at index: Int) {}
```
rnystrom: can remove

Classes/Settings/SettingsVersionInfoSectionController.swift
Sherlouk: Bottom is unnecessary here, the lack of top is to make it closer to the log out button (hence the changes in the first place)

With the original spacing it was too far away which looked weird

Classes/Utility/Bundle+Version.swift
Sherlouk: Done!

Classes/Views/LabelCell.swift
Sherlouk: Can't have this as private as need access to it to set it's text value

Classes/Views/Styles.swift
Sherlouk: Reverted back to old value, just changed the bottom value on Log Out to give the same effect

PR #138, opened 2017-07-15
Add list of reactors on long press

### LINKED ISSUE #66: Long press to see who reacted with an emoji
img_1094

Sherlouk:
Closes #66

Took me way too long to work out how GQL works 😅 Yes, I changed API.swift multiple times and was confused why it was being deleted until I read the comment where it says it's auto generated.

Supports multiple outputs, as such:
"x reacted with heart emoji"
"x and y reacted with heart emoji"
"x, y and z reacted with heart emoji"
"x, y, z and 1 other reacted with heart emoji"
"x, y, z and 15 others reacted with heart emoji"

There's a limit of 10 names before it adds the "and x other(s)" logic, this was plucked out of the air and is simply to prevent the list from getting huge!

You get something like this:
Preview


Classes/Issues/Comments/Reactions/ReactionContent+ReactionType.swift
```diff
@@ -19,4 +19,16 @@ extension ReactionContent {
         case .thumbsDown: return "👎"
         }
     }
+    
+    /// User readable name of the current reaction type
+    var name: String {
+        switch self {
+        case .confused: return "confused"
+        case .heart: return "heart"
+        case .hooray: return "horray"
```
BasThomas: Typo, `hooray`.

Not sure if this is needed also, but we might want to wrap these in `NSLocalizedString`. Also, I've used `var localizedString` in the `IssueType` (instead of `name`), but I guess that's NBD.

Classes/Issues/IssueViewModels.swift
```diff
@@ -62,7 +62,17 @@ func createIssueReactions(reactions: ReactionFields) -> IssueCommentReactionView
         // do not display reactions for 0 count
         let count = group.users.totalCount
         guard count > 0 else { continue }
-        models.append(ReactionViewModel(content: group.content, count: count, viewerDidReact: group.viewerHasReacted))
+        let users = group.users.nodes!.map({ $0?.login })
```
BasThomas: Is this used anywhere?

Classes/Issues/Comments/Reactions/ReactionContent+ReactionType.swift
Sherlouk: oops! Fixing now! I'll also change the variable name to be somewhat consistent

Classes/Issues/IssueViewModels.swift
Sherlouk: Er... 😅 Nope!

Sherlouk: Good spot, fixed

Classes/Issues/Comments/Reactions/IssueCommentReactionCell.swift
```diff
@@ -156,6 +156,23 @@ UICollectionViewDelegateFlowLayout {
...
+        let difference = model.count - users.count
+        if difference > 0 {
+            let plural = difference > 1 ? "s" : ""
+            users.append("\(difference) other\(plural)")
+        }
+        
+        let lastUser = users.removeLast()
+        var message = users.joined(separator: ", ")
+        message += users.count > 0 ? " and " + lastUser : lastUser
+        message += " reacted with " + model.content.localizedString + " emoji"
```
rnystrom: A few things: 

- these aren't localized strings 
- we need to use localized formatting strings and not con at (other languages comma separate differently)

So probably need to check which case it is (ither count or not) and then use a formatting string like `NSLocalizedString("%@ and %i others", "").

Also let's drop the "with heart", the tooltip already displays above the emoji, should be obvious.

Classes/Issues/Comments/Reactions/IssueCommentReactionCell.swift
```diff
@@ -156,6 +156,23 @@ UICollectionViewDelegateFlowLayout {
...
         cell.label.text = "\(model.content.emoji) \(model.count)"
         cell.contentView.backgroundColor = model.viewerDidReact ? Styles.Colors.Blue.light.color : .clear
+        
+        var users = model.users
+        guard users.count > 0 else { return cell }
+        
+        let difference = model.count - users.count
+        if difference > 0 {
+            let plural = difference > 1 ? "s" : ""
+            users.append("\(difference) other\(plural)")
```
rnystrom: Check out how I'm doing localized plural with date formats. We should do the same here.

gql/Fragments.graphql
```diff
@@ -2,7 +2,10 @@ fragment reactionFields on Reactable {
   viewerCanReact
   reactionGroups {
     viewerHasReacted
-    users {
+    users(first: 10) {
```
rnystrom: 10 seems a bit high, no? I like your example using up to 3 before showing "n others".

Classes/Issues/Comments/Reactions/ReactionContent+ReactionType.swift
```diff
@@ -19,4 +19,16 @@ extension ReactionContent {
...
+    var localizedString: String {
+        switch self {
+        case .confused: return NSLocalizedString("confused", comment: "")
+        case .heart: return NSLocalizedString("heart", comment: "")
+        case .hooray: return NSLocalizedString("hooray", comment: "")
+        case .laugh: return NSLocalizedString("laugh", comment: "")
+        case .thumbsUp: return NSLocalizedString("thumbs up", comment: "")
+        case .thumbsDown: return NSLocalizedString("thumbs down", comment: "")
+        }
+    }
```
rnystrom: See note about dropping the emoji name from the tooltip, we can probably get rid of these.

gql/Fragments.graphql
Sherlouk: Done!

Classes/Issues/Comments/Reactions/ReactionContent+ReactionType.swift
Sherlouk: Done!

Classes/Issues/Comments/Reactions/IssueCommentReactionCell.swift
Sherlouk: Dropped the emoji name, will work on the localization formatting (never done it before!)

Classes/Issues/Comments/Reactions/IssueCommentReactionCell.swift
Sherlouk: Done!

Classes/Issues/Comments/Reactions/IssueCommentReactionCell.swift
Sherlouk: Got the "x others" pluralisation into localisation, but can't work out how to turn an array of names into a localised comma separated string? Any tips?

Sherlouk:
If anyone knows how, could do with some assistance with localizing an array of names into a single string like:
"name1, name2, name3 and name4"

rnystrom:
Have some ideas, will write them up later. The gist is I think we need to create localized format strings for all permutations.

rnystrom:
So I think we should make the following localizable format strings:

- `"%@ and %@"`
- `"%@, %@, and %@"`
- `"%@, %@, and %@ and %d other(s)"`

Then add plural localization strings [like these](https://github.com/rnystrom/Freetime/blob/master/Resources/Localizable.stringsdict#L5-L27) which ends up being used [like this](https://github.com/rnystrom/Freetime/blob/8a15908c9ec15e0489953c41f01b632344dd4f5c/Classes/Views/Date%2BDisplay.swift#L57-L58).

rnystrom:
Sorry for the rebase nightmare, I rewrote history to purge some client keys. Either need to create a new PR or rebase. Lmk what you want to do here!

Sherlouk:
I'll give this a go ASAP, was hoping there would be a better solution than literally a new localised string for every permutation!

Sherlouk:
Moved to #169

PR #155, opened 2017-07-16
Adds ability to scroll to bottom in Feed finishLoading

### LINKED ISSUE #85: Scroll to bottom by default
When reading through an issue, I think it would make sense to scroll the user to the last comment, rather than the first one.

Cheers!

flovilmart:
Closes #85

The issue controller actually doesn't support paging (for the moment, AFAICT), so scrolling to the last element will probably always be more comfortable / useful than staying at the top.

### DIFF
--- a/Classes/Issues/IssuesViewController.swift
+++ b/Classes/Issues/IssuesViewController.swift
@@ -225,7 +225,7 @@ FeedSelectionProviding {
             }
 
             self.models = results
-            self.feed.finishLoading(dismissRefresh: true)
+            self.feed.finishLoading(dismissRefresh: true, scrollToLastItem: true)
         }
     }
 
--- a/Classes/Systems/Feed.swift
+++ b/Classes/Systems/Feed.swift
@@ -71,12 +71,15 @@ final class Feed: NSObject, UIScrollViewDelegate {
         }
     }
 
-    func finishLoading(dismissRefresh: Bool, animated: Bool = true) {
+    func finishLoading(dismissRefresh: Bool, scrollToLastItem: Bool = false, animated: Bool = true) {
         status = .idle
         let block = {
             self.adapter.performUpdates(animated: animated) { _ in
                 if dismissRefresh {
                     self.collectionView.refreshControl?.endRefreshing()
+                }
+                if scrollToLastItem {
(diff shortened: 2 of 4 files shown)


flovilmart:
@rnystrom @BasThomas ACAICT even on long threads this seems to be OK.

Classes/Utility/UICollectionView+Utils.swift
```diff
@@ -0,0 +1,33 @@
...
+        return IndexPath(item: lastItem, section: lastSection)
+    }
+
+    func scrollToLastItem(at position: UICollectionViewScrollPosition = .top, animated: Bool) {
+        guard let indexPath = indexPathOfLastItem() else {
+            // empty collectionView
+            return
+        }
+        scrollToItem(at: indexPath, at: position, animated: false)
+    }
```
rnystrom: Quick question, why not just change the content offset based on the size+inset?

https://github.com/rnystrom/Freetime/blob/master/Classes/Issues/IssuesViewController.swift#L212-L214

flovilmart: That could work too :) tried with scrollToRectVisble but it didn’t work as expected

Classes/Utility/UICollectionView+Utils.swift
```diff
@@ -0,0 +1,17 @@
...
+
+import UIKit
+
+extension UICollectionView {
+    func scrollToBottom(animated: Bool) {
+        var offset = contentOffset
+        offset.y = contentSize.height + contentInset.bottom - bounds.height
+        setContentOffset(offset, animated: animated)
+    }
+}
```
rnystrom: Hahaha, funny, I added literally the exact same extension on my own today.

rnystrom:
The `IssuesViewController` class changed a lot upstream, might want to rebase and try this again. I can just take these changes and apply them to the branch manually too instead of trying to fix it up. Lmk.

flovilmart:
Just cherry picked the changes. lmk

rnystrom:
Took this and fixed merges, also executes after the refresh control finishes so the animations don't collide. Thanks @flovilmart!

PR #169, opened 2017-07-20
Add list of reactors on long press

### LINKED ISSUE #66: Long press to see who reacted with an emoji

Sherlouk:
Closes #66

Took me way too long to work out how GQL works 😅 Yes, I changed API.swift multiple times and was confused why it was being deleted until I read the comment where it says it's auto generated.

Supports multiple outputs, as such:
"x reacted"
"x and y reacted"
"x, y and z reacted"
"x, y, z and 1 other reacted"
"x, y, z and 15 others reacted"

There's a limit of 3 names before it adds the "and x other(s)" logic

Each permutation is individually localized

You get something like this:
Preview


Sherlouk:
One query I did have, this get's the first 3 people to react to the post. Should it be the last 3 (most recent)?

rnystrom:
> Should it be the last 3 (most recent)?

Hmmmmm, good question. Probably nobody will notice this detail. But I say keep w/ the first 3, then it's not changing constantly. Let's roll w/ this and see how it goes.

Btw you're a 🚀 ⭐️  for pumping this out!

PR #176, opened 2017-07-23
Updated `SplitViewControllerDelegate` to display placeholder under correct conditions.

shaps80:
Introduced FeedSelectionProviding to indicate when a selection has been made
Updated IssuesViewController and NotificationsViewController to conform to FeedSelectionProviding
Updated NotificationsViewController to show the placeholder when toggling the filter
Updated the SplitViewDelegate to depend on the current selection rather than a placeholder view


rnystrom:
Gonna merge + do a quick cleanup pass to get formatting consistent (e.g. `// MARK:` for protocols)

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -119,6 +123,8 @@ NotificationNextPageSectionControllerDelegate {
         )
         feed.finishLoading(dismissRefresh: dismissRefresh, animated: animated)
         updateMarkAllEnabled()
+        
+        showDetailViewController(SplitPlaceholderViewController(), sender: self)
```
rnystrom: I was suspect of this, going to remove since it pushes a placeholder VC on iPhone when starting in portrait.

shaps80: Ohh true.. My bad.
To be honest the implementation you've done with placeholder isn't quite right so I was trying to work within your constraints.

Normally your detail view controller is defined in the storyboard and you use the splitViewDelegate to override or wateva for things like an empty controller.

Sorry about that.

rnystrom: Np! easy fix

PR #184, opened 2017-07-25
Replace manual emoji map with JSON

### LINKED ISSUE #181: Autocomplete Emoji Alternative Names
Always a minor annoyance of mine when I can't remember the exact name for an emoji (so autocomplete won't pick it up!)

Who knew 🎉  was `tada` 🤔 

Maybe revise to use something like this JSON file: https://github.com/github/gemoji/blob/master/db/emoji.json (from GitHub, so most likely to be up-to-date) and autocomplete on aliases and tags?

Sherlouk:
Original JSON comes from: https://raw.githubusercontent.com/github/gemoji/master/db/emoji.json

Closes #181 

Had contemplating adding support for the "ios_version" in the JSON, but thought it probably wasn't worth it!
Also there's a couple emojis in the JSON which don't have an actual emoji tied to them. These are ones that GitHub add but are not in unicode so they render as an image on the site instead. Example: :rage1: :rage2: :rage3: :trollface: (Non of those will render in-app, nor will they show up in autocomplete)

Whenever we need to update the emoji list, we can just replace the copy in the repo with the one from the url above


Classes/Systems/Autocomplete/EmojiAutocomplete.swift
```diff
@@ -39,15 +39,16 @@ final class EmojiAutocomplete: AutocompleteType {
...
-        }
-
+        let results: [Result] = GithubEmojis.flatMap({ emoji in
+            // Iterate through each of the names, and find the first one which is prefixed with the search term
+            guard let matchingName = emoji.names.first(where: { emojiName in
+                emojiName.lowercased().hasPrefix(lowerword)
+            }) else { return nil }
+            
+            return Result(emoji: emoji.emoji, term: ":\(matchingName):")
+        })
```
rnystrom: Nice, I'm going to land this then do a quick refactor to speed this up (n^2 atm)

Sherlouk: Mm yeah sorry I really need to learn up on function complexity/speed like that!

PR #197, opened 2017-07-28
Adds XCConfig support to remove client id and secret from git.

shaps80:
The Contributors.md file explains in more detail how this works. 

Basically there is a new build phase script that generates a default XCConfig file where the developer can put their Client ID and Secret. This file is NOT committed to the repo so there's no risk of accidentally sharing your details (or needing to re-write history @rnystrom 😬😂)


rnystrom:
Hahaha I just broke my beta submission because of this too, was literally reading up on ways to do this. Awesome!

BasThomas:
Awesome! Can you remove `Setup.md` and link to the new file in the Readme?

shaps80:
@BasThomas great idea.

shaps80:
README updated!

BasThomas:
Can you remove `Setup.md` too?

shaps80:
Sorry I forgot that – done!

rnystrom:
@shaps80 I had to revert this b/c it ended up breaking my local build:

- I got errors about missing config files. Followed your steps to clean+build but the files never generated
- Then I created both files manually, changing debug/release:

```
#include "Pods/Target Support Files/Pods-Freetime/Pods-Freetime.debug.xcconfig"

CLIENT_ID="abc"
CLIENT_SECRET="abc123"

GLOBAL_CONSTANTS = CLIENT_ID=\$(CLIENT_ID) CLIENT_SECRET=\$(CLIENT_SECRET)
GCC_PREPROCESSOR_DEFINITIONS = \$(inherited) \$(GLOBAL_CONSTANTS)
```

But then the ObjC files failed to build, complaining about missing variables. A couple notes:

- Is there something that I missed? I figured this would "just work"?
- Can the secret/id be set in a single file instead of both the debug/release? Single source of truth, avoids any catastrophic issue of accidentally sending stale secrets to prod.

dakeshi:
@rnystrom how was to use https://github.com/orta/cocoapods-keys/ to handle keys?

# step to setup for cocoapods-keys
1. gem install cocoapods-keys
2. Add the following code in Podfile
```ruby
# setup for cocoapods-keys
plugin 'cocoapods-keys', {
  :project => "Freetime",
  :keys => [
  "GithubAPIClientID",
  "GithubAPIClientSecret"
]}
```
3. Edit Secrets.swift

```swift
import Keys
fileprivate let keys = FreetimeKeys()

public enum GithubAPI {

    static let clientID = keys.githubAPIClientID
    static let clientSecret = keys.githubAPIClientSecret

}
```

shaps80:
## Build Phase

@rnystrom Interesting I've had no issues with it once I got it all into this PR. Did @AnyoneElse have issues with it?

One thing to check would be the build phase and make sure "Run script only when installing" is **NOT** checked. Also check the console output to make sure the phase is being run.

It definitely should JUST work.

## Single File

You could absolutely update script to not check for `DEBUG` specifically and just use the same file for both. Will also need to update the project settings to reflect that obviously. I set it up with 2 because you had previously mentioned wanting to use a test account when debugging, which this would allow. 

But its an easy change.

## Cocoapods

@dakeshi I've not seen this before so I can't really comment on it. Seems like an interesting approach though. I quite like it and perhaps its simpler – requiring less changes to build phases, project files, etc...

@rnystrom ? Thoughts?

rnystrom:
Ok cool, I'll dig into both solutions a bit and try to repro the bugs I ran into. I should have documented them better. FWIW I **definitely** want to move to this solution vs the `Secrets` file!

dakeshi:
@shaps80 I have got the error in the build time. It is related to `[CP] Check Pods Manifest lock` script. So I run `pod install`-> `clean build` -> `build` again. But the build is failed with same phrase. In the build log, I can see `Generate Xcconfig file` script did not have a chance to run. 

Um, so I did change the order of the script between `[CP] Check Pods Manifest lock`and `Generate Xcconfig file` in the build phrase. (I don't know this is a good solution because the order become reset when the project is opened). Anyway, I build it and it works!

But the most interesting part is that I get build success when I test it again with clean status based on this repo. 😕

rnystrom:
@dakeshi decided against using cocoapods-keys b/c I'd like to avoid anyone having to install dependencies to start using this app locally.

I think I'm going to try to redo @shaps80's work w/ just a single config file.

shaps80:
Sounds good. Maybe check the order of the build phase (as per @dakeshi's suggestion) because if things don't run in the right order, you would have issues where the file doesn't exist. 

Otherwise I agree. Preferable approach IMHO.

dakeshi:
@rnystrom I agree too. If we solve build issues, I prefer to use @shaps80's suggestion to avoid dependency problems.

PR #200, opened 2017-07-29
Search for repositories

Sherlouk:
Awaiting Initial Review

This branch is aiming to create the ability to search for repositories by name, and then view the issues and pull requests of the selected repo.

[Search Page Screenshot](https://user-images.githubusercontent.com/15193942/28745356-c63a5038-746e-11e7-9633-12127aa25699.png)

Spin up issues:
- Filtering/Ordering search results and repo issues/prs
- Further accessibility improvements


rnystrom:
That looks SO slick!

Sherlouk:
[Pull Requests Screenshot](http://prntscr.com/g1uq7s)

Page can now switch between issues/pull requests, icons now update depending on state (color and image)

Having few issues with paging but will get that sorted next

Sherlouk:
Also gonna do a big refactor to make the notification view model a bit more generic and share between notifications and repository view - same approach for a few things.

Sherlouk:
Turns out Apollo doesn't have proper support for GraphQL directives which makes this super annoying 😞 

Rather than one query which can return issues, pull requests or both. Got to have two separate queries, bit messy at the moment but once I make the generic interface I can tidy it up a bunch

Sherlouk:
Issue and Pull Requests can both now be paged - When you first load the first it will load 25 issues and 25 pull requests. If more is available then a "Load More" appears at the bottom. Loading more will load another 25 of either issues or pull requests depending on which one you're looking at.

Doing a pull to refresh will reload everything (issues and pull requests, resetting back to 25 of each).

If this sounds stupid then someone please tell me 😅

Sherlouk:
Tab bar controller is in!

[Portrait](http://prntscr.com/g25yhn)
[Landscape](https://user-images.githubusercontent.com/15193942/28755061-4a42b1a0-754a-11e7-9386-9bcaa5fcbcbd.png)

Reworked the navigation a little so the search results appear in the side pane of the split view controller, as does the repo view - Only the issue itself loads in the right panel.

Sherlouk:
@rnystrom How do you feel about moving Settings from Notifications '...' to it's own tab?

rnystrom:
I think Settings needs to stay a bar button+modal. It's too low use to take up a whole tab. Once we have a "profile" tab I'll move it there.

viktorgardart:
@Sherlouk @rnystrom thoughts on adding the searchbar to the navigation instead? 
<details>
  <summary>Screenshot</summary>

screen shot 2017-07-31 at 10 52 04
</details>

Sherlouk:
@viktorgardart I had considered that already! My only reasons for not doing it were I wanted to add a button to the nav for filtering, and was also thinking about adding search result count to title

Sherlouk:
Was toying with search and found a repo updated "9 years ago". I feel we should add a new case to our relative time which is "a decade ago" 😂

Sherlouk:
@rnystrom @BasThomas (@AnybodyElse) Would you mind doing an initial review?

Pretty much all the functionality I plan on adding is in there!

Sherlouk:
@rnystrom Nudge! Requiring a review

rnystrom:
Will give this a go soon. Note that I'm gonna leave it in a holding pattern so that editing labels and some issue/PR actions (close, lock, etc) ship _before_ sending this out.

Sherlouk:
Other than a couple points, it's mostly separated so shouldn't have any difficult conflicts! Thanks for taking a look

Classes/Repository/IssueSummaryType.swift
```diff
@@ -0,0 +1,74 @@
...
+        case PullRequestState.closed.rawValue where !isIssue: name = "git-pull-request"
+        case PullRequestState.merged.rawValue where !isIssue: name = "git-merge"
+        default: return nil
+        }
+        
+        return UIImage(named: name ?? "")?.withRenderingMode(.alwaysTemplate)
+    }
+    
+    var stateColor: UIColor? {
+        var hex: String?
```
BasThomas: Will the compiler complain when you make this a `let` and a `String` instead of a `String?` because of the `return nil`?

Classes/Repository/RepositoryClient.swift
```diff
@@ -0,0 +1,152 @@
...
+
+extension RepositoryLoadable {
+    var nameWithOwner: String {
+        return owner + "/" + name
+    }
+}
+
+protocol RepositoryQuery {
+    init(owner: String, name: String, after: String?)
+    func summaryTypes(from data: GraphQLMappable) -> [IssueSummaryType]?
```
BasThomas: Is there a difference between this being empty or `nil`? Feel like you could just make this non-optional?

Classes/Repository/RepositoryEmptyResultsSectionController.swift
```diff
@@ -0,0 +1,38 @@
...
+    }
+    
+    override func sizeForItem(at index: Int) -> CGSize {
+        guard let size = collectionContext?.containerSize else { fatalError("Missing context") }
+        return CGSize(width: size.width, height: size.height - topInset - topLayoutGuide.length)
+    }
+    
+    override func cellForItem(at index: Int) -> UICollectionViewCell {
+        guard let cell = collectionContext?.dequeueReusableCell(of: RepositoryEmptyResultsCell.self, for: self, at: index) as? RepositoryEmptyResultsCell else {
+                fatalError("Missing context, object, or cell is wrong type")
```
BasThomas: nit: too much indentation?

Classes/Repository/RepositoryEmptyResultsCell.swift
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+    
+    private let icon = UIImageView()
+    private let label = UILabel()
+    
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        
+        accessibilityTraits = UIAccessibilityTraitStaticText
```
BasThomas: Awesome! Will do a more thorough review of accessibility when I find the time.

Also, please use `|=` as to not cancel out any previous traits.

Classes/Search/SearchBarCell.swift
```diff
@@ -0,0 +1,44 @@
...
+import SnapKit
+
+final class SearchBarCell: UICollectionViewCell, UISearchBarDelegate {
+
+    weak var delegate: SearchBarSectionControllerDelegate?
+    private let searchBar = UISearchBar()
+    
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+
```
BasThomas: Accessibility? And - sorry for asking this - what is this? A cell with a searchbar? Why's that?

Classes/Search/SearchNoResultsCell.swift
```diff
@@ -0,0 +1,71 @@
...
+final class SearchNoResultsCell: UICollectionViewCell {
+    
+    private let emoji = UILabel()
+    private let label = UILabel()
+    private let shadow = CAShapeLayer()
+    
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        
+        accessibilityTraits = UIAccessibilityTraitStaticText
```
BasThomas: `|=`

Classes/Search/SearchResult.swift
```diff
@@ -0,0 +1,72 @@
...
+    }
+    
+    func isEqual(toDiffableObject object: ListDiffable?) -> Bool {
+        if self === object { return true }
+        guard let object = object as? SearchResult else { return false }
+        return id == object.id && name == object.name
+    }
+}
+
+struct GithubLanguage {
```
BasThomas: Put this in a separate file / or at the top maybe? Quite hard to find / easy to miss

Classes/Search/SearchBarCell.swift
Sherlouk: Just a cell with a search bar, correct - Reason for not putting it directly in the view controller came down to the layout mostly - Couldn't find a nice way (that wasn't causing issues with autolayout) to add the search bar above the collection view/below the nav bar hence it's in a cell.

I don't think this needs more accessibility than the search bar provides already?

Classes/Repository/RepositoryEmptyResultsSectionController.swift
Sherlouk: Correct, originally else was on it's own line which would've meant the indentation would be acceptable but in it's currently form yea

Classes/Repository/IssueSummaryType.swift
```diff
@@ -0,0 +1,70 @@
...
+        case IssueState.closed.rawValue where isIssue: return Styles.Colors.Red.medium.color
+        case PullRequestState.open.rawValue where !isIssue: return Styles.Colors.Green.medium.color
+        case PullRequestState.closed.rawValue where !isIssue: return Styles.Colors.Red.medium.color
+        case PullRequestState.merged.rawValue where !isIssue: return Styles.Colors.purple.color
+        default: return nil
+        }
+    }
+}
+
+class IssueSummaryModel: ListDiffable {
```
heumn: nit: Ryan has a  nice habit of locking down his classes unless they are meant for subclassing. Maybe continue with that? 🤔  (adding `final`)

Classes/Repository/RepositoryViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+    func reload() {
+        client.load(containerWidth: view.bounds.width) {
+            self.update(dismissRefresh: true, animated: true)
+        }
+    }
+    
+    func loadNextPage() {
+        if selection.issuesSelected {
+            client.loadMoreIssues(containerWidth: view.bounds.width) {
+                self.update(dismissRefresh: true, animated: false)
```
heumn: This _could_ leak and thus you should probably not retain `self` here? I am personally a fan of `unowned let` to make sure I crash if this block is called when it shouldn't (`client` should never outlive `self`). Ryan mostly defaults to `[weak self]` that is probably best? 🤔

Classes/Repository/RepositoryViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+            }
+        }
+        
+        return builder
+    }
+    
+    func listAdapter(_ listAdapter: ListAdapter, sectionControllerFor object: Any) -> ListSectionController {
+        guard let object = object as? ListDiffable else { fatalError("Object does not conform to ListDiffable") }
+        
+        // 28 is the default height of UISegmentedControl
```
heumn: 👍

Classes/Repository/RepositoryViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+        let controlHeight = client.repo.hasIssuesEnabled ? 28 + 2*Styles.Sizes.rowSpacing : 0
+        
+        if object === noIssuesResultsKey { return RepositoryEmptyResultsSectionController(topInset: controlHeight, topLayoutGuide: topLayoutGuide, type: .issues) }
+        else if object === noPullRequestsResultsKey { return RepositoryEmptyResultsSectionController(topInset: controlHeight, topLayoutGuide: topLayoutGuide, type: .pullRequests) }
+        else if object === selection { return SegmentedControlSectionController(delegate: self, height: controlHeight) }
+        else if object === loadMore { return SearchLoadMoreSectionController(delegate: self) }
+        else if object is IssueSummaryModel { return RepositorySummarySectionController(client: client.githubClient, repo: client.repo) }
+        
+        fatalError("Could not find section controller for object")
```
heumn: 💯

Classes/Search/GithubClient+Search.swift
```diff
@@ -0,0 +1,50 @@
...
+        case success(String?, [SearchResult])
+    }
+    
+    func search(query: String, before: String? = nil, containerWidth: CGFloat, completion: @escaping (SearchResultType) -> ()) {
+        let query = SearchReposQuery(search: query, before: before)
+        
+        apollo.fetch(query: query, cachePolicy: .fetchIgnoringCacheData) { (result, error) in
+            guard error == nil, result?.errors == nil else {
+                ShowErrorStatusBar(graphQLErrors: result?.errors, networkError: error)
+                return
```
heumn: I have seen these a couple of times in this PR (a `return` when the method accepts a completion handler). As a developer interactive with this API I would always expect my completion block to be called? 🤔  Is there something I am missing here that this is maybe intentional?

Classes/Repository/RepositoryClient.swift
```diff
@@ -0,0 +1,152 @@
...
+                summaryType.attributedTitle.computeSize(containerWidth)
+                return IssueSummaryModel(info: summaryType)
+            }
+            
+            completion(.success(RepositoryLoadSuccessPayload(models: summaries, nextPage: query.nextPageToken(from: data))))
+        }
+    }
+    
+    func loadMoreIssues(containerWidth: CGFloat, completion: @escaping () -> ()) {
+        guard repo.hasIssuesEnabled else { return }
```
heumn: I have seen these a couple of times in this PR (a `return` when the method accepts a completion handler). As a developer interacting with this API I would always expect my completion block to be called? 🤔  Is there something I am missing here that makes this intentional?

Classes/Search/GithubClient+Search.swift
Sherlouk: Admittedly had similar questions when doing this, but for the most part was copied and modified versions of the Notifications fetching - Error handling is probably something that could be handled better throughout the app as it is.

I'll take a look into this specifically though and see what improvements can be made

Classes/Repository/RepositoryClient.swift
```diff
@@ -0,0 +1,152 @@
...
+        
+        load(queryType: RepoIssuesQuery.self, repo: repo, after: issuesNextPage, containerWidth: containerWidth) { result in
+            switch result {
+            case .error:
+                completion()
```
heumn: 🤔  but here you actually _do_ call the completion, even if there are no issues?

Classes/Repository/RepositoryClient.swift
```diff
@@ -0,0 +1,152 @@
...
+                }
+                
+                self.issuesNextPage = payload.nextPage
+                completion()
+            }
+        }
+    }
+    
+    func loadMorePullRequests(containerWidth: CGFloat, completion: @escaping () -> ()) {
+        load(queryType: RepoPullRequestsQuery.self, repo: repo, after: pullRequestsNextPage, containerWidth: containerWidth) { result in
```
heumn: I think you forgot a `[weak self]` here :)

Classes/Repository/RepositoryEmptyResultsSectionController.swift
```diff
@@ -0,0 +1,38 @@
...
+    }
+    
+    override func sizeForItem(at index: Int) -> CGSize {
+        guard let size = collectionContext?.containerSize else { fatalError("Missing context") }
+        return CGSize(width: size.width, height: size.height - topInset - topLayoutGuide.length)
+    }
+    
+    override func cellForItem(at index: Int) -> UICollectionViewCell {
+        guard let cell = collectionContext?.dequeueReusableCell(of: RepositoryEmptyResultsCell.self, for: self, at: index) as? RepositoryEmptyResultsCell else {
+            fatalError("Missing context, object, or cell is wrong type")
```
heumn: 👍  nice use of fatalError

Classes/Repository/RepositorySummaryCell.swift
```diff
@@ -0,0 +1,86 @@
...
+        
+        reasonImageView.image = result.stateIcon
+        reasonImageView.tintColor = result.stateColor
+    }
+    
+    override var accessibilityLabel: String? {
+        get {
+            return contentView.subviews
+                .flatMap { $0.accessibilityLabel }
+                .reduce("", { $0 + ".\n" + $1 })
```
heumn: This is interesting, could you explain to me why you do this? 😄

Classes/Repository/RepositorySummarySectionController.swift
```diff
@@ -0,0 +1,44 @@
...
+    private let repo: RepositoryLoadable
+    
+    init(client: GithubClient, repo: RepositoryLoadable) {
+        self.client = client
+        self.repo = repo
+        super.init()
+    }
+    
+    override func sizeForItem(at index: Int) -> CGSize {
+        guard let width = collectionContext?.containerSize.width else { fatalError("Missing context") }
```
heumn: Off topic and just more of a question for @rnystrom. After I have been using IGListKit for a while I have come to the conclusion that most of the time, what I want here is not the raw size of the container, but the one with the insets `collectionContext?.insetContainerSize`. Thus if there are higher level insets on this collectionView, I do not have to care about them and still size my items accordingly :) Is there a reason (ryan) you are not defaulting to that?

Classes/Repository/RepositorySummaryCell.swift
Sherlouk: Same trick BasThomas has used for most of the cells; Nicely reads out all of the subviews with a pause in between each label.

Better than default, Easier than dynamically updating the text every time a new value is added

I'm sure Bas can convince you of it's merits ;)

heumn: That is actually a really nice trick. Def going into my "bag of UIKit tricks". Thanks for sharing!

Classes/Repository/RepositoryViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+//  Freetime
+//
+//  Created by Sherlock, James on 29/07/2017.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import UIKit
+import IGListKit
+
+class RepositoryViewController: UIViewController,
```
heumn: I think you forgot `final` here :)

Classes/Repository/RepositoryViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+        }
+    }
+    
+    func loadNextPage() {
+        if selection.issuesSelected {
+            client.loadMoreIssues(containerWidth: view.bounds.width) {
+                self.update(dismissRefresh: true, animated: false)
+            }
+        } else {
+            client.loadMorePullRequests(containerWidth: view.bounds.width) {
```
heumn: `[weak self]` I presume :)

Classes/Search/SearchBarSectionController.swift
```diff
@@ -0,0 +1,38 @@
...
+import IGListKit
+
+protocol SearchBarSectionControllerDelegate: class {
+    func didFinishSearching(term: String?)
+    func didCancelSearching()
+}
+
+final class SearchBarSectionController: ListSectionController {
+
+    weak var delegate: SearchBarSectionControllerDelegate? = nil
```
heumn: Tip you probably knew, but you do not actually need to set `= nil` for optional variables

Classes/Repository/RepositorySummaryCell.swift
BasThomas: Yes, returning a label constructed by all the subviews (that have a label). Needed because we want to add a trait to the cell, but then enable VoiceOver to still see the subviews (eg. `UILabel`) in the cell.

Classes/Search/SearchLoadMoreCell.swift
```diff
@@ -0,0 +1,39 @@
...
+        }
+        label.text = NSLocalizedString("Load More", comment: "")
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    override var accessibilityLabel: String? {
+        get { return NSLocalizedString("Load More Search Results", comment: "") }
```
heumn: I must say, 100 bonus points for defaulting to adding `accessibilityLabel` 👏  You inspire me. I will stop being a lazy developer from now on.

Classes/Search/SearchResult.swift
```diff
@@ -0,0 +1,72 @@
...
+
+import Foundation
+import IGListKit
+
+struct GithubLanguage {
+    let name: String
+    let color: UIColor?
+}
+
+class SearchResult: ListDiffable, RepositoryLoadable {
```
heumn: Missing final :)

Classes/Search/SearchBarSectionController.swift
Sherlouk: Sometimes my brain and fingers don't sync up 😬 😅

Classes/Search/SearchResultSectionController.swift
```diff
@@ -0,0 +1,42 @@
...
+
+    private let client: GithubClient
+    
+    init(client: GithubClient) {
+        self.client = client
+        super.init()
+    }
+    
+    override func sizeForItem(at index: Int) -> CGSize {
+        guard let width = collectionContext?.containerSize.width else { fatalError("Missing context") }
```
heumn: This is so much better than what I previously did (`return .zero` Hah, good luck debugging that you fool!)

Sherlouk:
@heumn Thanks for the thorough review - Will be sure to fix it all up soon!

Classes/Repository/RepositorySummaryCell.swift
BasThomas: And because Objc, you have this awkward `get{} set{}` with the empty `set{}`. 🤷‍♂️

Do maybe want to do a PR to change all these to use string interpolation (as opposed to the plussing) to prevent the compiler from complaining about the expression being too complex.

Classes/Search/SearchViewController.swift
```diff
@@ -0,0 +1,151 @@
...
+//  Freetime
+//
+//  Created by Sherlock, James on 28/07/2017.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import UIKit
+import IGListKit
+
+class SearchViewController: UIViewController,
```
heumn: Missing final :)

Classes/Search/SearchViewController.swift
```diff
@@ -0,0 +1,151 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        feed.viewDidLoad()
+        feed.collectionView.refreshControl?.endRefreshing()
+        feed.adapter.dataSource = self
+        title = "Search"
```
heumn: I think you forgot to localize this

Classes/Search/SearchViewController.swift
```diff
@@ -0,0 +1,151 @@
...
+    // MARK: Data Loading/Paging
+    
+    private func update(dismissRefresh: Bool, animated: Bool = true) {
+        feed.finishLoading(dismissRefresh: dismissRefresh, animated: animated)
+    }
+    
+    private func handle(resultType: GithubClient.SearchResultType, append: Bool, animated: Bool) {
+        switch resultType {
+        case .error:
+            print("😞 Something went wrong, here's some pizza 🍕")
```
heumn: :rofl:

Classes/Search/SearchViewController.swift
Sherlouk: Universal term, it's "Search" in all languages 😉 Oops... 😄

Classes/Search/SearchViewController.swift
Sherlouk: Jokes aside while a better message could probably be shown, it does show the status bar error screen and then an empty results placeholder view so the user isn't left in limbo

Similar experience to notifications view

Classes/Repository/RepositoryClient.swift
```diff
@@ -0,0 +1,152 @@
...
+            }
+            
+            completion(.success(RepositoryLoadSuccessPayload(models: summaries, nextPage: query.nextPageToken(from: data))))
+        }
+    }
+    
+    func loadMoreIssues(containerWidth: CGFloat, completion: @escaping () -> ()) {
+        guard repo.hasIssuesEnabled else { return }
+        
+        load(queryType: RepoIssuesQuery.self, repo: repo, after: issuesNextPage, containerWidth: containerWidth) { result in
```
heumn: [weak self] would probably be nice here :)

Classes/Search/SearchBarSectionController.swift
heumn: Hah, story of my life. Mostly I would say none of them seem to work on my end ;)

heumn:
Just a bunch of nit and praise. Good job man! 💯

dkhamsing:
Hey @rnystrom,  any time to look at this now?

BasThomas:
ping @rnystrom  :)

rnystrom:
Ok cool, here's all the follow up that I think I'll take on, as I'm at the point where I'd like to start using this:

- Fixup merges
- Rename the "Search" tab to "Repos" (maybe "Repositories")
- Default load _your_ repositories on the repo tab (could take advantage of init state here)
- Put the search bar in the nav view
- Search as you type (instead of waiting for enter)
- Auto load next page as you scroll down search results
- Merge `SearchLoadMoreCell` and `IssueNeckLoadCell` into one
- Redesign repo cells
  - Remove "updated" date, I don't think this is valuable info in a search result
  - Balance colors
- Fix bug showing empty state while loading issues
- Clean up empty state design (no black)
- Clean up Issue/PR cell design to match notifications (bold, spacing)
- Bug tapping issue presents VC, can't go back (both on new repo and old one)
- iPad shows 2 search tabs?
- Cells need highlight state
- Cells need selection/deselection states

@Sherlouk idk how much you wanna sign up to fix? I'll need to take care of all of this before I can ship a build.

rnystrom:
I'm actually gonna merge this and just start cleaning stuff up as I see fit. It'll be a good forcing function for me to ship this. I've been wishing I could browse older IGLK issues, so this is a good start!

BasThomas:
Maybe make an issue to keep track of the cleanup? :)

Sherlouk:
Well this was unexpected! Definitely agree let's get an issue to track this, and I'll definitely help tidy up where possible! 🎉

Sherlouk:
Also this PR (finally) put Swift as our primary language 😃 

image

PR #214, opened 2017-07-30
Change Secret keys to Xcode Text Placeholders

lottadot:
* The project won’t compile until the developer updates with their own keys.
* Added comment to documentation about setting up the keys.

### DIFF
--- a/Classes/Other/Secrets.swift
+++ b/Classes/Other/Secrets.swift
@@ -10,7 +10,8 @@ import Foundation
 
 public enum GithubAPI {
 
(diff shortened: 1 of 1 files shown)


rnystrom:
In gonna land #197, might want to double check that all works together (haven't loon at the changes here yet).

shaps80:
This won't work since we no longer define the keys there. Also, this isn't really ideal anyway because it would still require the develop to remember to revert this in their commits.

#197 IMHO is a much better fix.

PR #226, opened 2017-08-04
Changed maximumNumberOfLines to Int from UInt

gtchance:
This fixes the playground errors and keeps the maximumNumberOfLines
property type consistent with maximumNumberOfLines on NSTextContainer.


rnystrom:
<3

PR #228, opened 2017-08-04
Feature add network activity indicator

### LINKED ISSUE #221: Add network activity indicator
Version 1.4.0 (1136)
Device: iPhone 7 Plus (iOS 10.3.3)

gtchance:
Closes #221
I kept the startDelay at the default 1 second. Most of the requests are
so fast the indicator doesn’t show, but it shows when it counts, i.e.
requests that last longer than a second.


Sherlouk:
How does this work with GraphQL/Apollo queries, because unless I'm being completely stupid I don't think that runs through Alamofire?

gtchance:
It only covers Alamofire. From what I can tell, that includes everything except loading content in the IssuesViewController feed, which already includes a large activity indicator.

rnystrom:
I'll merge this for now. There will def be more GraphQL requests as time goes on, but I think this is a good start!

PR #270, opened 2017-08-31
stop reloading  the detail view

### LINKED ISSUE #244: Selecting the currently selected issue reloads the detail view.
Very low priority tbh just adding for posterity because it’s a less than ideal UX

Version 1.5.0 (1330)
Device: iPad7,3 (iOS 11.0)

TheCodeTalker:
Fixed #244

### DIFF
--- a/Classes/Notifications/NotificationSectionController.swift
+++ b/Classes/Notifications/NotificationSectionController.swift
@@ -15,7 +15,7 @@ SwipeCollectionViewCellDelegate {
 
     private let client: NotificationClient
     private let dataSource: NotificationsDataSource
-
+    private static var selectedCellId : String? = ""
     init(client: NotificationClient, dataSource: NotificationsDataSource) {
         self.client = client
         self.dataSource = dataSource
@@ -49,9 +49,11 @@ SwipeCollectionViewCellDelegate {
             cell.isRead = true
             client.markNotificationRead(id: object.id, isOpen: true)
         }
-
+        if NotificationSectionController.selectedCellId != object.id || UIDevice.current.userInterfaceIdiom == .phone{
+            NotificationSectionController.selectedCellId = object.id
         let controller = NavigateToNotificationContent(object: object, client: client.githubClient)
         viewController?.showDetailViewController(controller, sender: nil)
+        }
     }
 
     // MARK: Private API
@@ -62,6 +64,10 @@ SwipeCollectionViewCellDelegate {
         collectionContext?.performBatch(animated: true, updates: { context in
             context.reload(self)
         })
+    }
+    
+    deinit {
+        NotificationSectionController.selectedCellId = nil
     }
 
     var considerObjectRead: Bool {


Classes/Notifications/NotificationSectionController.swift
```diff
@@ -49,9 +49,11 @@ SwipeCollectionViewCellDelegate {
             cell.isRead = true
             client.markNotificationRead(id: object.id, isOpen: true)
         }
-
+        if NotificationSectionController.selectedCellId != object.id || UIDevice.current.userInterfaceIdiom == .phone{
```
BasThomas: Why is there iPhone specific code in there?

Classes/Notifications/NotificationSectionController.swift
```diff
@@ -15,7 +15,7 @@ SwipeCollectionViewCellDelegate {
 
     private let client: NotificationClient
     private let dataSource: NotificationsDataSource
-
+    private static var selectedCellId : String? = ""
```
BasThomas: Why did you opt for a static variable here?

Classes/Notifications/NotificationSectionController.swift
rnystrom: +1, let's remove

Classes/Notifications/NotificationSectionController.swift
rnystrom: Ya, let's make this a local variable. Also can it just be a `Bool` since there's only one cell?

Classes/Notifications/NotificationSectionController.swift
```diff
@@ -63,6 +65,10 @@ SwipeCollectionViewCellDelegate {
             context.reload(self)
         })
     }
+    
+    deinit {
+        NotificationSectionController.selectedCellId = nil
+    }
```
rnystrom: Can remove if we use instance vars

Classes/Notifications/NotificationSectionController.swift
TheCodeTalker: the Details view reloading problem with IPad not in iPhone i am just bypassing the condition check to showDetailViewController and in iPad its checks for the id if not matching i am calling showDetailViewController in that way I don't need reload again.  or every time it click the cell it returning IssuesViewController new instance because of that it reloading again then probably we need to change the object creating style .

Classes/Notifications/NotificationSectionController.swift
TheCodeTalker: if i declared as local variable to NotificationSectionController then  i seen NotificationsViewController returning  ex- 5 cell then 5 NotificationSectionController and all are having selectedCellId to default as false if i use .  and your correct since there's only one cell per NotificationSectionController storing the state change to how can know with Bool so that i can rest to false . so **i choose global to all section Controller** to track state change  .??? and in deinit in releasing the variable form memory..

rnystrom: @TheCodeTalker sorry having trouble following. There is only one `NotificationSectionController` instance per cell, so tracking a single `Bool` per section controller will map directly to the cell. Global state will inevitably lead to trouble.

TheCodeTalker: @rnystrom   is the click of cell changes how to reset the bool value in that case???

PR #271, opened 2017-08-31
Readds accesability escape to settings

BasThomas:
Seems like it was accidentally removed in https://github.com/rnystrom/GitHawk/commit/87150d26e4189ce676331766f500b9ef7b2dda8b.


rnystrom:
I think we can close this now since Settings is in a tab.

PR #275, opened 2017-09-02
More cleanup after #200

Sherlouk:
Mostly making things `[weak self]` to prevent possible retain cycles, marking things `final` and some other nits raised by @heumn in the original pull request

Only outstanding ones are those pertinent to error handling and your wall of text (@rnystrom) but I'll put these in a checklist in a new issue to handle separately


rnystrom:
@Sherlouk I did a lot of moving around here, gonna close this out since I think most of the stuff is outdated already. Lmk if there's something critical you think I'm missing.

PR #308, opened 2017-09-14
Added the pod `HTMLString`, fixes #307

### LINKED ISSUE #307: Convert HTML entities when using in NSAttibutedString
Version 1.10.0 (1775)
Device: iPhone 7 Plus (iOS 10.3.3) 

E.g. these emoji

https://raw.githubusercontent.com/Instagram/IGListKit/master/README.md

weyert:
Updated the Markdown parsing so it calls `HTMLString`'s `removingHTMLEntities`-method so the html entities get converted to the real deal like a Emoji
Fixed the issue that the height of a table row was hard-coded to `30` now it's being calculated and the biggest tone is used for all other rows

### DIFF
--- a/Classes/Issues/Comments/Tables/IssueCommentTableModel.swift
+++ b/Classes/Issues/Comments/Tables/IssueCommentTableModel.swift
@@ -31,21 +31,12 @@ final class IssueCommentTableModel: NSObject, ListDiffable {
         }
     }
 
-    let rowHeight: CGFloat
+    let rowHeight = Styles.Sizes.labelEventHeight
     let columns: [Column]
     let totalHeight: CGFloat
 
     init(columns: [Column]) {
         self.columns = columns
-        self.rowHeight = columns.reduce(0) { currentRowHeight, column  in
-            let maxHeight: CGFloat = column.rows.reduce(0.0) { currentMaxHeight, row in
-                let textSize = row.text.computeSize(0)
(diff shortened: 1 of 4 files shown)


Classes/Issues/Comments/Tables/IssueCommentTableModel.swift
```diff
@@ -31,12 +31,21 @@ final class IssueCommentTableModel: NSObject, ListDiffable {
...
         self.columns = columns
+        self.rowHeight = columns.reduce(0) { currentRowHeight, column  in
+            let maxHeight: CGFloat = column.rows.reduce(0.0) { currentMaxHeight, row in
+                let textSize = row.text.computeSize(0)
+                return max(currentMaxHeight, textSize.height)
+            }
+            
+            let inset = IssueCommentTableCollectionCell.inset.top + IssueCommentTableCollectionCell.inset.bottom
+            return max(currentRowHeight, maxHeight + inset)
+        }
```
rnystrom: I'm going to strip this part out and revisit. This is an n^2 algo that should be calculated at the parsing phase. Asking the text sizing for the computeSize again might end up doing wasted work.

weyert: Makes sense, have to say I didn't like it that much myself but fixed the problem 🤗 Parsing phase is happening MMarkdown, I presume?

PR #309, opened 2017-09-15
Removes search results when search bar textfield cleared/emptied, fixes #305

### LINKED ISSUE #305: On iPad, there's no cancel button for the search bar in Search
Version 1.10.0 (1770)
Device: iPad Pro (iOS 10.3.3)

rizwankce:
As system restricts Cancel button on iPad, added an option to remove search results and show recent searches when search bar text field is cleared.

### DIFF
--- a/Classes/Search/SearchViewController.swift
+++ b/Classes/Search/SearchViewController.swift
@@ -157,6 +157,15 @@ SearchRecentHeaderSectionControllerDelegate {
     }
 
     // MARK: UISearchBarDelegate
+    
+    func searchBar(_ searchBar: UISearchBar, textDidChange searchText: String) {
+        guard let term = searchBar.text?.trimmingCharacters(in: .whitespacesAndNewlines),
+            term.characters.count > 0 else {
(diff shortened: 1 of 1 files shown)


Classes/Search/SearchViewController.swift
```diff
@@ -157,6 +157,15 @@ SearchRecentHeaderSectionControllerDelegate {
     }
 
     // MARK: UISearchBarDelegate
+    
+    func searchBar(_ searchBar: UISearchBar, textDidChange searchText: String) {
+        guard let term = searchBar.text?.trimmingCharacters(in: .whitespacesAndNewlines),
```
Sherlouk: Personally not a fan of using a guard for the sole purpose of using the `else` statement - Can we change this out to either be an `if` statement or flip the guard around to check that the count is equal to 0 else return

BasThomas: If you flip, you can also do `isEmpty` instead of `== 0`.

PR #312, opened 2017-09-16
Swipe between images in same comment

### LINKED ISSUE #123: Unable to swipe through multiple images in-app
Eg. #120

Sherlouk:
Closes #123 

In it's current implementation it puts the tapped image first and then swipe through the other ones. Could maybe make it simpler by keeping them in original order and changing the initial image with NYT - The downside of this is if you get put in the middle of a gallery, might be a bit confusing to the user?

Was just toying around not precious about this change!


Classes/Issues/Comments/Images/IssueCommentImageCell.swift
```diff
@@ -24,7 +24,8 @@ UIGestureRecognizerDelegate {
 
     weak var delegate: IssueCommentImageCellDelegate? = nil
     let imageView = UIImageView()
-
+    
+    private var imageUrl: URL!
```
rnystrom: This shouldn't be implicit unwrap since `bindViewModel(...)` **must** be called first. Can we change it to optional and just `guard` around it?

Sherlouk: Done!

Classes/Systems/PhotoViewHandler.swift
```diff
@@ -8,24 +8,36 @@
...
-        let photo = IssueCommentPhoto(image: image)
-        let photosViewController = NYTPhotosViewController(photos: [photo])
+        
+        var photos = [IssueCommentPhoto(image: image)]
+        
+        imageModels?.forEach({
+            guard $0.url.absoluteString != url.absoluteString else { return }
+            guard let image = SDImageCache.shared().imageFromCache(forKey: $0.url.absoluteString) else { return }
+            photos.append(IssueCommentPhoto(image: image))
+        })
```
rnystrom: noting to myself that this implies that `SDImageCache` will always be used. Ok for now.

Sherlouk: Like I said was just toying around to see what was possible - completely understand if these limitations are too much

NYT is definitely a limiting factor here as they require an image straight up and don't allow any sort of async loading. Maybe a local fork or alternative library would be better suited - Card it up?

Sherlouk:
Gonna go ahead and close this, think there are bigger issues around the photo library we use and it's limitations

Issue #314, opened 2017-09-17
Load local path images


rnystrom:
Images in repos can have relative paths. Should pick this up and prepend base URL so SDWebImage can load it.

rnystrom:
Picking this back up over here. @Sherlouk what did you do to make this work in #315? Wanted to get this working in 1.10.0, but ok sacrificing for now. If you had it working, wondering if I broke something you did?

PR #315, opened 2017-09-17
Allow relative images to display in README files

Sherlouk:
Relates to #314 

So this does fix the issue in the GitHawk README, but we're still going to have issues with links going to relative destinations such as to a guide MD file or similar.

Thinking we need to check the file extension, see what the type is?

@rnystrom Advice?

### DIFF
--- a/Classes/Issues/Comments/Html/IssueCommentHtmlCell.swift
+++ b/Classes/Issues/Comments/Html/IssueCommentHtmlCell.swift
@@ -58,6 +58,7 @@ final class IssueCommentHtmlCell: UICollectionViewCell, ListBindable, UIWebViewD
 
     @objc private let webView = UIWebView()
     private var body = ""
+    var webViewBaseURL: URL?
 
     override init(frame: CGRect) {
         super.init(frame: frame)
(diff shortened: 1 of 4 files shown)


rnystrom:
Good point. Lets tackle that next.

rnystrom:
Ah dang, this doesn't work now that I removed issue shorthand from readme's (b/c the `owner` and `repo` is `nil` in that case). Might need to change that into an `options` mask.

Sherlouk:
Ahh bummer! Revert the change and I'll try and approach the issue again!

rnystrom:
@Sherlouk np! This was the right way to do it, just stepped on your toes lol. I'm fixing it up by adding the options mask to MD parsing, that way we can always deliver `owner`+`repo` but add options like issue shorthand, base url, etc and have different parsing behavior on different surfaces.

rnystrom:
Hmm, even after fixing it, I can't get the relative image to show up in the GitHawk readme. I'll push my changes @Sherlouk mind taking a look?

Sherlouk:
Taken a quick look but may need to leave it until morning - The base URL is definitely getting through correctly when the request is being made so bit confused actually

Was definitely working with the original implementation on the GitHawk readme

rnystrom:
Ya the base URL is being used for sure. Maybe it's a simulator issue, trying on device.

Sherlouk:
Admitting defeat for tonight because I need sleep, will take look tomorrow

Maybe a different approach would be to use a regex detecting relative URLs, iterate through them and prefix them manually before we even render the UIWebView? That way we can switch it out depending on the URL extension (MD different than PNG etc)

PR #326, opened 2017-09-20
Always dismiss keyboard on search result selection

matt-holden:
Issue:

It appears that https://github.com/rnystrom/GitHawk/commit/66bdf3d835528bba22c8a0ff64e7b8d958178bf0 introduced a bug with search result selection.

After the change from `navigationController.pushViewController` to `showDetailViewController`, it was possible to enter a state where the `RepositoryViewController` would be pushed onto the stack without dismissing the keyboard.

To recreate:

1. Have at least 1 recent search.
2. Select a recent search item.
3. Tap in the search bar to make it the first responder.
4. Select an item in the list from the original search results.
5. RepositoryViewController will be pushed onto the stack, but the keyboard will remain visible.

Solution:

Declare and implement `SearchResultSectionControllerDelegate`. This delegate is used to notify the `SearchViewController` that a search result has been selected.

In turn, the `SearchViewController` will resign first responder status of its search bar.

### DIFF
--- a/Classes/Search/SearchResultSectionController.swift
+++ b/Classes/Search/SearchResultSectionController.swift
@@ -8,12 +8,18 @@
 
 import IGListKit
 
+protocol SearchResultSectionControllerDelegate {
+    func didSelect(result: SearchRepoResult)
+}
+
 final class SearchResultSectionController: ListGenericSectionController<SearchRepoResult> {
 
     private let client: GithubClient
-    
-    init(client: GithubClient) {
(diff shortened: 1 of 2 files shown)


matt-holden:
### UPDATE  

(and facepalm)

My reproduction steps aren't correct... and now for the life of me I can't seem to recreate the issue that prompted me to fix this in the first place.

For now, I'll close this and see if I can't recreate the original bug.

Oops!

Issue #114, opened 2017-07-11
Improve compile times


rnystrom:
I took a stab at removing CocoaPods and didn't notice any build improvements. Instead of hacking at it, it'd be good to do some [profiling](http://irace.me/swift-profiling) of builds. Will do this and report back.

Any good tools for profiling Swift builds?

rnystrom:
Dang, some gnarly ones:

https://github.com/rnystrom/Freetime/blob/master/culprits.txt

Sherlouk:
Also worth just making sure the whole project optimisation settings are correct; that alone can save a lot of build times!

MosheBerman:
I’ve used this one before: https://github.com/RobertGummesson/BuildTimeAnalyzer-for-Xcode/

I don’t know if works with Xcode 9, haven’t used it since Swift 2.3.

rnystrom:
Should be helpful

http://khanlou.com/2016/12/guarding-against-long-compiles/

PR #327, opened 2017-09-20
Improve compile times (Issue #114)

matt-holden:
Re: #114 

_This fix dropped my average re-build (no code changes) time from 60-90 seconds to 7-10 seconds._

### Summary

It appears to me that the largest contributor to compile times is `API.swift` and its re-generation in the "Generate Apollo GraphQL API" build phase.

The build script currently overwrites `API.swift` on every build, resulting in a new last-modified timestamp. As such, the swift compiler won't use its cached build of the file.  (You'll have to excuse me, I don't know much about the compiler, so I may be butchering terminology.)

In my estimation, this causes XCode to not only recompile `API.swift`, but also _all files that reference its contents_.

### Fix

Modify the "Generate Apollo GraphQL API" build phase to only overwrite `API.swift` if the output of `check-and-run-apollo-codegen.sh` results in a different output than the previous build.

The diff is hard to read, as it's in the pbxproj, so I've included the relevant lines below:

```sh
TEMP_FILE=$(mktemp)
[[ -f API.swift ]] || touch API.swift # ensure sure file exists

$APOLLO_FRAMEWORK_PATH/check-and-run-apollo-codegen.sh generate $(find . -name '*.graphql') --schema schema.json --output $TEMP_FILE

cmp API.swift $TEMP_FILE || cp $TEMP_FILE API.swift
```

In short, it outputs the generated API file to `TEMP_FILE`. Then, it
uses `cmp` to test if `TEMP_FILE` is identifical to the existing `API.swift`. If they are not, it overrwrites `API.swift` with the newly generated code.

### DIFF
--- a/Freetime.xcodeproj/project.pbxproj
+++ b/Freetime.xcodeproj/project.pbxproj
@@ -1674,7 +1674,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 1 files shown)


rnystrom:
> This fix dropped my average re-build (no code changes) time from 60-90 seconds to 7-10 seconds.

You are my hero 🌟 🎉 ❤️  🌮 🦅 🇺🇸 💥 🎉 😍  🌟

rnystrom:
@matt-holden you might want to consider suggesting this change upstream:

- [Apollo iOS](https://github.com/apollographql/apollo-ios)
- [Their script docs](http://dev.apollodata.com/ios/installation.html#adding-build-step)

matt-holden:
Will do!

matt-holden:
@rnystrom -- opened https://github.com/apollographql/apollo-ios/issues/147

Issue #328, opened 2017-09-20
Webpage handoff


BasThomas:
Support Handoff in the app when viewing a page that's also available on web (so all pages that we can open in Safari too).

PR #331, opened 2017-09-20
Add handoff support for issues

BasThomas:
Ref #328.

While implementing this I was wondering though... opening a web page with Handoff would automatically mark an issue as read (as you would load the page on GitHub.com). Do we maybe want to only allow Handoff when the user has this setting enabled?

This would only make sense for the issues though - we can still support handoff for readmes, profiles etc. Not sure what would make most sense here.


Sherlouk:
Is this a two way handoff or only from App -> Web? Would be handy to quickly pick up where I was!

I also feel as a user you expect that going to the web will have slightly different functionality, I think it's fine that it reads it

BasThomas:
> Would be handy to quickly pick up where I was!

What? You were working on this?

And yes, this is only app > web.

Sherlouk:
> You were working on this?

No! Sorry I meant it would be handy to pick up where I was so that if I was replying to issues on my desktop and wanted to jump in a taxi then I could quickly pick it up on my phone

BasThomas:
Ha, was confused. It is unfortunately not possible to go from Web -> app, unless it’s your domain. As we are working with GitHub... 😔

What we **could** do is to have a url scheme / share extension that can open GH urls in the app?

BasThomas:
What we can also do is Handoff from app -> app. So switching from iPad to iPhone would be easier.

Sherlouk:
App to App would be good, I also know a month or so ago I spoke to Ryan about having a custom domain as a middleman for GitHub

Long explanation short a domain could be something like githawk.com/rnystrom/GitHawk/pull/331 - If the app is installed then it will take them to it, otherwise it just redirects the user to the github.com alternative. (Universal linking)

That in combination with share extensions would be a powerful combination

BasThomas:
Yep, the middelman route is pretty neat.

rnystrom:
I'll take this for now and we can try it out. I honestly don't see this getting a ton of use, so we can fix the edge cases and UX if ppl start using it a lot.

PR #354, opened 2017-09-21
Add FLEX to TestFlight builds

### LINKED ISSUE #90: Enable FLEX for Testflight builds
- `git mv` the `FlexController` to `/Classes/Systems`
- Add compilation flag
- Use flag in builds for Testflight
  - Somehow add the custom flag when doing `fastlane beta`
- Make sure FLEX is [compiled in](https://github.com/rnystrom/Freetime/blob/91261835abca858adba7c62a3a4ab4b48386e137/Classes/Views/FlexController.m)

Sherlouk:
Closes #90 

Haven't actually tested the full end-to-end flow because yano I don't have an app to test pilot

Changes are masked by pod so I'll explain:

- Duplicated the `Release` configuration and named it `TestFlight`
- Added a preprocessor flag `TESTFLIGHT` for builds with that release
- Updated the `Podfile` to have FLEX install for builds against the `TestFlight` configuration
- Updated the `Fastfile` so that `fastlane beta` uses the `TestFlight` configuration
- Created `fastlane appstore` that does very little (May want to fill that out yourself) but by not specifying the configuration it defaults to `Release`
- Updated `FLEX` to trigger if you have the preprocessor flag `TESTFLIGHT`

Makes no sense? Great 😅


Sherlouk:
Admittedly it's currently failing to compile because of SlackTextViewController

Investigating, the premise is there though

Sherlouk:
Got it building and flex works if you use the keyboard shortcuts, just trying to see if I can get the triple tap working

Sherlouk:
Ignore me I read it as a triple tap, not using 3 fingers to tap. It should work on a phone!

Sherlouk:
Temporarily changed it to 1 finger tap to get it to show, and on a TestFlight build can confirm it works

fastlane/Fastfile
```diff
@@ -15,8 +15,14 @@ default_platform :ios
 # info which you can add here later
 lane :beta do
   # match(type: "appstore")
-  gym(scheme: "Freetime")
+  gym(scheme: "Freetime", configuration: "TestFlight")
```
rnystrom: sick TIL

rnystrom:
@Sherlouk you're my hero, this is so cool

Sherlouk:
@rnystrom An extension to this I'd quite like to see (maybe spin up a new card) is if on test flight to add a new section to the settings page which has some stuffs under the hood?

Namely (most interesting to me) if we ping off to the rate limits API for both v3 and v4 and display that information. I think it's important for us to monitor this under "normal" usage and see if we're making too many requests or if we're doing fine!

Obviously this is per-person and depends on usage, and I don't think we'll get a true representation until it's in a testflight build!

Sherlouk:
@rnystrom Saw a new build this morning, did it all go successfully?

(I do have FLEX so that must mean something went well 😄)

rnystrom:
Yup! Then it worked!

BasThomas:
How do I access it?

rnystrom:
@BasThomas three finger tap

Issue #356, opened 2017-09-21
Build Number Auto-increment


Sherlouk:
Honestly the auto-incrementing build number has always been an annoyance in team projects because every PR immediately has a conflict and while it takes two seconds to fix I'd rather not have it 😂 

Obviously this is complete personal preference but an alternative solution would be to do it based on git:

```shell
#!/bin/sh

# Retrieve git information
commitHash=$(git rev-parse --short HEAD)
commitCount=$(git rev-list HEAD --count)

if [ $CONFIGURATION = "Release" ]
then
    echo "RELEASE BUILD"
    bundleVersion=$commitCount
else
    echo "DEVELOPMENT BUILD"
    bundleVersion=$commitHash
fi

# Update Build Identifier
echo "CFBundleVersion = $bundleVersion"

if [ -f "${BUILT_PRODUCTS_DIR}/GitHawk.appex/Info.plist" ]; then
    /usr/libexec/PlistBuddy -c "Set :CFBundleVersion $bundleVersion" "${BUILT_PRODUCTS_DIR}/GitHawk.appex/Info.plist"
else
    echo "GitHawk not found so bundle update not applied to that plist"
    echo "- ${BUILT_PRODUCTS_DIR}/GitHawk.appex/Info.plist"
fi
```

So this does add another dependency (which we could/should store in the repository) but with this as a runtime script it doesn't directly alter the info.plist, it changes the one in the actual build product ultimately meaning we never see it increment and therefore no conflicts

@rnystrom Happy to close this as a non-issue if you're happy with it as is but thought I'd raise an alternative solution regardless!

rnystrom:
What's the dependency?

Sherlouk:
PlistBuddy is the dependency, used to manipulate the Info.plist after the fact.

Which having just checked we already use, so fun fact no new dependencies. So this can be as complicated or as simple as you make it, as an open-source app it might actually be nice to use the hash as it's publicly available and makes it clear what's included in the production app!

rnystrom:
IMO let's move to hash, I think that's a neat idea. Really ties the whole open source bit together nicely!

Sherlouk:
`commitHash=$(git rev-parse --short HEAD)` returns the 7 character hash which GitHub handles fine

rnystrom:
🚀🚀🚀 lets do it!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Sherlouk:
[...]

PR #386, opened 2017-09-23
Updated "Increment Build Number" script

Sherlouk:
Relates to #356 

No more Info.plist merge conflicts 😱 

I've never tried pushing to Apple with a hash value as the build number, if they do cry about it we can always do the original script in the attached issue which uses the commit count for release builds.


rnystrom:
Looks like it will probably get rejected 

https://developer.apple.com/library/content/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364

Another script:

https://gist.github.com/jpwatts/966838

From here 

https://stackoverflow.com/questions/8192921/set-xcode-bundle-version-to-git-hash

Sherlouk:
The issue with `git rev-list HEAD --count` is that it's only at 892 😂 
Our Info.plist is over 1000 so we'd probably hit issues with conflicting/lower build numbers

rizwankce:
Apple will reject if build number is fit hash and there won’t be any conflict if we increment a version. 

BasThomas:
I think you can reset the build number after every version though, so that shouldn’t be a problem @Sherlouk

Sherlouk:
Pushed using commit count, if what @BasThomas has said is true then this should work for us going forward

Sherlouk:
There's an irony to the conflicted file

rnystrom:
@Sherlouk how about having the build # be this count, but then we include the hash as a custom key/value in the info plist? I really really like the idea of being able to tap into the release commit from settings, haha.

Sherlouk:
@rnystrom Honestly I'd like to see us take more use of Releases on GitHub and correctly tag up when we push. Not only will we have a user-readable tag to link the user to, but we can attach release notes etc

This could be done automatically using fastlane

rnystrom:
@Sherlouk agree, time to get more serious about these releases. That's why I setup the milestones too.

> This could be done automatically using fastlane

How so?

Sherlouk:
> How so?

https://docs.fastlane.tools/actions/add_git_tag/
https://docs.fastlane.tools/actions/push_git_tags/

BasThomas:
We should create an issue for that :)

Resources/Info.plist
```diff
@@ -32,7 +32,7 @@
 		</dict>
 	</array>
 	<key>CFBundleVersion</key>
-	<string>1920</string>
+	<string>1</string>
```
rnystrom: @Sherlouk shouldn't this read 892 if the script works?

Sherlouk: Nope, the whole point of this script is that Resources/Info.plist **never changes**!

The actual compiled version of the app (found in your derived data) will have a different Info.plist with the correct value.

Again, the whole point of this change is to stop merge conflicts from every branch having a different build version.

rnystrom:
Note: reverted for now, getting in the way too much from shipping TF builds atm. Will def try to revisit.

Issue #282, opened 2017-09-04
Document Apollo install steps


BasThomas:
Just set up fresh, and noticed we haven't documented the steps to install the Apollo node package. Then it gave an error the Apollo framework and latest node version are out of sync - we might want to update the framework also?

rnystrom:
Hmm good point. Ya I updated apollo once and it all went to hell. Pinned to a specific version.

Sherlouk:
`npm install -g apollo-codegen@10.0`

Sherlouk:
Gonna close this since we have NPM now and the contributing guide also references this.

It's also worth noting that when we upgrade Apollo, it has automatic version syncing, meaning it'll automatically ensure you have the correct version installed before building - So this won't be an issue any further

PR #387, opened 2017-09-23
Setup prebuild script to ease buddybuild on boarding 

Palleas:
Hi there!

@BasThomas showed me the app and I really like it. Since I'm a dev evangelist at buddybuild, I asked about CI and he redirected me to #183. 

I've forked the app and [on boarded it in buddybuild](https://dashboard.buddybuild.com/public/apps/59c652e2f96f890001f0aa44/build/59c660c184d8e8000188ce27#logs
), if you feel like taking buddybuild for a spin, it should work right away, the only missing stuff was adding the Apollo Codegen thingie. 

I also enabled the execution of tests, but they seem to be failing for now: https://dashboard.buddybuild.com/public/apps/59c652e2f96f890001f0aa44/build/59c668f7fc18d80001f033df

Let me know if there is anything else I can do to help!

PR #401, opened 2017-09-23
Add dependency to Apollo-codegen

Palleas:
Addresses #282.

I figured the best way to do this was to use `yarn` to pin a specific version, let me know if you think it's overkill. 🤔

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,4 @@ fastlane/report.xml
 fastlane/Preview.html
 fastlane/screenshots
 fastlane/test_output
+node_modules
--- a/README.md
+++ b/README.md
@@ -38,6 +38,11 @@ GitHawk will be released with (at least) a three-phased rollout:
     - Accept, reject, and merge PRs
     - :rocket:
 
+## Installation
+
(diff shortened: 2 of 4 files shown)


BasThomas:
Holy cow, that `yarn.lock` 😱

Sherlouk:
Obviously this comes down to personal preference, but I feel `npm` has been round for a lot longer n more commonly used. Not sure if it's easier just to get people to do `npm install -g apollo-codegen@0.10`.

Palleas:
That make sense, I can switch to npm! That said, I’d recommend using a package.json (especially because people will need npm anyway), that way there is no doc to update when you manage to upgrade Apollo. Let me know what you think! :)

(Contributing to this app using this app is **awesome**)

Sherlouk:
I'm definitely all up for a package.json, at least the steps will be down to just `npm install` and will work for future dependencies (if we get any) and as you say saves us when we update versions!

Can also tie this to #387 to make buddybuild play nicely?

Palleas:
@Sherlouk Done!

rnystrom:
Awesome!!

PR #409, opened 2017-09-24
Add node bin to path before looking for apollo

Palleas:
I just realized the Apollo bass script doesn't handle local binaries, so we need to add this folder to the $PATH env var.


Palleas:
The diff isn't super obvious, so: 

screen shot 2017-09-24 at 9 57 07 am

Issue #280, opened 2017-09-04
Repo Cleanup


Sherlouk:
Root directory has gotten pretty messy with all the playgrounds, resources and other files being added

Think it's time we gave it a little clean up (moving playgrounds into their own file, etc)?
Maybe also use local Podspecs to reference the extra dependencies (Not a major issue at the moment, but I feel it's better if we have one approach to bringing in dependencies)

Can probably get rid of the culprits.txt now, or regen it and move it?
Move README assets into own file?

agisilaos:
I had the same idea yesterday night, now the root directory, feels a bit overwhelming.

rnystrom:
Go for it!

BasThomas:
Think we can close this, or does anybody think we need more cleanup?

Sherlouk:
Good enough for now!

PR #415, opened 2017-09-24
Initial repository spring clean

Sherlouk:
Relates to #280 

So I've ensured SwipeCellKit, SlackTextViewController and MMMarkdown all have Podspecs, I've moved them to their own directory ("Local Pods") and then reference them via our Podfile. This means our project file no longer has all of the separate projects in it!

I've also moved the various playgrounds into their own folder


rnystrom:
Note that this broke beta builds, it can't find the new frameworks. Will investigate. Another +1 to having CI setup so we can land-block this stuff.

Sherlouk:
Probably to do with the TestFlight builds, was hoping CocoaPods would auto-transfer those for us as it does for all the other pods.

Feel free to revert if you can't work it out

Issue #413, opened 2017-09-24
UISwitch colour should match API status green


dylanslewis:
Version 1.10.0 (1900)
Device: iPhone 7 Plus (iOS 11.0) 

I believe the app’s styling would feel more consistent if the tint colour of the UISwitch matched the API status “Good” green. Happy to submit a PR if you agree

![2692fffc-35c2-4111-8fbe-0064fc9dc22c](https://user-images.githubusercontent.com/8267782/30781367-a9d26c9a-a115-11e7-8220-b5adaf0de7b7.png)

rnystrom:
Lovely detail! Good eye

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

dylanslewis:
Raised PR: https://github.com/rnystrom/GitHawk/pull/421

PR #421, opened 2017-09-24
[413] Use shared colours for standard UI components

dylanslewis:
# Implements #413

## Changes
- Update `onTintColor` of all `UISwitch`es to `Green.medium`
- Update `tintColor` of all `UIAlertController`s to `Blue.medium`
- Update `preferredControlTintColor` of all `SFSafariViewController`s to `Blue.medium`

## Notes

I created an `SFSafariViewController` factory which performs the HTTP scheme check, and also applies the `preferredControlTintColor`. I’ve updated every initializer of `SFSafariViewController` to use this factory instead, so that the styling only needs to be set in one place. In doing so, I’ve had to force unwrap in `CreateProfileViewController` and `NotificationNavigation` to preserve the method signatures, but let me know if you want this handled better.

I have also created a similar factory which configures the `tintColor` for `UIAlertController`.

I’ve created these factories in their respective `UIViewController` `extension` classes since I didn’t see a clear place for them in the project structure, but this isn't ideal so let me know if you want them moved.

## Attachments

| Before | After |
| --- | --- |
| simulator screen shot - iphone 7 plus - 2017-09-24 at 21 54 01 | simulator screen shot - iphone 7 plus - 2017-09-24 at 22 15 38 |
| simulator screen shot - iphone 7 plus - 2017-09-24 at 21 23 42 | simulator screen shot - iphone 7 plus - 2017-09-24 at 21 48 31 |
| simulator screen shot - iphone 7 plus - 2017-09-24 at 20 56 33 | simulator screen shot - iphone 7 plus - 2017-09-24 at 21 06 15 |

### DIFF
--- a/Classes/Issues/IssuesViewController.swift
+++ b/Classes/Issues/IssuesViewController.swift
@@ -251,6 +251,7 @@ IssueTextActionsViewDelegate {
         alert.addAction(safariAction())
         alert.addAction(UIAlertAction(title: Strings.cancel, style: .cancel))
         alert.popoverPresentationController?.barButtonItem = sender
+		alert.view.tintColor = Styles.Colors.Blue.medium.color
         present(alert, animated: true)
     }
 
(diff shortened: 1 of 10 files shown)


rnystrom:
It's beautiful!!

PR #423, opened 2017-09-25
Fixes referenced issue view controller being pushed without nav bar

### LINKED ISSUE #388: Navigation bar missing in detail view on iPad
e68b5f05-2538-4940-af4d-ae5c54e63a8f

rizwankce:
Fixes #388 

Am I missing something here?

### DIFF
--- a/Classes/Issues/Comments/IssueCommentSectionController.swift
+++ b/Classes/Issues/Comments/IssueCommentSectionController.swift
@@ -213,7 +213,7 @@ AttributedStringViewIssueDelegate {
 
     func didTapIssue(view: AttributedStringView, issue: IssueDetailsModel) {
         let controller = IssuesViewController(client: client, model: issue)
-        viewController?.showDetailViewController(controller, sender: nil)
+        viewController?.show(controller, sender: nil)
     }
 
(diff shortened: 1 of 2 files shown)


rnystrom:
Ahhhh you’re right. Nice one.

Sherlouk:
Admittedly I haven't tried this solution, but shouldn't we be using `pushViewController` from the view controller's navigation controller?

rizwankce:
@Sherlouk tried this solution on both iPad and iPhone and it works fine. I just figured out this from that `show()` actually overrides and handle the presentation. 

From [Documentation](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621377-show)
>The UISplitViewController and UINavigationController classes override this method and handle the presentation according to their design. For example, a navigation controller overrides this method and uses it to push vc onto its navigation stack.

Sherlouk:
Well today I learn, that's pretty helpful actually 🤔 

LGTM

rizwankce:
Even I didn't know this before. After reading your comment was wondering how it was working when I tested. Then checked the documentation to understand the behaviour 👍

PR #440, opened 2017-09-27
WIP: Update Apollo

Sherlouk:
So I fixed all the compile time issues, but I broke something because every comment is the same comment (oops 😂)

PR just for comparison so I can see where I went wrong

@rnystrom I ended up updating to `0.7.0-alpha-5` in order to get the compile time fixes as well as the support for Swift 4 (which we'll soon need). I understand it's not a full release, but assuming I get everything working would we be okay with this or na?


Sherlouk:
Cool fixed that mistake! With a quick test it seems to be functional again, definitely needs some refactoring now though because a lot of it just doesn't make sense now - the way they've changed fragments makes it sorta ugly!

rnystrom:
Interesting, I’ll give this a look. The nice part about fragments is that we could actually share the code between parsers. Is that not the case anymore?

Sherlouk:
Yes and No, if you cast it out to the object using the snapshot (as I have) then you can pass around those objects actually more freely than you could before as they're no longer coupled to a specific query (Good!)

You can also access all the fields more directly as I said before, so if we can use that then we really should and preferably use protocols to cast to mutual interfaces. However, if we do need the shared fragments from Apollo then we can still use them even if it's less efficient

Sherlouk:
Worst bit about this is remembering to downgrade Apollo when you wanna work on other stuff 😂

Sherlouk:
@rnystrom Mind taking a look at this sooner rather than later; suspect if any major features go in then we'll have some annoying conflicts to fix! (think the current one just needs the project to be built and it'll fix itself)

rnystrom:
@Sherlouk sure! Still WIP or this is ready to merge?

Sherlouk:
It "works" but as per the comments in the thread, could probably do with some refactor work (probably safe to merge though and do it over time)

Also needed confirmation you were happy to be using an alpha release.

I'm up for merging though!

rnystrom:
I'll skim it over, I can prob handle any major refactoring after playing around w/ it (code review will take too long). And we can ignore the conflicts since it'll just be overwritten.

martijnwalraven:
Hi! I'm the author of Apollo iOS, so it's awesome to see you're using it for this project! 

If you're up for it, I'd love to set up a call to learn more about your experiences using Apollo iOS, and what you feel is still missing from it. I'm a little confused by the need to wrap the generated models in your own models for example, because ideally the generated code could be used directly (with the ability to write your own extensions for convenience or utility methods of course). If that doesn't work for you, maybe there are other things we can improve.

Also, I want to warn against using `snapshot` directly. These are fundamentally unsafe because they break type checking. You should be able to access fragments you defined in your query through `fragments`, and if you want to convert an object into another object you can just use a no argument constructor that actually checks whether all the fields defined on the target are there. 

(The only reason the `snapshot` constructor is public is because we need to be able to call it from framework code, which lives in a different module, since Swift doesn't have a `friend` access level. We should probably change the argument naming in the constructor to make it clear this is meant to be an internal constructor.)

rnystrom:
@martijnwalraven I’d love to! Traveling Europe this week, will be home next week. You can email rnystrom at whoisryannystrom dot com.

I’m using view models instead of the Apollo models themselves for a _bunch_ of reasons. Namely:

- Precomputing text
- Parsing markdown
- No Apollo dependency on each surface
- Custom data transforms

martijnwalraven:
Great! Where are you traveling to? I'm actually based in Amsterdam, but I'm currently at a conference in Oslo.

My idea has always been that the generated models *are* view models, because they only include the data required for a specific UI component. That is also how most React developers are using GraphQL. But I can definitely see your point. I wonder how much of this could be solved by extension methods or client-side fields (the ability to define resolvers on the client and store the values they result in together with the server values).

martijnwalraven:
@Sherlouk, would you also like to join?

Sherlouk:
@martijnwalraven  If the time and day is suitable for me (British lad) then I’ll happily jump in but don’t try and cater for me especially 🙂

martijnwalraven:
@Sherlouk: I just sent an email to Ryan proposing what would be 4pm your time next Tuesday or Wednesday. Would that work for you?

Sherlouk:
If we could push it back ~45 mins I could make that viable - Would’ve just finished work!

martijnwalraven:
That would work for me on Wednesday, but not on Tuesday. Let's see whether @rnystrom can make it then.

rnystrom:
Ok, took the time to upgrade to codegen 0.16.5 and iOS fmk 0.6.5 (both stable). At least gets us a little closer.

rnystrom:
Alrighty, finished this up and pushed! Using alpha versions of Apollo iOS and codegen, but it's building!

martijnwalraven:
@Sherlouk: We have a meeting planned for Wednesday at 5pm your time. I don't have your email so I couldn't add you to the event. Can you send me an email?

PR #475, opened 2017-09-29
Image Upload for Bug Reports

Sherlouk:
Basic version is setup now, when you click on the upload icon it'll take you to the picker. You click on the image, wait a couple seconds, and then it'll insert the markdown for you with the correct link.

Need to work on the UX/flow when a user taps on the image - Indicating that something is happening. Currently thinking to dismiss the picker view, and add a progress bar just under the navigation bar which goes across while all the background work is happening? Not sure though as the cursor could move in that time...

Open for suggestions on that front

I've also got the networking stuff ready for ensuring rate limits are not exceeded, but this isn't wired up yet. I think this'll come just before I do the upload step, but after the user has selected the image. That way it'll be the most accurate information! Purpose for this step is to prevent us from ever using 100% of our rate limit, as if we do this 5 times in a month then Imgur will not give us any more allowance for the rest of the month - I'd rather have an error saying "No more uploads today" than "No more uploads this month". Regardless I don't think we'll reach our limits, but this will be monitored and we can see how it goes! Better to be safe.

Another decision to be had is the text that get's associated with the image. We can provided a Name, Title and Description as well as the text that goes with the markdown (though I'm currently using the same value as the name). We can provide nothing for all these values, but up to us really.

I've also got pretty harsh compression on at the moment (0.2). I've tried it at 0.5 and it does still work, but using PNG representation was exceeding file size limits (granted the default images are super high-res, but got to account for these situations!). If people think the quality of image is *too* bad we can up the number until a point where we're all happy.


BasThomas:
Thanks for the detailed message description! On the imgur name / title / desc, maybe just use the title and desc of the issue (w/o the image)? 

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -225,4 +237,28 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         case .line(let left): bodyField.replace(left: left, right: nil, atLineStart: true)
         }
     }
+    
+    // MARK: UIImagePickerControllerDelegate
+    
+    func imagePickerController(_ picker: UIImagePickerController, didFinishPickingMediaWithInfo info: [String : Any]) {
+        guard let image = info[UIImagePickerControllerOriginalImage] as? UIImage else { return }
```
BasThomas: Would this allow video too? If not, let’s show it to the user? Or maybe show the picker w/o video

BasThomas:
Nice checkpoints by the way 😂

Sherlouk:
@BasThomas 

> maybe just use the title and desc of the issue

The issue is we don't know that information, the way I've designed this feature is that a user can drop the image in at any point. So they could insert the image first, middle or last - It's not a 100% that we will get the information.

Again, it's all optional so we don't **have** to put something. I think if we give it a name we could just do "GitHawk upload by BasThomas"?

> let’s not forget error handling

Definitely haven't! Just wanted to push a version with the most basic approach working. Surprisingly though when I was working on Sim (maybe it's diff) it never asked me for permissions, or required the `Info.plist` change (privacy description).

Honestly at the moment I'm considering to an entirely custom UI, just so that it fits more in line with the app. Currently you get the modal over the top, select an image and then it freezes - What I may do is something more along the lines of Slack where you tap on the image and it asks you for a title/desc, you click upload and then a progress bar?

> Nice checkpoints by the way 😂

Thanks 😂  Was having some difficulties with all the image conversion to Base64 so was trying to work out whether it was the network call or the image conversion that I was waiting for

> Would this allow video too?

Picker has video's disabled at the moment, my plan is with a version two approach to take videos and convert them maybe to a GIF format. But wanting to keep it fairly simple for now, so just images!

> Compression isn’t too bad, either

This was definitely one of my concerns, there's not really an easy way to sort of say "Compress as much as you need to until your file size is below this limit". So in theory with the current system, if you have a low-res image and upload it, it's going to come out super granulated - We will just need to play around with this once it's in TF. I feel this feature will be more commonly used for iOS screenshots anyway, as opposed to an Instagram client.

BasThomas:
> I think if we give it a name we could just do "GitHawk upload by BasThomas"?

Also for consistency, I think this is more than OK. Also good point to let the user upload whenever.

Sherlouk:
So as part of the most recent commit I've added a new upload UI (à la Slack) with the ability for the user to define a title and description. The name is "GitHawk upload by {username}".

When you enter the upload UI I automatically (in the background) compress and convert the image to a base64 image. This just makes the whole system a little bit less error prone once you get to submitting the app.

Hooked up the rate limiting request which ensures we always have enough "tokens" to actually upload something. (See root comment for reasoning)

As part of the V2 iteration, a few ideas I had:
- Support for markup on the preview image, giving the user to circle key elements etc
- Support for videos/live images as GIFs

@BasThomas If you wanna take a look, now's the time as it's basically all wired up 😃

Classes/Image Upload/ImageUploadViewController.swift
```diff
@@ -0,0 +1,214 @@
...
+    // MARK: Image Preparation
+    
+    /// Compressed and Encodes in Base64 the provided UIImage.
+    ///
+    /// Process is moved to a background thread in order to prevent UI blocking.
+    ///
+    /// Compression is a value between 0.0 and 1.0. Lower is smaller file size but worse quality.
+    private func compressAndEncodeImage(_ image: UIImage, compression: CGFloat = 0.2, completion: @escaping (Result<String>) -> Void) {
+        DispatchQueue.global(qos: .background).async {
+            let data = UIImageJPEGRepresentation(image, compression)
```
weyert: Are you sure this shroud be `UIImageJPEGRepresentation`?

Classes/Image Upload/ImageUploadViewController.swift
```diff
@@ -0,0 +1,214 @@
...
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        // Set the preview image
+        previewImageView.image = image
+        
+        // Set title field delegate so return moves to next field
+        titleTextField.delegate = self
+        
+        // Add borders to imitate UITableView grouped design
```
weyert: Why not a `UITableView` this time? 🤓

Sherlouk: Thought I fixed this 😂 It's meant to be a table view! Oops!

Classes/Image Upload/ImageUploadViewController.swift
Sherlouk: As opposed to?

I need to convert a UIImage to a Base64 string, if there's a better way of doing this then shout!

weyert: I misread your initial comment. I thought it meant to be PNG instead of JPG hence my comment. Sorry!

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -197,6 +202,15 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
             IssueTextActionOperation(icon: UIImage(named: "bar-ul"), operation: .line("- ")),
             IssueTextActionOperation(icon: UIImage(named: "bar-indent"), operation: .line("  ")),
             IssueTextActionOperation(icon: UIImage(named: "bar-link"), operation: .wrap("[", "](\(UITextView.cursorToken))")),
+            IssueTextActionOperation(icon: UIImage(named: "cloud-upload"), operation: .execute({ [weak self] in
+                guard let strongSelf = self else { return }
+                
+                let imagePicker = UIImagePickerController()
+                imagePicker.delegate = strongSelf
+                imagePicker.modalPresentationStyle = .formSheet
```
rizwankce: @Sherlouk One quick suggestion here for iPad,
New issue view controller is already presented as form sheet, so if we present image picker also a form sheet it might not look good.

Also apple [suggests](https://developer.apple.com/documentation/uikit/uiimagepickercontroller) to use `UIImagePicketController` as `popover` if we target iPad

Sherlouk: Will investigate further, it looked fine when I was testing but didn't get `popover` an attempt

Sherlouk:
@rnystrom Ready for a full review on this when you're ready 😄 Would love to get this in as it unlocks a number of things!

Classes/New Issue/NewIssueTableViewController.swift
Sherlouk: Updated to use popover, honestly I think it looks worse than form sheet because there isn't anything visible to attach it to. It's also too small.

Still think formsheet looks better, easy switch back though

Classes/Image Upload/ImageUploadViewController.swift
Sherlouk: Fixed!

Sherlouk:
😢 So lonely in here

Classes/Image Upload/ImageUploadTableViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+                    title: self?.titleText ?? "",
+                    description: self?.descriptionText ?? "") { [weak self] result in
+                        
+                        // UI Work, so ensure it's on the main thread
+                        DispatchQueue.main.async {
+                            switch result {
+                            case .error:
+                                ToastManager.showGenericError()
+                                self?.setRightBarItemIdle()
+                                return
```
BasThomas: Why the `return` here?

Classes/Image Upload/ImageUploadTableViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+    }
+    
+    // MARK: Image Preparation
+    
+    /// Compressed and Encodes in Base64 the provided UIImage.
+    ///
+    /// Process is moved to a background thread in order to prevent UI blocking.
+    ///
+    /// Compression is a value between 0.0 and 1.0. Lower is smaller file size but worse quality.
+    private func compressAndEncodeImage(_ image: UIImage, compression: CGFloat = 0.2, completion: @escaping (Result<String>) -> Void) {
```
BasThomas: `compressAndEncode(image: UIImage ...`?

BasThomas: Or a private extension on UIImage here? Nbd as it's just used here, but hey.

Classes/Image Upload/ImgurClient.swift
```diff
@@ -0,0 +1,78 @@
...
+            
+            // Takes 10 tokens to upload an image, a buffer has been added to prevent us using 100% of our allowance as this
+            // will mean our app gets temporarily blocked from Imgur!
+            completion(userRemaining > 20 && clientRemaining > 100)
+        }
+    }
+    
+    func uploadImage(base64: String, name: String, title: String, description: String, completion: @escaping (Result<String>) -> Void) {
+        let params = [
+            "image": base64, // Base64 version of the provided image
```
BasThomas: As you felt the need to explain the var here, maybe rename the `base64` arg to `base64Image`?

Classes/Image Upload/ImgurClient.swift
```diff
@@ -0,0 +1,78 @@
...
+            "type": "base64",
+            "name": name,
+            "title": title,
+            "description": description
+        ]
+        
+        request("image", method: .post, parameters: params) { response in
+            guard let dict = response.value as? [String: Any], let data = dict["data"] as? [String: Any] else {
+                completion(.error(nil))
```
BasThomas: Why not pass this `response.error`?

Classes/Image Upload/ImgurClient.swift
```diff
@@ -0,0 +1,78 @@
...
+                return
+            }
+
+            guard let link = data["link"] as? String else {
+                completion(.error(nil))
```
BasThomas: Same as above

Classes/Other/Secrets.swift
```diff
@@ -14,3 +14,9 @@ public enum GithubAPI {
     static let clientSecret = "YOUR_CLIENT_SECRET"
 
 }
+
+public enum ImgurAPI {
+    
+    static let clientID = "<# Imgur API client ID #>"
```
BasThomas: Why the `<# #>`? Would just keep this in line with the others (so `IMGUR_API_CLIENT_ID`)

Classes/Views/IssueTextActionsView+Markdown.swift
```diff
@@ -31,7 +31,24 @@ extension IssueTextActionsView {
             IssueTextActionOperation(icon: UIImage(named: "bar-ul"), operation: .line("- ")),
             IssueTextActionOperation(icon: UIImage(named: "bar-indent"), operation: .line("  ")),
             IssueTextActionOperation(icon: UIImage(named: "bar-link"), operation: .wrap("[", "](\(UITextView.cursorToken))")),
-            ]
+            IssueTextActionOperation(icon: UIImage(named: "cloud-upload"), operation: .execute({ [weak viewController] in
+                guard let strongVc = viewController else { return }
+                
+                let imagePicker = UIImagePickerController()
+                imagePicker.delegate = strongVc as? (UINavigationControllerDelegate & UIImagePickerControllerDelegate)
```
BasThomas: Neat!

Classes/Image Upload/ImgurClient.swift
Sherlouk: That was less the need to explain the var, more the fact I didn't delete the comment after I had empty implementations mid-dev and needed to remind myself what each value was 😅  Will rename though

Classes/Image Upload/ImageUploadTableViewController.swift
Sherlouk: Bad habits

Classes/Other/Secrets.swift
Sherlouk: More as a forceful reminder for Ryan to populate the value when testing/releasing

Classes/Image Upload/ImgurClient.swift
Sherlouk: I can do, but the actual error is never used - just the fact that an error state was returned in order to reset the view etc

BasThomas: Especially since the error is already an optional, I'd return it in the completion and leave the implementation up to the caller.

Classes/Other/Secrets.swift
BasThomas: Hm, but the build doesn't complain even with the placeholder there. Which made me think - maybe we can print some warnings in debug mode when these are set to default?

Sherlouk: Huh!? Xcode 9 has always given me a compile error if there's a placeholder in code... I can give it a default value I'm not fussed - was intended to just be a reminder!

BasThomas: Thought so as well, but seems to work fine. Maybe because it's in a `String`?

BasThomas: Yep, `"<# test #>"` is just a string containing `"<# test #>"` at compile time it seems.

Sherlouk: Maybe 🤔 Today I learn

Sherlouk:
Also fixed the build issues that are in master as part of this, for some reason a bunch of classes were added to the test target where we should be using `@testable` instead.

BasThomas:
Should we use `@testable` over adding classes to the test target? I just followed what @rnystrom was already doing.

Sherlouk:
I'd always vote for `@testable` over adding to targets. AFAIK it's only as part of Swift (maybe why Ryan defaulted to adding to target).

It's also better for Swift access control as it enables you to test internal methods as if they were public. Useful for when you separate out into multiple frameworks which we have already said we intend to do.

I personally find adding classes to a target as a very fragile way, as (proven once again here) it's extremely easy to forget to target classes. `@testable` just says "I want to test this module" and lets you get on with it.

BasThomas:
I agree on using `@testable`, definitely. Just curious to see if there was a reason behind **not** doing it. :)

Classes/Views/IssueTextActionsView+Markdown.swift
rnystrom: @Sherlouk should we update the `viewController` param type so this is explicit? Otherwise this will silently break, right?

Classes/Views/IssueTextActionsView+Markdown.swift
```diff
@@ -31,7 +31,24 @@ extension IssueTextActionsView {
...
+                imagePicker.modalPresentationStyle = .popover
+                imagePicker.popoverPresentationController?.canOverlapSourceViewRect = true
+                
+                let sourceFrame = viewController?.view.frame ?? .zero
```
rnystrom: `strongVc`?

Classes/Views/IssueTextActionsView+Markdown.swift
```diff
@@ -31,7 +31,24 @@ extension IssueTextActionsView {
...
+                let sourceRect = CGRect(origin: CGPoint(x: sourceFrame.midX, y: sourceFrame.minY), size: CGSize(width: 1, height: 1))
+                imagePicker.popoverPresentationController?.sourceView = viewController?.view
```
rnystrom: `strongVc.view`

Classes/Utility/Image+Base64.swift
```diff
@@ -0,0 +1,31 @@
...
+import UIKit
+
+extension UIImage {
+    
+    /// Compressed and Encodes in Base64 the given image.
+    ///
+    /// Process is moved to a background thread in order to prevent UI blocking.
+    ///
+    /// Compression is a value between 0.0 and 1.0. Lower is smaller file size but worse quality.
+    func compressAndEncode(compression: CGFloat = 0.2, completion: @escaping (Result<String>) -> Void) {
```
rnystrom: Hot tip that `0.65` is a great compression v quality ratio. Much experience here 😉

Classes/Image Upload/ImageUploadTableViewController.swift
```diff
@@ -0,0 +1,193 @@
...
+            }
+            
+            var name = "GitHawk Upload"
+            
+            if let username = self?.username {
+                name += " by \(username)"
+            }
+            
+            // Ensure the upload step is on the background thread
+            DispatchQueue.global(qos: .userInitiated).async {
```
rnystrom: Shouldn't the `request(...)` already happen in the background? This isn't necessary, right? At the very least, we should hide the async inside `uploadImage(...)`

Classes/Image Upload/ImageUploadTableViewController.swift
```diff
@@ -0,0 +1,193 @@
...
+                self?.client.uploadImage(
+                    base64Image: compressionData,
+                    name: name,
+                    title: self?.titleText ?? "",
+                    description: self?.descriptionText ?? "") { [weak self] result in
```
rnystrom: Accessing UI stuff while on a bg queue ⚠️ ☠️ 👎

Classes/Image Upload/ImageUploadTableViewController.swift
```diff
@@ -0,0 +1,193 @@
...
+                        
+                        // UI Work, so ensure it's on the main thread
+                        DispatchQueue.main.async {
```
rnystrom: Again let's hide the dispatch inside the `uploadImage` API, if its even necessary. The networking API already dispatches to main, so I don't think this is required.

Classes/Image Upload/ImgurClient.swift
```diff
@@ -0,0 +1,83 @@
...
+
+final class ImgurClient {
+    
+    enum ImgurError: Error {
+        // missingLink: - Received a valid response, but no link was available in the payload
+        case missingLink
+    }
+    
+    static let hostpath = "https://api.imgur.com/3/"
+    static let headers: HTTPHeaders = ["Authorization": "Client-ID \(ImgurAPI.clientID)"]
```
rnystrom: `private`?

Classes/Image Upload/ImgurClient.swift
```diff
@@ -0,0 +1,83 @@
...
+            }
+            
+            guard let userRemaining = data["UserRemaining"] as? Int, let clientRemaining = data["ClientRemaining"] as? Int else {
+                completion(false)
+                return
+            }
+            
+            // Takes 10 tokens to upload an image, a buffer has been added to prevent us using 100% of our allowance as this
+            // will mean our app gets temporarily blocked from Imgur!
+            completion(userRemaining > 20 && clientRemaining > 100)
```
rnystrom: Awesome!

Classes/Image Upload/ImgurClient.swift
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+    
+    func uploadImage(base64Image: String, name: String, title: String, description: String, completion: @escaping (Result<String>) -> Void) {
```
rnystrom: Mind putting each param on its own line? Easier to read + review (& consistent w/ rest of app)

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -198,4 +202,25 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
...
+    
+    func imagePickerControllerDidCancel(_ picker: UIImagePickerController) {
+        dismiss(animated: true, completion: nil)
+    }
+    
+    // MARK: ImageUploadDelegate
+    
+    func imageUploaded(link: String, altText: String) {
+        bodyField.replace(left: "![\(altText)](\(link))", right: nil, atLineStart: true)
+    }
```
rnystrom: Any change we can put this in something that can be shared elsewhere? Would be awesome to have this in the commenting eventually too!

Classes/Views/IssueTextActionsView+Markdown.swift
```diff
@@ -31,7 +31,24 @@ extension IssueTextActionsView {
             IssueTextActionOperation(icon: UIImage(named: "bar-ul"), operation: .line("- ")),
             IssueTextActionOperation(icon: UIImage(named: "bar-indent"), operation: .line("  ")),
             IssueTextActionOperation(icon: UIImage(named: "bar-link"), operation: .wrap("[", "](\(UITextView.cursorToken))")),
-            ]
+            IssueTextActionOperation(icon: UIImage(named: "cloud-upload"), operation: .execute({ [weak viewController] in
```
rnystrom: Doesn't this show up in `IssueViewController`? But it doesn't conform to the delegate or implement any of the delegate stuff to handle succesful upload. So wont the button show up but do nothing?

Classes/Utility/Image+Base64.swift
Sherlouk: We can make the make the number bigger here, by all means - the issue is the Imgur API has a maximum file size. So if the number is too big (I had it fail with a 0.8) then the image will just fail. 0.2 is probably a bit too extreme though 😅

Classes/Image Upload/ImageUploadTableViewController.swift
Sherlouk: Probably leftover from when I had the compression task just before the network request. Will neaten.

Classes/Views/IssueTextActionsView+Markdown.swift
Sherlouk: Yeah my mistake, was building this for bug report only originally but moved it after a merge conflict 😅 Will fix

Sherlouk:
Opted to move the image upload stuff to the controller that's shared by everything. I've also added a parameter to the initial setup function which decides whether or not we support upload. I thought while we were testing the feature we should just use it in bug reports and then we could roll it out to other repos at a later date once we're confident with it?

Changed threading etc and reduced the level of compression

PR #483, opened 2017-10-02
Replaced status bar error with Toaster

kimdv:
I think it looks pretty 😄 

simulator screen shot - iphone 8 plus - 2017-10-02 at 10 41 30

Maybe there could be some more layout changes.

What do you think?


Sherlouk:
Thanks for picking this up! Haven't had a chance to fully look at this yet but one thing I immediately noticed is we still have `JDStatusBarNotification` in the Podfile - Could we remove that as well as part of this?

Also is the only colour option red here, or can we have contextual colours based on severity? Errors are a red, for example?

kimdv:
@Sherlouk I have removed `JDStatusBarNotification ` now. 

I have also made some refactor, so we dont repeat our self and renamed the `Statusbar` class to `Toaster`

Is is posible to change the background color. Thus yes, we can make it context based. For now I have made them red as default.

Classes/Other/Secrets.swift
```diff
@@ -10,7 +10,7 @@ import Foundation
 
 public enum GithubAPI {
 
-    static let clientID = "YOUR_CLIENT_ID"
-    static let clientSecret = "YOUR_CLIENT_SECRET"
+    static let clientID = "d611a03290c738ba4f3f"
```
Sherlouk: Can you remove your GitHub API details please! (We really should add this to the `.gitignore` to stop people making this mistake 😰)

kimdv: oooh 😰😰😰

Remove from Github or Project? 

Sorry about that! 
But yes, we should add it to .gitignore

Sherlouk: Just from the project, leaving it unchanged from where it was -- Up to you if you want to regenerate your Oauth app to receive new ID/secrets just in case.

kimdv: It should be changed back!

BasThomas:
I'm not sure about the toaster overlapping with the tab bar, how would it look / could it work when it is shown just on top?

kimdv:
simulator screen shot - iphone 8 - 2017-10-02 at 13 39 42

@BasThomas I have moved it a little up

Classes/Systems/Toaster.swift
```diff
@@ -0,0 +1,49 @@
...
+    }
+    
+    static func showError(message: String,
+                          textColor: UIColor = UIColor.white,
+                          backgroundColor: UIColor = Styles.Colors.Red.medium.color ) {
+        let toast = Toast(text: message, duration: Delay.long)
+
+        toast.view.backgroundColor = backgroundColor
+        toast.view.textColor = textColor
+        toast.view.bottomOffsetPortrait = 60
```
rnystrom: Can you run this in the iPhone X simulator and post how it looks? Wonder if we should use an inset API for this instead?

kimdv:
@rnystrom I ran it on the iPhone X, and the Toaster was in the middle of the tabbar. 

But I fixed it with the SafeArea. 

simulator screen shot - iphone x - 2017-10-03 at 08 03 25
simulator screen shot - iphone 8 - 2017-10-03 at 08 09 30

Classes/Systems/Toaster.swift
```diff
@@ -0,0 +1,55 @@
...
+    }
+    
+    static func showError(message: String,
+                          textColor: UIColor = UIColor.white,
+                          backgroundColor: UIColor = Styles.Colors.Red.medium.color ) {
+        let toast = Toast(text: message, duration: Delay.long)
+        var bottomInset: CGFloat = 0.0
+
+        if #available(iOS 11, *) {
+            let insets = UIApplication.shared.delegate?.window??.safeAreaInsets
```
BasThomas: What's with the double `??` here?

kimdv: https://stackoverflow.com/questions/28901893/why-is-main-window-of-type-double-optional

It is some sort of double optional 🤔🤔🤔

BasThomas: Oh really? Thought it was just an optional. Makes sense 👍

Classes/Systems/Toaster.swift
```diff
@@ -0,0 +1,55 @@
...
+
+            bottomInset = insets?.bottom ?? 0
+        }
```
rnystrom: We need to handle iOS <11 in the `else` right?

kimdv: The only reason I use the if available is to get the iPhone X safe area. 

If it is a iPhone 8 or 7 etc the insets?.bottom will be 0. So the same as pre iOS 11

rnystrom:
😬 I might have an idea for building our own version of this with a few more customizations. I’d love to add an info “i” button to the right. Tapping it presents a VC with error details VC. That VC lets you highlight the error text (text view) and also has a “new bug report” button. Lets us capture bad requests maybe?

rnystrom:
Could also do some fun stuff like UIDynamics and swipe to dismiss the toast.

kimdv:
We can copy the `Toaster` codebase into the GitHawk project and continue from there. 

Because should the user have the ability to report a bug because the revoked their token etc? 😅

rnystrom:
Ya! Or weird parsing errors, etc. at the very least might help me to notice issues reported for GH service outages or something.

rnystrom:
Also prob easier to start from scratch instead of fork. I find you end up knowing the thing end to end, and learn stuff along the way.

kimdv:
Okay, I will close this PR then and make a new issue where we can discuss the GitHawk Toast 🚀

BasThomas:
*Maybe*, just *maybe* start thinking about a GitHawk org  where the Toaster code can be separated?

rnystrom:
Eventually, don’t wanna bog down progress with red tape _quite_ yet 😅

BasThomas:
Think it would be good to keep in mind while working on that, though!

PR #494, opened 2017-10-05
🌹 Add blog URL to settings page

### LINKED ISSUE #437: Add URL to app to blog site

viktorgardart:
~Fixes #437~

Edit: Doesn't actually solve #437

### DIFF
--- a/Classes/Settings/SettingsViewController.swift
+++ b/Classes/Settings/SettingsViewController.swift
@@ -22,6 +22,7 @@ NewIssueTableViewControllerDelegate {
     @IBOutlet weak var reviewAccessCell: StyledTableCell!
     @IBOutlet weak var reportBugCell: StyledTableCell!
     @IBOutlet weak var viewSourceCell: StyledTableCell!
+    @IBOutlet weak var viewBlogCell: StyledTableCell!
     @IBOutlet weak var signOutCell: StyledTableCell!
     @IBOutlet weak var backgroundFetchSwitch: UISwitch!
     @IBOutlet weak var openSettingsButton: UIButton!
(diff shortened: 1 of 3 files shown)


rnystrom:
Lovely!

BasThomas:
Oh ha, meant "add a url to the app on blog.githawk.com", but this is also nice! 😀

Classes/Settings/SettingsViewController.swift
```diff
@@ -115,6 +118,12 @@ NewIssueTableViewControllerDelegate {
 		presentSafari(url: url)
     }
 
+    func onViewBlog() {
+        guard let url = URL(string: "http://blog.githawk.com/")
```
Sherlouk: This is less a note with this PR but more a general statement, are we okay with generic strings like this being placed throughout the codebase? Or would we rather move them to either a constants file, or maybe the Info.plist under it's own dictionary? Easier to manage long-term?

viktorgardart: This the best shot would be to store them in a constants file. And then call it by: `Constants.url.blog` or something similar

BasThomas: We already have [Strings.swift](https://github.com/rnystrom/GitHawk/blob/master/Classes/Views/Strings.swift), maybe that works for now as well?

viktorgardart: Created a new issue for this:

https://github.com/rnystrom/GitHawk/issues/498

BasThomas: Awesome!

rnystrom:
Dang, sorry bug I'm actually going to back out of this. The more I thought about it, it might end up just bloating the settings. I want to keep them as light as possible for now. Will revive if the need comes up!

BasThomas:
Think at some point we should improve there. Think it can be better even at this point. Maybe declutter by having eg. the switches in a new vc, like accounts?

Issue #420, opened 2017-09-24
Update GitHub settings so auth makes sense


rnystrom:
https://twitter.com/hansemannnn/status/912048068698234881

Sherlouk:
Ha just mentioned this in #418 but yeah got to make sure we release and update on GitHub at same time otherwise people won't be able to login 😱

BasThomas:
Also, we should update the docs

BasThomas:
@Sherlouk that’s a tricky one then. Would break all previous versions of the app... if we want to go that route, we might want an intermediate release so we can then show a dialog asking users to update again? Not sure if that would help / make sense.

rnystrom:
Ya I don’t want to do break any backwards compatibility. I haven’t explored what display options I have on the GitHub app yet either. At the very least I could change the callback url.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Sherlouk:
The only way to have complete backwards compatibility is to keep it as `freetime://`, obviously we could do forced upgrade but I'm not always a fan

My thoughts behind my original comment is that a user of the app will "most likely" have logged in. Which means the only people who are going to have issues, are new people trying to log in who "should" be on the latest version as they're new! By that logic, if we push the app out first with the new callback URL then it shouldn't be too bad in terms of breaking for users.

Obviously the easier solution here is just make another Oauth application on GitHub with the URL set to `githawk://` and in the new release just update everything to reference the new scheme and the new oauth application. Seamless transition, in a couple weeks/months just delete the old application on GitHub (if you care)

rnystrom:
Ya, I think just creating a new Oauth app is the way to go. Should I create a GitHawk-App org/user so the text reads as "Authorize GitHawk-App"?

Sherlouk:
Sure!

gingerbeardman:
Why does CodeHub not have the auth issues GitHawk has? Since the privacy change that came with the Microsoft buyout, I'm able to interact with less and less repos

rnystrom:
@gingerbeardman the name issue?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

gingerbeardman:
Basically I have to request reach repo to allow GitHawk client permission for me to comment or like

gingerbeardman:
Basically this for every repo @rnystrom https://github.com/GitHawkApp/GitHawk/issues/1858

rnystrom:
[...]

PR #501, opened 2017-10-07
Added SFAuthenticationSession for iOS 11 login

### LINKED ISSUE #426: iOS 11 should use SFAuthenticationSession for login
The changes to `SFSafariViewController` now mean that you are stuck w/ w/e user you logged in as. For instance, my simulator is now stuck logged in as my test user. If I log out, the safari VC is already auth'd to GitHub w/ the test user so it immediately redirects back to the app and logs in. Even if I open Safari.app and log out of everything.

joerocca:
Hello! 

fixes #426 

Something to note here is that the content dialog shows the project name "Freetime".

simulator screen shot - iphone 8 - 2017-10-07 at 02 30 48


Classes/Login/LoginSplashViewController.swift
```diff
@@ -24,6 +25,17 @@ final class LoginSplashViewController: UIViewController, GithubSessionListener {
     @IBOutlet weak var activityIndicator: UIActivityIndicatorView!
     private weak var safariController: SFSafariViewController? = nil
 
+    @available(iOS 11.0, *)
+    private var authSession: SFAuthenticationSession? {
+        get {
+            return _authSession as? SFAuthenticationSession
```
Sherlouk: Sorry I'm probably just being inept, why are we having to use this getter/setter pattern rather than just using the var `authSession` in the first place?

rnystrom:
Oof ya we should be able to change that easily. This probably displays for any permissions dialogs too. Maybe when you enable badging for the first time?

rnystrom:
@joerocca thank you for this!! Will review and merge ASAP. Will help immensely with local testing on Xcode 9 too!

Classes/Login/LoginSplashViewController.swift
joerocca: @available can not be used on stored properties. This solution allows the type to be defined.

I figured it is better than having to type cast.

```swift
(self?.authsession as! SFAuthenticationSession).start()
```

Another solution would be using a lazy var, but I thought the get/set pattern was most readable.

```swift
@available(iOS 11.0, *)
lazy var authSession: SFAuthenticationSession? = { return nil }()
```

joerocca:
No problem!
It looks like "GitHawk" is used when enabling badging.

joerocca:
Another problem here is AfAuthenticationSession does not share the session with SFSafariViewController, so when   I press “Open in Safari” on a private repo I get a 404 page and have to sign in inside the SFSafariViewController before I can view the page. Wasn’t able to find a solution for this with SFSafariViewController. Maybe possible to solve this by using a WKWebView and injecting the cookie?

BasThomas:
> Something to note here is that the content dialog shows the project name "Freetime".

Linking to #420 as well since that's a related issue.

rnystrom:
Oops thought I merged this. Def need to send this in to a TestFlight build.

Issue #342, opened 2017-09-21
3d touch shortcuts 


rnystrom:
Anything cool we could do? Random ideas:

- [x] Switching accounts
- [ ] “new issue” on last viewed repo
- [x] straight to search

rnystrom:
Also maybe jump to the latest 2 notifications (ala Spotify)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BasThomas:
Damn it.

BasThomas:
On Spotify: I am a frequent user of 3D Touch and Quick actions, but practically never use Spotify’s

rnystrom:
@BasThomas can you clarify a little? You never use Spotify's shortcuts?

BasThomas:
Yeah, never use their 3D Touch Quick Actions from the homescreen

BasThomas:
Reopening as we don’t have the new issue on last viewed repo yet

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BasThomas:
Guess I am going to close this after all; lets pick this back up once needed.

Issue #418, opened 2017-09-24
URL Schemes & Opener Support


Sherlouk:
Opener is an iOS app which basically lets the user choose where to open links, basically takes advantage of URL schemes.

We should add support for basic URL schemes to deeplink to certain UI within the app. Namely:

freetime://issue/rnystrom/GitHawk/418 would go to this issue!
freetime://repository/rnystrom/GitHawk
freetime://notifications

(Also aware that we may want to rename our URL Scheme??)

Once we add this feature, we should add our app to the [Opener manifest](https://github.com/timonus/OpenerManifest). It's not a perfect solution, but it would be a nice companion app to start opening GitHub stuff in our app rather than Safari!

Sherlouk:
Also worth noting that this will go slightly hand-in-hand with any work on home screen widgets, apple watch stuff, etc hence why I've given it a mid priority and not low!

![](https://media.giphy.com/media/GZc9TILZ8pax2/giphy.gif)

BasThomas:
I want to take a look at this soon :)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

timonus:
Would happily integrate into Opener once there’s a URL scheme 😊

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rnystrom:
URL scheme architecture is on my mind for the next release!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BasThomas:
Somebody could take a look at this now we have routing in place :)

palmin:
I plan to link to repositories and issues in GitHawk as soon as the App Store version of GitHawk supports this. 

![image](https://user-images.githubusercontent.com/1230554/47716297-9032ca00-dc42-11e8-879e-d2579b42232d.png)

palmin:
@rnystrom: Is there anything I can do to assist?

I'm not sure if the url format representing routes is going to change and I'm also not able to tell if the app currently looks at URL's when launched.

But if you give me some specific tasks I can move this forward.

rnystrom:
@palmin the App Store version should be able to handle deep links now using the route scheme in GitHawkRoutes

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

palmin:
@rnystrom I would expect 
  `freetime://RepoRoute?owner=GitHawkApp&branch=master&repo=GitHawk`
[...]

PR #502, opened 2017-10-08
3D Touch shortcuts

viktorgardart:
Added shortcuts for:
* Search
* Switch accounts

**Note:**
I haven't tried out the switch account item since i only have one account so @rnystrom please try that first. 

#342

### DIFF
--- a/Classes/Systems/ShortcutHandler.swift
+++ b/Classes/Systems/ShortcutHandler.swift
@@ -1,0 +1,71 @@
+//
+//  ShortcutHandler.swift
+//  Freetime
+//
+//  Created by Viktor Gardart on 2017-10-08.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+struct ShortcutHandler {
+
+    private enum Items: String {
+        case search
+        case switchAccount
+    }
+
+    static func configure(application: UIApplication, sessionManager: GithubSessionManager) {
+        application.shortcutItems = generateItems(sessionManager: sessionManager)
+    }
+
+    static func handle(shortcutItem item: UIApplicationShortcutItem,
+                       sessionManager: GithubSessionManager,
+                       navigationManager: RootNavigationManager) -> Bool {
+        let itemType = Items(rawValue: item.type)
+        switch itemType {
+        case .some(.search):
+            navigationManager.selectViewController(atIndex: 1)
+            break
+        case .some(.switchAccount):
+            if let index = item.userInfo?["sessionIndex"] as? Int {
+                let session = sessionManager.userSessions[index]
+                sessionManager.focus(session, dismiss: false)
+            }
+            break
+        default:
+            return false
+        }
+        return true
+    }
+
+    private static func generateItems(sessionManager: GithubSessionManager) -> [UIApplicationShortcutItem] {
+        var items: [UIApplicationShortcutItem] = []
+
+        // Search
+        let searchIcon = UIApplicationShortcutIcon(templateImageName: "search")
+        let searchItem = UIApplicationShortcutItem(type: Items.search.rawValue,
+                                                   localizedTitle: NSLocalizedString("Search", comment: ""),
+                                                   localizedSubtitle: nil,
+                                                   icon: searchIcon,
+                                                   userInfo: nil)
+        items.append(searchItem)
+
+        // Switchuser
+        if sessionManager.userSessions.count >= 2 {
+            let userSession = sessionManager.userSessions[1]
+            if let username = userSession.username {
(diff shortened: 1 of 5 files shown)


Classes/Systems/ShortcutHandler.swift
```diff
@@ -0,0 +1,71 @@
...
+        application.shortcutItems = generateItems(sessionManager: sessionManager)
+    }
+
+    static func handle(shortcutItem item: UIApplicationShortcutItem,
+                       sessionManager: GithubSessionManager,
+                       navigationManager: RootNavigationManager) -> Bool {
+        let itemType = Items(rawValue: item.type)
+        switch itemType {
+        case .some(.search):
+            navigationManager.selectViewController(atIndex: 1)
```
viktorgardart: Maybe change this later when we support urlschemes, `githawk://search` or something.

BasThomas: Maybe add a constant referencing search is at index 1 atm?

Classes/Systems/ShortcutHandler.swift
```diff
@@ -0,0 +1,71 @@
...
+//  ShortcutHandler.swift
+//  Freetime
+//
+//  Created by Viktor Gardart on 2017-10-08.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+struct ShortcutHandler {
```
BasThomas: Like this!

Classes/Systems/ShortcutHandler.swift
```diff
@@ -0,0 +1,71 @@
...
+        case .some(.switchAccount):
+            if let index = item.userInfo?["sessionIndex"] as? Int {
+                let session = sessionManager.userSessions[index]
+                sessionManager.focus(session, dismiss: false)
+            }
+            break
+        default:
+            return false
+        }
+        return true
```
BasThomas: I feel this is not as readable as it could be. Would prefer two `return`s instead of two `break`s. Also, do we silently want to ignore the `default` case?

Classes/Systems/ShortcutHandler.swift
```diff
@@ -0,0 +1,71 @@
...
+        let searchItem = UIApplicationShortcutItem(type: Items.search.rawValue,
+                                                   localizedTitle: NSLocalizedString("Search", comment: ""),
+                                                   localizedSubtitle: nil,
+                                                   icon: searchIcon,
+                                                   userInfo: nil)
+        items.append(searchItem)
+
+        // Switchuser
+        if sessionManager.userSessions.count >= 2 {
+            let userSession = sessionManager.userSessions[1]
```
BasThomas: Nice!

Classes/Systems/ShortcutHandler.swift
viktorgardart: Can fix the two `return`, 
About the `default` case: that wont probably happen and if it does, what should we do? Since we add the `Items` in `generateItems` we have full control over what actions will be sent.

BasThomas: It's a tricky one, yeah - `fatalError`ing would probably not be what we want to do. Let's leave it as is for now 👍

viktorgardart: Once we get the `toaster` in place we could probably do something with that.

Classes/Systems/ShortcutHandler.swift
```diff
@@ -0,0 +1,74 @@
...
+    static func configure(application: UIApplication, sessionManager: GithubSessionManager) {
+        application.shortcutItems = generateItems(sessionManager: sessionManager)
+    }
+
+    static func handle(shortcutItem item: UIApplicationShortcutItem,
+                       sessionManager: GithubSessionManager,
+                       navigationManager: RootNavigationManager) -> Bool {
+        let itemType = Items(rawValue: item.type)
+        switch itemType {
+        case .some(.search):
```
Sherlouk: Could we maybe change the `let itemType ...` to be a guard, and the switch reference the enum directly and not through the optional `.some` value?

viktorgardart: Sure!

viktorgardart: Done! 🙂 👍

BasThomas:
Want to look into #418 before merging this, @Sherlouk?

Sherlouk:
@BasThomas Honestly forgot I assigned that to myself, anyone is welcome to go for that one!

It would be nice to merge the two though!

rnystrom:
This is really well done 😊

PR #505, opened 2017-10-08
Fixed a bunch of compiler warnings

weyert:


Classes/Views/UIColor+Hex.swift
```diff
@@ -18,7 +18,7 @@ extension UIColor {
             cString.remove(at: cString.startIndex)
         }
 
-        if ((cString.characters.count) != 6) {
+        if ((cString.count) != 6) {
```
Sherlouk: `if cString.count != 6 {`?

BasThomas: Can clean up all of them here I suppose. Think Ryan coped this straight from SO 😜

Sherlouk: Noo Ryan wouldn't do that... would he?? 😉

BasThomas: So: omit `:String` in line 15
Omit `()` in 17
Omit `(())` in 21
Omit `UIColor` in 22.

weyert:
Updated!

PR #503, opened 2017-10-08
Search for Inbox

weyert:
Search for Inbox,  ticket #485

PR #506, opened 2017-10-08
Add search to issues

weyert:
Fixes #503 

After the initial pull request is accepted we could continue with this one

@rnystrom I didn't know how to force a refresh of CV easily without passing details so I added a method `performUpdate` to the base view controller

@Sherlouk here you go ;)


Classes/Repository/RepositoryIssuesViewController.swift
```diff
@@ -49,7 +51,34 @@ class RepositoryIssuesViewController: BaseListViewController<NSString>, BaseList
         }
     }
 
-    // MARK: Overides
+    // MARK: Issue filtering
+
+    func filter(_ items: [ListDiffable], _ query: String) -> [ListDiffable] {
+        guard !query.isEmpty else {  return items }
+
+        let searchText = query.lowercased()
```
Sherlouk: As per https://github.com/rnystrom/GitHawk/pull/503#issuecomment-335028446

We could split the query by space, and remove any that start with `is:` and add a flags enum (array) then in the filter group we can do something like:

`if flags.contains(.open) && item.closed { return false }`
`if flags.contains(.closed) && !item.closed { return false }`

Spitballing but this'll be so helpful and I think with the ability to filter open/closed it'll be even better given how annoying it is to scroll through 500 notifications even though half of them are closed 🙄 In hindsight not my best moment

weyert: Yeah, I will finish my code for this tomorrow 👍

weyert:
@rnystrom  @Sherlouk Here you go! What do you think of this approach?

rnystrom:
@weyert mind sorting out the profile file conflicts really quick?

weyert:
Merged and fixed the compiling issues :)

Classes/Utility/SearchQueryTokenizer.swift
```diff
@@ -0,0 +1,53 @@
...
+    case status = "is"
+    case user = "user"
+}
+
+typealias Token = (action: TokenAction, value: String)
+
+/// Tokenizes a search query and returns an instance of ParseResult which consists
+/// of all the found "search tokens" such as `is:closed` and separately returns
+/// the text of the search query which isn't part of any search token
+func tokenizeSearchQuery(_ query: String) -> SearchQueryParseResults {
```
Sherlouk: Genuinely looking forward to this functionality! Such a drag scrolling through tonnes of closed issues to get to old open ones! Almost feel we should mimmick GitHub and default search to `is:open` 🤔 

Great work overall!

weyert: Yes, once this commit is in I will make a change so that you can search the inbox in a style of `is:pr` / `is:issue`, `type:pr` and `type:issue`.

rnystrom: @Sherlouk ya, I like the idea of defaulting to open. Much easier. I think we'll want to extend this search bar eventually. Maybe use GH's `is:open` stuff?

Sherlouk: `is:open` is already supported by this PR! 😅 That was my whole feature request!

Classes/Repository/RepositoryIssuesViewController.swift
```diff
@@ -73,13 +75,32 @@ class RepositoryIssuesViewController: BaseListViewController<NSString>, BaseList
         }
     }
 
+    override func objects(for listAdapter: ListAdapter) -> [ListDiffable] {
+        let allObjects = super.objects(for: listAdapter)
```
rnystrom: Accckkkkkkk I really really don't like this. Makes me regret making the base VC 😭 

I feel like we should probably bake filtering into the base VC and create a "search filterable" protocol on our models. I'll refactor all of this.

rnystrom:
Just noting that I'm going to pull parts of this out for now. Love the idea of this, but being on mobile, I think actual filter controls will be a lot simpler to use rather than typing things like `is:user` etc on an iPhone kb. Awesome work here, but going a little beyond what we need up front IMO.

Classes/Repository/RepositoryIssuesViewController.swift
weyert: Hehe, this was the best thing I could come up while working on it :)

PR #510, opened 2017-10-09
[DO NOT MERGE] Only add a reaction if we haven't yet

### LINKED ISSUE #492: Tap reaction to +1 and the use menu for same emoji lets you react twice
API figures this out and negates the second, but our UI +1s twice which is wrong.

<details>
<pre>
Version 1.11.0 (1943)
Device: iPhone 7 Plus (iOS 11.0.1)
TestFlight: false
</pre>
</details>

BasThomas:
Fixes #492 

This now mimics the GitHub behavior: reacting twice removes the first, reacting once... adds it, obviously.

### DIFF
--- a/Classes/Issues/Comments/Reactions/IssueCommentReactionCell.swift
+++ b/Classes/Issues/Comments/Reactions/IssueCommentReactionCell.swift
@@ -127,27 +127,35 @@ UICollectionViewDelegateFlowLayout {
     }
 
     @objc private func onThumbsUp() {
-        delegate?.didAdd(cell: self, reaction: .thumbsUp)
+        addOrRemove(reaction: .thumbsUp)
     }
 
     @objc private func onThumbsDown() {
-        delegate?.didAdd(cell: self, reaction: .thumbsDown)
+        addOrRemove(reaction: .thumbsDown)
     }
 
     @objc private func onLaugh() {
-        delegate?.didAdd(cell: self, reaction: .laugh)
+        addOrRemove(reaction: .laugh)
     }
 
     @objc private func onHooray() {
-        delegate?.didAdd(cell: self, reaction: .hooray)
+        addOrRemove(reaction: .hooray)
     }
 
     @objc private func onConfused() {
-        delegate?.didAdd(cell: self, reaction: .confused)
+        addOrRemove(reaction: .confused)
     }
 
     @objc private func onHeart() {
-        delegate?.didAdd(cell: self, reaction: .heart)
+        addOrRemove(reaction: .heart)
+    }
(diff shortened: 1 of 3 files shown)


Classes/Issues/Comments/Reactions/IssueCommentReactionCell.swift
```diff
@@ -127,27 +127,35 @@ UICollectionViewDelegateFlowLayout {
...
-        delegate?.didAdd(cell: self, reaction: .confused)
+        addOrRemove(reaction: .confused)
     }
 
     @objc private func onHeart() {
-        delegate?.didAdd(cell: self, reaction: .heart)
+        addOrRemove(reaction: .heart)
+    }
+
+    func addOrRemove(reaction: ReactionContent) {
```
Sherlouk: also known as `toggle` 😅

BasThomas: What's this language you are speaking? 😁

Classes/Issues/Comments/Reactions/IssueCommentReactionCell.swift
```diff
@@ -127,27 +127,35 @@ UICollectionViewDelegateFlowLayout {
...
+        toggle(reaction: .confused)
     }
 
     @objc private func onHeart() {
-        delegate?.didAdd(cell: self, reaction: .heart)
+        toggle(reaction: .heart)
+    }
+
+    func toggle(reaction: ReactionContent) {
+        if let didReact = reactions.first(where: { $0.content == reaction })?.viewerDidReact, didReact == true {
```
rnystrom: I'm not crazy about this. We're putting biz logic into the view here.

Also, I think the menu should be an additive action only, and no-op if the reaction is already there (per #492).

I think just a `guard` checking that the reaction doesn't exist yet would be good here:

https://github.com/rnystrom/GitHawk/blob/master/Classes/Issues/Comments/IssueCommentSectionController.swift#L256

BasThomas: Agree on the biz logic. The no-op would be a deviation from how GitHub does it now, though. That ok?

rnystrom: Oooh didn't realize that's what GH does. Hmm. Maybe let's stick w/ how GH does it, just move the biz logic into the controller?

BasThomas:
This now totally no-ops, but wanted to add the addition of the delegate. Talked with Ryan and he'll take a look at this soon 👍

PR #516, opened 2017-10-09
Ensure the `Signout`-button has the destructive red text colour

### LINKED ISSUE #515: Sign Out button lacks destructive coloring
@rnystrom Super minor but storyboard says it should be red 🤔 Weird one

(Latest TF build)

weyert:
Fixes #515


weyert:
@rnystrom Probably, only then we need to go through all the code where `StyledTableCell` is used and ensure the colour properly set.

rnystrom:
Took care of this myself, all good to just clean up the base class

BasThomas:
Wasn’t it also centered? 

rnystrom:
A while ago ya, not anymore 

PR #519, opened 2017-10-10
Enable/Disable New issue submit button based on title text

### LINKED ISSUE #518: New Issue Submit button could initially be disabled
- This is an enhancement that would better match what Github currently does with issues
- Essentially, the buttons enabled state is tied to validity of the new issue (i.e. having a title). 
  - The Submit button starts off disabled until you've entered a non-empty title
  - Blank strings count as a non-empty string despite the API giving an error if you tap the Submit button.

### Github New Issue View

submit_1
submit_2

### GitHawk New Issue View

simulator screen shot - iphone 8 - 2017-10-09 at 17 32 06

simulator screen shot - iphone 8 - 2017-10-09 at 17 37 28

(Latest master build as of October 9th, 2017)

rizwankce:
Closes #518

### DIFF
--- a/Classes/New Issue/NewIssueTableViewController.swift
+++ b/Classes/New Issue/NewIssueTableViewController.swift
@@ -83,6 +83,10 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         return viewController
     }
     
+    deinit {
+        titleField.removeTarget(self, action: nil, for: .editingChanged)
+    }
+    
     override func viewDidLoad() {
         super.viewDidLoad()
 
--- a/Classes/New Issue/NewIssueTableViewController.swift
+++ b/Classes/New Issue/NewIssueTableViewController.swift
@@ -203,7 +203,7 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
     /// Called when editing changed on the title field, enable/disable submit button based on title text
     @objc
     func textFieldEditingChanged(_ textField: UITextField) {
-        navigationItem.rightBarButtonItem?.isEnabled = textField.hasText
+        navigationItem.rightBarButtonItem?.isEnabled = titleText != nil
     }
 
 }
--- a/Classes/New Issue/NewIssueTableViewController.swift
+++ b/Classes/New Issue/NewIssueTableViewController.swift
@@ -83,10 +83,6 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         return viewController
     }
     
-    deinit {
-        titleField.removeTarget(self, action: nil, for: .editingChanged)
-    }
-    
     override func viewDidLoad() {
         super.viewDidLoad()
 
@@ -103,11 +99,6 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         
         // Make the return button move on to description field
         titleField.delegate = self
-        titleField.addTarget(
-            self,
-            action: #selector(NewIssueTableViewController.textFieldEditingChanged),
-            for: .editingChanged
-        )
         
         // Setup markdown input view
         bodyField.githawkConfigure(inset: false)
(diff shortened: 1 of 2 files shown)


Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -99,6 +99,11 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         
         // Make the return button move on to description field
         titleField.delegate = self
+        titleField.addTarget(
+            self,
+            action: #selector(NewIssueTableViewController.textFieldEditingChanged),
```
Sherlouk: Are we not able to use the titleField delegate we already define for this functionality? Transversely, do we need to remove the target in deinit?

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -189,5 +195,11 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         bodyField.becomeFirstResponder()
         return false
     }
+    
+    /// Called when editing changed on the title field, enable/disable submit button based on title text
+    @objc
+    func textFieldEditingChanged(_ textField: UITextField) {
+        navigationItem.rightBarButtonItem?.isEnabled = textField.hasText
```
ajfigueroa: How does `textField.hasText` handle blank text?

Classes/New Issue/NewIssueTableViewController.swift
rizwankce: Checked text field delegate method `shouldChangeCharactersInRange` and thought it would need extra care when handling characters since the method has to return a `bool` for changing the text. 

Forgot about the remove target. Thx for pointing that out

Classes/New Issue/NewIssueTableViewController.swift
rizwankce: Intentionally used `hasText` to match with Github behaviour. It will not handle the blank text. As like you mentioned in #518 

>Blank strings count as a non-empty string despite the API giving an error if you tap the Submit button.

On Submit we will show an error as usual which checks for an empty string

Sherlouk: While I definitely agree with your point regarding GitHub behaviour, I feel like if we're adding this protection for no text we may as well add it for trimmed text as well - even if we already
 have the validation a little later 😅

ajfigueroa: ^ That would make for a better user experience.

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -189,5 +199,11 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         bodyField.becomeFirstResponder()
         return false
     }
+    
+    /// Called when editing changed on the title field, enable/disable submit button based on title text
+    @objc
+    func textFieldEditingChanged(_ textField: UITextField) {
```
ajfigueroa: Since this controller is currently in a Storyboard, would it be better to hook up the `.editingChanged` event as an `@IBAction`? It doesn't make a difference functionally but it is less code.

I see it [used](https://github.com/rnystrom/GitHawk/search?utf8=✓&q=ibaction&type=) in a few places in the codebase.

Classes/New Issue/NewIssueTableViewController.swift
rizwankce: Just checked, we already have a property `titleText` which checks for trimmed/empty check. So reused that instead of `hasText`.
Will this be fine? @Sherlouk

Classes/New Issue/NewIssueTableViewController.swift
rizwankce: @Sherlouk `IBAction` or target which one you prefer? here

Sherlouk: If we can get away with `IBAction` then sure, cleans it self up better 😃

Classes/New Issue/NewIssueTableViewController.swift
Sherlouk: Yup! I knew that would be helpful when I built it 😉

Classes/New Issue/NewIssueTableViewController.swift
rizwankce: Done! lot simpler now with `IBAction`

PR #526, opened 2017-10-11
Disable Apollo when triggered via BuddyBuild

Sherlouk:
This change checks for an env variable called "DISABLE_APOLLO", and if that is present then exits successfully out of any further changes.

We shouldn't be running Apollo on CI because A. If it does actually make changes, then they should be picked up by the PR not CI as they wouldn't get committed back in, and B. it doesn't work 😅 

Files changed is difficult to see, so:
```shell
if [[ ! -z "${DISABLE_APOLLO}" ]]; then
  echo "Stopping Task: Remove 'DISABLE_APOLLO' env variable to continue."
  exit
fi

... Continue with old stuff
```

### DIFF
--- a/Classes/Issues/AddCommentClient.swift
+++ b/Classes/Issues/AddCommentClient.swift
@@ -6,6 +6,8 @@
 //  Copyright © 2017 Ryan Nystrom. All rights reserved.
 //
 
+// BuddyBuild requires a source change to build, will remove
+
 import Foundation
 
 protocol AddCommentListener: class {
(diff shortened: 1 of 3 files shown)


Sherlouk:
Nailed it!

https://dashboard.buddybuild.com/apps/59d904ff8588d60001729c4e/build/59ddf14fc4613f000158ed87#details

Can change the scheme to be TestFlight now, and setup a schedule for auto-deployments to iTC 😄

Palleas:
Nice!

PR #528, opened 2017-10-11
Add ability to lock/unlock issues

### LINKED ISSUE #143: Lock/unlock issues and PRs

Sherlouk:
Closes #143


Classes/Issues/GithubClient+Issues.swift
```diff
@@ -183,4 +183,19 @@ extension GithubClient {
...
+        request(Request(
+            path: "repos/\(owner)/\(repo)/issues/\(number)/lock",
+            method: locked ? .put : .delete,
+            completion: { (response, _) in
+                // As per documentation this endpoint returns no content, so all we can validate is that
+                // the status code is "204 No Content".
+                if response.response?.statusCode == 204 {
+                    completion(.success(true))
+                } else {
+                    completion(.error(nil))
```
BasThomas: Will it actually never error?

Classes/Issues/IssuesViewController.swift
```diff
@@ -352,6 +363,42 @@ IssueCommentSectionControllerDelegate {
...
+        guard let currentStatus = current?.status else { return }
+        
+        let localModel = IssueStatusModel(
+            status: currentStatus.status,
+            pullRequest: currentStatus.pullRequest,
+            locked: locked
+        )
+        let localEvent = IssueStatusEventModel(
+            id: UUID().uuidString,
+            actor: client.sessionManager.focusedUserSession?.username ?? Strings.unknown,
```
BasThomas: Can omit the `Strings` here I think

Classes/Issues/IssuesViewController.swift
```diff
@@ -352,6 +363,42 @@ IssueCommentSectionControllerDelegate {
...
+            repo: model.repo,
+            number: model.number,
+            locked: locked
+        ) { [weak self] result in
+            switch result {
+            case .error:
+                self?.localStatusChange = nil
+                self?.feed.adapter.performUpdates(animated: true)
+                StatusBar.showGenericError()
+            default: break // dont need to handle success since updated optimistically
```
BasThomas: Any reason to use `default:` over `.success:` and break there?

Sherlouk: Carbon copy of the updateStatus method above it with a few minor tweaks for locking - Didn't rewrite what was already working

Classes/Issues/GithubClient+Issues.swift
Sherlouk: I'm not following? If it does error for what ever reason this will capture it and pass it through to the view controller which will then update the comment feed and display a message to the user

BasThomas: Ah, but the error param is always `nil` here, as you ignore/discard it [here](https://github.com/rnystrom/GitHawk/pull/528/files#diff-5d794bd4eac3501e9bf9ef3658a7c4f4R190)

Classes/Issues/IssuesViewController.swift
Sherlouk: Doesn't look like we can, the username is of type String but because the .unknown is namespaced into a Strings enum it can't be inferred

Classes/Issues/GithubClient+Issues.swift
Sherlouk: Actually the value being discarded there is GitHub paging info, this endpoint can't be paged. There is technically a way to get network errors out of the response as we do on many requests, but we never show a detailed error regardless so knowing what error is not as important as knowing there is _an_ error.

Classes/Issues/IssuesViewController.swift
rnystrom: Note to self, init this VC w/ the viewer or get name from GQL. Feels a little scary reaching into a session mgr for this and open/close status changes.

rnystrom: Also I should come up w/ a way to better streamline creating local changes like this.

Sherlouk: Agree a better way for optimistic changes is needed, but it's worth noting that the name used there is only used locally for a short period of time. Soon as you reload the view via network it will update to be the correct name (if there was any confusion). This is purely to mock what will happen, and doesn't change what's sent via the API!

Sherlouk:
@rnystrom Any changes needed on this one? Got some time tomorrow! 😄

Classes/Issues/IssuesViewController.swift
rnystrom: Ya I know, I've just seen dependency on session managers leaking all over a codebase over time, turns into spaghetti code.

rnystrom:
Nope I think we're good to go here!

PR #530, opened 2017-10-11
Add ability to delete comments

### LINKED ISSUE #144: Delete comments
Prompt with alert

Sherlouk:
Closes #144


Classes/Issues/Comments/IssueCommentSectionController.swift
```diff
@@ -75,6 +76,25 @@ AttributedStringViewIssueDelegate {
         return AlertAction(AlertActionBuilder { $0.rootViewController = weakSelf?.viewController })
             .share([url], activities: [TUSafariActivity()]) { $0.popoverPresentationController?.sourceView = sender }
     }
+    
+    func deleteAction() -> UIAlertAction? {
+        guard object?.viewerCanDelete == true, let number = object?.number else { return nil }
+        
+        return AlertAction.delete({ [weak self] _ in
```
BasThomas: Nit: Why the `({` instead of the ` {`? :)

Classes/Issues/Comments/IssueCommentSectionController.swift
```diff
@@ -75,6 +76,25 @@ AttributedStringViewIssueDelegate {
...
+            let title = NSLocalizedString("Are you sure?", comment: "")
+            let message = NSLocalizedString("Deleting this comment is irreversible, do you want to continue?", comment: "")
```
BasThomas: 🤔, do we maybe want to add a feature were you can revert this until you leave the VC?

BasThomas: Not sure if that many people would use it though.

Classes/Issues/Comments/IssueCommentSectionController.swift
```diff
@@ -75,6 +76,25 @@ AttributedStringViewIssueDelegate {
...
+            let alert = UIAlertController.configured(title: title, message: message, preferredStyle: .alert)
+            
+            alert.addActions([
+                AlertAction.cancel(),
+                AlertAction.yes({ [weak self] _ in
```
BasThomas: I think this should be a `delete` action, not `yes`.

Classes/Issues/GithubClient+Issues.swift
```diff
@@ -183,4 +184,16 @@ extension GithubClient {
...
     }
     
+    func deleteComment(owner: String, repo: String, commentID: Int, completion: @escaping (Result<Bool>) -> ()) {
+        request(Request(path: "repos/\(owner)/\(repo)/issues/comments/\(commentID)", method: .delete, completion: { (response, _) in
+            // As per documentation this endpoint returns no content, so all we can validate is that
+            // the status code is "204 No Content".
+            if response.response?.statusCode == 204 {
+                completion(.success(true))
+            } else {
+                completion(.error(nil))
```
BasThomas: I think we should propagate the error here.

Classes/Issues/IssuesViewController.swift
```diff
@@ -386,7 +389,14 @@ IssueCommentSectionControllerDelegate {
             objects.append(rootComment)
         }
         
-        objects += current.timelineViewModels
+        objects += current.timelineViewModels.filter({ (model) -> Bool in
```
BasThomas: Not sure if this matters, but I *think* `append` might be more performant here? Thinking about the concatenation of strings. But like I said, not sure. 😛

Classes/Issues/IssuesViewController.swift
```diff
@@ -504,5 +516,25 @@ IssueCommentSectionControllerDelegate {
...
+        // Actually delete the comment now
+        client.deleteComment(owner: model.owner, repo: model.repo, commentID: commentID) { [weak self] result in
+            switch result {
+            case .error:
+                if let index = self?.deletedComments.index(of: commentID) {
+                    self?.deletedComments.remove(at: index)
+                    self?.feed.adapter.performUpdates(animated: true)
+                }
+                
+                StatusBar.showGenericError()
```
BasThomas: Not for this PR, but I think we can do better regarding showing actual errors.

BasThomas:
Test

BasThomas:
Test

Classes/Utility/AlertAction.swift
```diff
@@ -106,6 +106,10 @@ struct AlertAction {
         return UIAlertAction(title: NSLocalizedString("Discard", comment: ""), style: .default, handler: handler)
     }
     
+    static func delete(_ handler: AlertActionBlock? = nil) -> UIAlertAction {
+        return UIAlertAction(title: NSLocalizedString("Delete", comment: ""), style: .default, handler: handler)
```
BasThomas: Any reason this is not `.destructive`?

BasThomas:
Test

Classes/Utility/AlertAction.swift
Sherlouk: Copy & Pasta, will change

Classes/Issues/IssuesViewController.swift
Sherlouk: Agreed

Classes/Issues/IssuesViewController.swift
Sherlouk: 🤷‍♂️  Didn't change it versus what was already there, maybe one to explore

Classes/Issues/Comments/IssueCommentSectionController.swift
Sherlouk: I did this initially then changed it based on how we handle other confirmation alerts, namely sign out of all accounts.

Happy to change

Classes/Issues/Comments/IssueCommentSectionController.swift
Sherlouk: To be quite honest, I've only ever deleted like three comments in the several years I've used GitHub... I feel this is a pretty rare case that someone would then regret it

Classes/Issues/Comments/IssueCommentSectionController.swift
Sherlouk: Copy and Pasta, will change

Sherlouk:
Ahh I see what you're saying, so the reason for that is when we create a new comment it's added to a different array (not the original view models array which I filter).

What I can do here is move the filter to right at the end before we return objects, that way every model is checked rather than just the original content models?

@BasThomas

Sherlouk:
@BasThomas Give those changes a shot, should be better 😄

Classes/Issues/IssuesViewController.swift
rnystrom: TBD w/ new toast

Classes/Issues/GithubClient+Issues.swift
```diff
@@ -183,4 +184,16 @@ extension GithubClient {
...
+        request(Request(path: "repos/\(owner)/\(repo)/issues/comments/\(commentID)", method: .delete, completion: { (response, _) in
+            // As per documentation this endpoint returns no content, so all we can validate is that
+            // the status code is "204 No Content".
+            if response.response?.statusCode == 204 {
+                completion(.success(true))
+            } else {
+                completion(.error(response.error))
+            }
+        }))
+    }
```
rnystrom: Another option for this would be to do all the deletion inside the section controller and then just return `0` for `numberOfItems`. That'd keep any of it from creeping back into the view controller, as well as make local vs remote comments exactly the same. Thoughts?

rnystrom:
This is so great, thanks for knocking this out!

Classes/Issues/GithubClient+Issues.swift
Sherlouk: Happy to make the change, but I do feel this will begin to add further complexity where it may not be super obvious where items are being filtered out/in. At the moment we have one place which defines all the models, and we expect something to show for them - this would then say "there should be a comment here" but then later on "but let's not show it".

Get what I mean?

Sherlouk: Still want this change made @rnystrom? If so will sort it out in morning 😄

rnystrom: Ya, let's move it to a 0 item. I'd rather keep the complexity local to the section controller instead of too many wires. Will be much more modular.

Sherlouk:
Oops I did it again, I played with your heart, got lost in the game.

Sherlouk:
@rnystrom All updated 😄

PR #532, opened 2017-10-11
Make impactful actions destructive

BasThomas:
This makes the actions colored red, to indicate a) they have consequences and b) to make them stand out more from other actions

### DIFF
--- a/Classes/Utility/AlertAction.swift
+++ b/Classes/Utility/AlertAction.swift
@@ -103,7 +103,7 @@ struct AlertAction {
     }
     
     static func discard(_ handler: AlertActionBlock? = nil) -> UIAlertAction {
-        return UIAlertAction(title: NSLocalizedString("Discard", comment: ""), style: .default, handler: handler)
+        return UIAlertAction(title: NSLocalizedString("Discard", comment: ""), style: .destructive, handler: handler)
     }
     
     static func toggleIssue(_ status: IssueStatus, handler: AlertActionBlock? = nil) -> UIAlertAction {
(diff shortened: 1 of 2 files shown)


Classes/Utility/AlertAction.swift
```diff
@@ -103,23 +103,23 @@ struct AlertAction {
...
+        return UIAlertAction(title: NSLocalizedString("Discard", comment: ""), style: .destructive, handler: handler)
     }
     
     static func toggleIssue(_ status: IssueStatus, handler: AlertActionBlock? = nil) -> UIAlertAction {
         let title = status == .open
             ? NSLocalizedString("Close", comment: "")
             : NSLocalizedString("Reopen", comment: "")
         
-        return UIAlertAction(title: title, style: .default, handler: handler)
+        return UIAlertAction(title: title, style: .destructive, handler: handler)
```
Sherlouk: Is reopening classified as a destructive measure? 🤔

BasThomas: Ha, was thinking about this one. I *think* it would be better? As in: "this action has consequences". Was thinking to do it only for close tho.

rnystrom: I get your point, but to me reopening isn't a destructive action. Red = "negative" actions. If anything this could be bold?

BasThomas: From the docs:

> Apply a style that indicates the action might change or delete data.

BasThomas: It's a tricky one - I lean towards destructive.

> As in: "this action has consequences"

rnystrom:
Alrighty, we can try it out

Issue #523, opened 2017-10-10
New iOS 11 nav bar titles?


rnystrom:
Would this looks good?

<details>
<summary>Bug Report Dump (Auto-generated)</summary>
<pre>
Version 1.11.0 (1997)
Device: iPhone 7 Plus (iOS 11.0.2)
TestFlight: false
</pre>
</details>

BasThomas:
Oh lol. https://github.com/rnystrom/GitHawk/pull/522#discussion_r143846185

Think it would look good, at least for the root vcs!

PR #533, opened 2017-10-11
[DO NOT MERGE] Add large titles

BasThomas:
As per #523.

Not sure if I like it that much. There's also a bug with this in the inbox screen atm, so please do not merge in this state.

As you can see, the title isn't actually large (until you scroll up) and there is some spacing between the nav bar and the segmented control. I think this was introduced somewhere when adding the tab bar?
simulator screen shot - iphone x - 2017-10-11 at 15 40 13

Doesn't feel like it adds too much here.
simulator screen shot - iphone x - 2017-10-11 at 15 40 15

I've left out a large title in the search tab, as there is no title there atm, just the search bar - which I think does enough on its own.

### DIFF
--- a/Classes/Notifications/NotificationsViewController.swift
+++ b/Classes/Notifications/NotificationsViewController.swift
@@ -56,6 +56,10 @@ TabNavRootViewControllerType {
         resetRightBarItem()
 
         navigationController?.tabBarItem.badgeColor = Styles.Colors.Red.medium.color
+        if #available(iOS 11.0, *) {
+            navigationController?.navigationBar.prefersLargeTitles = true
+            navigationItem.largeTitleDisplayMode = .automatic
+        }
     }
 
     override func viewWillAppear(_ animated: Bool) {
--- a/Classes/Settings/SettingsViewController.swift
+++ b/Classes/Settings/SettingsViewController.swift
@@ -49,6 +49,11 @@ NewIssueTableViewControllerDelegate {
             name: NSNotification.Name.UIApplicationDidBecomeActive,
             object: nil
         )
+
+        if #available(iOS 11.0, *) {
+            navigationController?.navigationBar.prefersLargeTitles = true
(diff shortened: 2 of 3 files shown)


BasThomas:
This is what Search looks like in the App Store app. If we do want to go ahead with this, I propose to also do something like this with the Search tab to keep the mean tabs consistent.

img_1616

Sherlouk:
If only we could do trending repositories 🤔 

I support "Recent Searches" would go there

BasThomas:
I'll name it just "Search" for now - we already have the "Recent Searches" once you actually have some.

BasThomas:
There's already a "Recent Searches" thing when you actually have some recent searches, so I propose:

- title "Search"
- search bar "Search GitHub"

rnystrom:
Hmph, what do you all think? Doesn't look worth it to me.

BasThomas:
If we think it looks better, we should go for it. It's an easy thing; I feel it being worth if we like it more than the current nav bars.

weyert:
A bit of waste of space, if you ask me! 😇

BasThomas:
FWIW, it seems like master has a large title on the Search tab already atm?

rnystrom:
Ya that was to fix the transition, but I don’t like the inconsistency here

BasThomas:
Feel like we should go for all big or all small?

rnystrom:
Ya, lets see if we can stick to small for now. Can fix the search VC or just revert it.

BasThomas:
Was just about to say that I defenitely DO like it on iPad tho!

rnystrom:
Trying this out now w/ inbox refactor

BasThomas:
Only seeing it on Inbox?

rnystrom:
Atm ya, will see how ppl (and me) like it

BasThomas:
Feel like we should go with it. These work amazingly well with searchbars as well imo, and now that we have quite a few of those...

PR #534, opened 2017-10-11
Notification mark read

### LINKED ISSUE #496: Mark read on open is triggered too soon
If the request fails for whatever reason it’s possible that it gets marked as read but the issue doesn’t load (just experienced this on the train with intermittent network)

<details>
<pre>
Version 1.11.0 (1942)
Device: iPhone 6s Plus (iOS 11.1)
TestFlight: true
</pre>
</details>

vadimue:
Closes #496

### DIFF
--- a/Classes/Notifications/NotificationSectionController.swift
+++ b/Classes/Notifications/NotificationSectionController.swift
@@ -45,11 +45,6 @@ SwipeCollectionViewCellDelegate {
             let cell = collectionContext?.cellForItem(at: index, sectionController: self) as? NotificationCell
             else { fatalError("Missing object, cell missing, or incorrect type") }
 
-        if NotificationClient.readOnOpen() {
-            cell.isRead = true
-            client.markNotificationRead(id: object.id, isOpen: true)
-        }
-
         switch object.identifier {
         case .hash(let hash):
             viewController?.presentCommit(owner: object.owner, repo: object.repo, hash: hash)
@@ -58,7 +53,8 @@ SwipeCollectionViewCellDelegate {
             let controller = IssuesViewController(
                 client: client.githubClient,
                 model: model,
-                scrollToBottom: true
+                scrollToBottom: true,
+                issueDownloaded: markRead
             )
             let navigation = UINavigationController(rootViewController: controller)
             viewController?.showDetailViewController(navigation, sender: nil)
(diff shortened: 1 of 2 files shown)


Classes/Notifications/NotificationSectionController.swift
```diff
@@ -45,20 +45,20 @@ SwipeCollectionViewCellDelegate {
...
+            }
             viewController?.presentCommit(owner: object.owner, repo: object.repo, hash: hash)
         case .number(let number):
             let model = IssueDetailsModel(owner: object.owner, repo: object.repo, number: number)
             let controller = IssuesViewController(
                 client: client.githubClient,
                 model: model,
-                scrollToBottom: true
+                scrollToBottom: true,
+                issueDownloaded: markRead
```
Sherlouk: `markRead` itself doesn't update `cell.isRead` and also doesn't check if `NotificationClient.readOnOpen()` is true (it's a user-configurable option).

An interim block is needed to check these values before forwarding the request 😄

vadimue: Yes, checking on `NotificationClient.readOnOpen()` I have missed. And I guessed that combination of  `client.markNotificationRead(id: object.id, isOpen: false)` and `context.reload(self)` shoud update `cell.isRead`. Also I haven not access to cell in the closure.

Sherlouk: You should be able to just do `self?.cell.isRead = true` as per the original block - Obviously ensuring you don't hold onto `self` as to hold a retain cycle! Not sure that's a safe assumption given the code was added before, regardless you can't view the cell anyway so it may not be important as when you navigate back it will likely update everything from cache anyway. - Needs testing though.

vadimue: @Sherlouk I believe this will be right 😀  Btw, I can view the updating cell in case of iPad split view

vadimue:
@rnystrom Yes, I agree with you about passing closure. I did not know about implemented one-to-many listener at the NotificationClient. I have implemented NotificationClientListener in NotificationSectionController. But how I can get the cell for update `isRead` property? I tried `let cell = cellForItem(at: 0) as? NotificationCell` and it return the wrong cell. Also, I add listener to NotificationClient for each `notificationModel`, so I need check `if id == object?.id` at the `willMarkRead`

Issue #520, opened 2017-10-10
UITextView for viewing code (changes, browsing files)


rnystrom:
Looking at Working Copy, this looks super clean and would make the implementation **so much** simpler.

![img_6364](https://user-images.githubusercontent.com/739696/31396086-6ed7769c-adb0-11e7-949b-e5f21c4f5721.PNG)

Sherlouk:
Linking to #59 as well because we'd obviously want syntax highlighting on this!

If we do go ahead and use Highlightr, we can take advantage of their real-time highlighting [here](https://github.com/raspu/Highlightr#codeattributedstring). So that you can edit code and have it highlight right there and then!

I do recall (#443) you mentioning wanting something similar with Markdown. Could maybe fork or make our own version as our own module (#512)

rnystrom:
Ya I think Hightlightr is the way to go (its also what WC uses according to their OSS attribution).

weyert:
Interesting. I guess a custom `UITextView` with a `UITextView` and a custom view for drawing the editor gutter?

weyert:
@rnystrom @Sherlouk I have been working on this, it's part of the pull request #537. See an example below.
![Image](https://i.imgur.com/kdY6y2w.png)

I only noticed that it won't draw background to the full line something to look at in the future. Let me know, what you think of it so far :)

Sherlouk:
It's looking amazing! can't wait!

I agree about the full line background, that'd be a huge thing!

Also it would be nice if we could make text-wrap optional? If it's all in a scrollView we should be able to do horizontal and vertical scrolling.

At line 9 in your example it displays part of the patch context  - That's not actually part of the file so shouldn't have it's own line number and needs to be formatted appropriately

A feature request, but fully understand if it's out of scope would be to highlight what's actually different (à la GitHub). So on line 32/33 the background behind attributeView/codeTextView is slightly darker.

Equivalently, as line 32 was replaced with line 33, the line number shouldn't increase

weyert:
Currently, I have added the gutter with line numbers to the `UITextView` which is just counting all lines which are given to the `attributedText`-field. I guess an even more customised `NSLayoutManager` which supports `diff patches`. If I understand text rendering `NSLayoutManager` is not `UITextView` thing as the default implementation is already used in `NSAttributedStringSizing`. 

The background color line issue might be fixable by somehow using paragraphs or maybe check if there is only one background color attribute for a line and then fill up the whole space. 

[...]

PR #537, opened 2017-10-12
WIP: Improved the patch file screen

weyert:
My first attempt to get color coding and diffing working together. Purposely attaching a screen where it's not working perfectly. Let me know, what you think of it. 

Screenshot:
Image


weyert:
@Sherlouk @BasThomas @rnystrom Played around a bit :)

Pods/Highlightr/Pod/Classes/HTMLUtils.swift
```diff
@@ -0,0 +1,294 @@
+
+import Foundation
+
+//Adapted from: https://gist.github.com/jrmgx/3f9f1d330b295cf6b1c6
+
+private let characterEntities : [String: Character] = [
```
Sherlouk: Does this duplicate anything from the HTMLString pod?

Classes/Views/DiffString.swift
```diff
@@ -7,6 +7,35 @@
...
 
 import UIKit
+import Highlightr
+
+func CreateColorCodedString(code: String, includeDiff: Bool=false) -> NSAttributedString {
+    guard let highlightr = Highlightr() else { return NSAttributedString(string: code) }
+
+    if includeDiff {
+        highlightr.setTheme(to: "github")
+        let colorCodedString = highlightr.highlight(code, as: "swift") ?? NSAttributedString(string: code)
```
Sherlouk: Don't think we can/should assume the language, could be anything! Objective-C 😷

Classes/Views/DiffString.swift
```diff
@@ -7,6 +7,35 @@
...
+            if line.hasPrefix("+") {
+                attributedText.addAttribute(.backgroundColor, value: Styles.Colors.Green.light.color as Any, range: nsrange)
+            } else if line.hasPrefix("-") {
+                attributedText.addAttribute(.backgroundColor, value: Styles.Colors.Red.light.color as Any, range: nsrange)
+            }
+        }
+
+        return attributedText
+    } else {
+        highlightr.setTheme(to: "github")
```
Sherlouk: Both parts of `if` statement set theme, should move this out top

Classes/Views/CodeTextView.swift
```diff
@@ -0,0 +1,62 @@
...
+//  CodeTextView.swift
+//  Freetime
+//
+//  Created by Weyert de Boer on 12/10/2017.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import UIKit
+
+final class CodeTextView: UIView, UITextViewDelegate {
```
Sherlouk: I'm liking where this is going, be great to see it finished!

Sherlouk:
In reference to the image above, is it possible to get green inbetween the gaps? Feels like I'm seeing `+` for every line but green only sometimes?

Out of scope for this card, but on tablet it would be nice to get side-by-side comparison à la GitHub

Pods/Highlightr/Pod/Classes/HTMLUtils.swift
weyert: I think it does but I haven't verified it.

Classes/Views/DiffString.swift
weyert: You can automatically detect the language but then it didn't looked that good. See:
Image

weyert:
@Sherlouk Yes, that's why purposely attached that image because this is the problem I am having. I am not sure why it makes the `+` on the first line green and why it sometimes goes wrong. I have to look into it. Currently the `Pod` doesn't support language + diffing together hence my current solution. This my first time I am messing around with `NSAttributedString`.

Classes/Views/CodeTextView.swift
weyert: Going to be my attempt to get line numbers. If you know a good approach don't hesitate to let me know!

weyert:
I am not sure why it not always find the appropriate + to mark them. Anyone here having more experience with `NSAttributedString`? My other approach would be to somehow customise the HTML parsing by ourselves or maybe somehow customise the Hightltr.js so that you can combine language + diff. Anyone got suggestions? @Sherlouk @rnystrom @BasThomas ?

Sherlouk:
https://github.com/raspu/Highlightr/blob/master/Pod/Assets/Highlighter/languages/diff.min.js

Highlightr does support a diff language?

weyert:
Yes, but don't you want to have color coding while displaying the diff?

Sherlouk:
image

Looking at the Highlightr demo page - The GitHub styling for diffing does have the color coding as we want:

https://highlightjs.org/static/demo/ (Select GitHub and scroll down to Diff)

weyert:
Yes, there is a `diff` language but you can't then color code the language that's why I wrote that specific code but some lines don't get picked up.

Sherlouk:
If you make sure colloborators can commit to your branch, I'll take a look tonight and see if I spot anything 😄

weyert:
@Sherlouk How do I do that?

Sherlouk:
@weyert All you need to do is tick the box on right 😄 

image

weyert:
Oh, that's checked 👍

Sherlouk:
Okay so the issue is this line:

```
guard let range = code.range(of: line) else { continue }
```

That's getting the range of the first instance of the provided line text. It's skipping lines like the one `//` under `// Freetime` because it's already handled a line that is exactly that. What we need to do is get a mutable version of the `code` variable. After each loop through the `lines` array, we need to remove `line` from the start of `code` ultimately getting us to the point where `code` is empty.

This will enable the `range(of: line)` to get the line we're actually looking at, rather than the first instance of it 😄

Make sense?

weyert:
Guess, it fails when the line is not unique in the `code` variable, correct? 🤓

Sherlouk:
Exactly

weyert:
Okay, I will try to resolve it later today. But don't hesitate to fix it if you want. I will work on it in a hour.

weyert:
Let's have a look!

weyert:
@Sherlouk Working: Image

Classes/Utility/NSMutableAttributedString+Replace.swift
```diff
@@ -0,0 +1,51 @@
...
+            if range.location != NSNotFound {
+                let lineRange = NSRange(location: range.location, length: searchLength)
+
+                switch mode {
+                case .first:
+                    self.addAttributes(attributes, range: lineRange)
+                    return
+                case .all:
+                    self.addAttributes(attributes, range: lineRange)
+                    break
```
BasThomas: You don't need to `break`; it's inferred.

weyert:
I have implemented a rough version of `CodeTextView` by adapting an existing Objective-C component. Currently, it looks like this In Playground: Image

I have committed the code, could someone try if it actually compiles in the project and works? I am experiencing  sudden issues with Xcode 9 (uncaught exception) and opening projects. Restarting computer doesn't work :( If someone could have a peek, that would be awesome!

weyert:
See the screenshot in #520. I think it's looking quite good already

weyert:
Better version of available via pull request #577

PR #540, opened 2017-10-13
Change all references of "Freetime" to "GitHawk"

joerocca:
Fixes issue in #501

This changes all non-breaking references of "Freetime" to "GitHawk".


joerocca:
Oops! Didn't realize this breaks NSCoding. Closing it.

PR #542, opened 2017-10-13
Use model.count instead of users.count which caps at 3

### LINKED ISSUE #541: Reaction detail label incorrect
Latest TF version

When 4+ people reacting it should say `name, name, name and x others`, but it seems to just stop after 3 here 🤔 

https://github.com/rnystrom/GitHawk/blob/master/Classes/Issues/Comments/Reactions/IssueCommentReactionCell.swift#L156-L176

img_0250

BasThomas:
Fixes #541


Classes/Issues/Comments/Reactions/IssueCommentReactionCell.swift
```diff
@@ -153,26 +153,22 @@ UICollectionViewDelegateFlowLayout {
...
         let detailText: String
-        switch users.count {
+        switch model.count {
         case 0:
             detailText = ""
         case 1:
             let format = NSLocalizedString("%@", comment: "")
-            detailText = String.localizedStringWithFormat(format, users[0])
-            break
+            detailText = .localizedStringWithFormat(format, users[0])
```
BasThomas: Why is this actually not `localizedString(format:)`. Just realizing this now, lol

Sherlouk: 🤷‍♂️

BasThomas:
I tested it (of course), let me add a pic

BasThomas:
From
screen shot 2017-10-13 at 14 31 48

To
screen shot 2017-10-13 at 14 33 00

Classes/Issues/Comments/Reactions/IssueCommentReactionCell.swift
```diff
@@ -153,26 +153,22 @@ UICollectionViewDelegateFlowLayout {
...
-            break
+            detailText = .localizedStringWithFormat(format, users[0], users[1])
         case 3:
             let format = NSLocalizedString("%@, %@ and %@", comment: "")
-            detailText = String.localizedStringWithFormat(format, users[0], users[1], users[2])
-            break
+            detailText = .localizedStringWithFormat(format, users[0], users[1], users[2])
         default:
             let difference = model.count - users.count
             let format = NSLocalizedString("%@, %@, %@ and %d other(s)", comment: "")
```
BasThomas: I just realized this `(s)` automatically works?!?!?!?!

Sherlouk:
Is it worth creating a quick unit test around this? Check that detailText is correct after 0-5 users?

rnystrom:
+1 @Sherlouk maybe we can just pull this logic into a standalone function and test it? Will help prevent regressions again (my b)

rnystrom:
@BasThomas beautiful!!

rnystrom:
Already hit a crash 😝 

If we're using `model.count` and then accessing indices on `users`, if anything is out of sync (our fault or GitHub) it will blow up. That's why I made this change originally. Fixing.

weyert:
You can also use `array.indices.contains(3432)` in Swift to check if the requested index exists in the array

PR #546, opened 2017-10-13
Add support to show static toasts to specific viewcontroller

### LINKED ISSUE #543: Report a bug doesn’t error when there is no internet connection
It will disable “Submit”, but then not handle any errors.

<details>
<pre>
Version 1.11.0 (1993)
Device: iPad Pro (iOS 11.0.3)
TestFlight: true
</pre>
</details>

BasThomas:
Fixes #543 

Bonus track: also doesn't disable the send button after a failure when it shouldn't


rnystrom:
Aha! Thanks for this!

PR #547, opened 2017-10-13
Add support for GIFs in comments

### LINKED ISSUE #75: Gif support

Sherlouk:
Closes #75 

@BasThomas 😉

This allows GIFs to be seen inline within comments, as well as when you tap on them to show them in NYT also - Haven't tested inside a README but they're treated same as issue comments for most part so should work.


Classes/Issues/Comments/Images/IssueCommentImageCell.swift
```diff
@@ -27,7 +27,7 @@ UIGestureRecognizerDelegate {
     weak var delegate: IssueCommentImageCellDelegate? = nil
     weak var heightDelegate: IssueCommentImageHeightCellDelegate? = nil
 
-    let imageView = UIImageView()
+    let imageView = FLAnimatedImageView()
```
BasThomas: Aha!

Sherlouk: That wasn't actually the necessary fix for what you was getting. The issue is the internals of SDWebImage - The getData function you were using was just getting the image data rather than the entire GIF data as SD discard it if it's not being used.

There's a different function you can use instead of loadImage to get the rawData, which we could've used to do GIFs when you tap on them.

Or in this situation, just store the data in an image view and get a double bonus!

BasThomas: Yes, true - but this was also part of the solution I didn't think of. :)

PR #555, opened 2017-10-13
Use actual filename for the file browsing screen

weyert:
Added an extension to the `File` extension so you can get the `potentialLanguage` (based on extension) and the actual name of the file expressed in the class

Updated the `IssuePatchContentViewController.swift` so that it only shoes the file name and not the full path


Classes/Models/File+Filename.swift
```diff
@@ -0,0 +1,27 @@
...
+        switch fileExtension {
+        case "swift": return "swift"
+        default:
+            return nil
+        }
+    }
+
+    /// Returns the actual file name
+    public var actualFileName: String {
+        return  self.filename.components(separatedBy: "/").last ?? ""
```
Sherlouk: Can we remove the double space, and if it fails it should default to just returning `fileName`

Classes/Models/File+Filename.swift
```diff
@@ -0,0 +1,27 @@
...
+
+import Foundation
+
+extension File {
+
+    /// Returns the potential programming language of the file, if unsure it will return `nil`
+    var potentialLanguage: String? {
+        guard let fileExtension = self.filename.lowercased().components(separatedBy: ".").last else { return nil }
+        switch fileExtension {
+        case "swift": return "swift"
```
BasThomas: Probably want to extract a list of known languages to an enum + separate file? And maybe we can already find a list like that somewhere?

Then we can `return Language(rawValue: fileExtension)` (or something similar, eg. when there are multiple extensions for one language) instead of this weird `switch`. :D

Classes/Models/File+Filename.swift
```diff
@@ -0,0 +1,27 @@
...
+//  Created by Weyert de Boer on 13/10/2017.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+extension File {
+
+    /// Returns the potential programming language of the file, if unsure it will return `nil`
+    var potentialLanguage: String? {
```
Sherlouk: Not sure this is necessary, at least yet? Also it shouldn't **only** support Swift, even if we do keep it

Classes/Models/File+Filename.swift
weyert: I have tidied up the code and removed the `potentialLanguage`. I was planning to use that for the 'you know what' PR as Highlight.js is not correctly automatically detected the Swift language 😅

rnystrom: Leaving this reply as a test for PR review threads, demonstrating how broken they are 😞

PR #559, opened 2017-10-14
Make lock action destructive, Append issue type

### LINKED ISSUE #143: Lock/unlock issues and PRs

### LINKED ISSUE #557: Append “Issue” or “Pull Request” to Close action
Help reinforce what action is being taken

<details>
<pre>
Version 1.11.0 (2013)
TestFlight: false
</pre>
</details>

Sherlouk:
Closes #143
Closes #557

Got to keep PR count >= 10 😉

### DIFF
--- a/Classes/Issues/IssuesViewController.swift
+++ b/Classes/Issues/IssuesViewController.swift
@@ -234,7 +234,7 @@ IssueCommentSectionControllerDelegate {
             status != .merged
             else { return nil }
         
-        return AlertAction.toggleIssue(status) { [weak self] _ in
+        return AlertAction.toggleIssue(status, issue: current?.status.pullRequest != true) { [weak self] _ in
             self?.setStatus(close: status == .open)
         }
     }
@@ -244,7 +244,7 @@ IssueCommentSectionControllerDelegate {
             return nil
         }
         
-        return AlertAction.toggleLocked(locked) { [weak self] _ in
+        return AlertAction.toggleLocked(locked, issue: current?.status.pullRequest != true) { [weak self] _ in
             self?.setLocked(!locked)
         }
     }
--- a/Classes/Issues/IssuesViewController.swift
+++ b/Classes/Issues/IssuesViewController.swift
@@ -234,7 +234,7 @@ IssueCommentSectionControllerDelegate {
             status != .merged
             else { return nil }
         
-        return AlertAction.toggleIssue(status, issue: current?.status.pullRequest != true) { [weak self] _ in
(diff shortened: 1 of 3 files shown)


Classes/Issues/IssuesViewController.swift
```diff
@@ -234,7 +234,7 @@ IssueCommentSectionControllerDelegate {
             status != .merged
             else { return nil }
         
-        return AlertAction.toggleIssue(status) { [weak self] _ in
+        return AlertAction.toggleIssue(status, issue: current?.status.pullRequest != true) { [weak self] _ in
```
BasThomas: Wait, do we not have access to the `NotificationType` from here?

Classes/Utility/AlertAction.swift
```diff
@@ -106,20 +106,24 @@ struct AlertAction {
         return UIAlertAction(title: NSLocalizedString("Discard", comment: ""), style: .destructive, handler: handler)
     }
     
-    static func toggleIssue(_ status: IssueStatus, handler: AlertActionBlock? = nil) -> UIAlertAction {
+    static func toggleIssue(_ status: IssueStatus, issue: Bool, handler: AlertActionBlock? = nil) -> UIAlertAction {
+        let issueType = issue ? "Issue" : "Pull Request"
```
BasThomas: Probably safe to wrap these in `NSLocalizedString`?

Sherlouk: No need as they're used in a NSLocalizedString directly beneath, if these were localized it would cause havoc down the line because you'd have

"Issue" = "**eisiúint**"
"Open **eisiúint**" = "Ceist oscailte" 

Which just over complicates things IMO

Classes/Issues/IssuesViewController.swift
Sherlouk: Pretty sure that's only available from the Notification level, not passed through once you go into an issue. This could be simplified slightly, in hindsight, by just doing `current?.pullRequest` but yh

Classes/Utility/AlertAction.swift
BasThomas: Good point 👍

BasThomas:
Was thinking about this - what about adding the Issue Type as a title / message to the AlertController instead of possibly repeating it in the actions themselves?

Sherlouk:
Wouldn’t be against it! Card it up?

rnystrom:
Ya! I love that idea

PR #569, opened 2017-10-15
Add done return key type that dismisses keyboard for notifications/issues/pr search

joerocca:
Gives the return key a purpose and also feels more natural(at least for me) than scrolling to dismiss.


BasThomas:
Why not `.search`?

joerocca:
I didn’t use .search because search happens in real time. I figured .done makes more sense because you are done searching when you press it.

BasThomas:
Ref #596

joerocca:
Done!

PR #577, opened 2017-10-16
Added color coding and gutter to the code related views, partly fixes #520

### LINKED ISSUE #520: UITextView for viewing code (changes, browsing files)

weyert:
This pull request will add support for auto detected language color coding and to display a gutter and line numbers. I have updated the patch and the code blow browser screen. Please, see how it looks like it in the screenshots.

**Patch Screen**
Image

**Code Browse Screen**
Image


Classes/Views/Styles.swift
```diff
@@ -73,8 +73,8 @@ enum Styles {
         static let button = UIFont.systemFont(ofSize: Styles.Sizes.Text.button)
         static let headline = UIFont.boldSystemFont(ofSize: Styles.Sizes.Text.headline)
         static let smallTitle = UIFont.boldSystemFont(ofSize: Styles.Sizes.Text.smallTitle)
-        static let code = UIFont(name: "Courier", size: Styles.Sizes.Text.body)!
-        static let secondaryCode = UIFont(name: "Courier", size: Styles.Sizes.Text.secondary)!
+        static let code = UIFont(name: "Menlo", size: Styles.Sizes.Text.body)!
```
Sherlouk: Any particular reason for the change of font? Also is Menlo monospaced?

Podfile.lock
```diff
@@ -86,6 +95,6 @@ SPEC CHECKSUMS:
   Tabman: 627b2bddc6c054bcd86c60919000c30c4adf2ec4
   TUSafariActivity: afc55a00965377939107ce4fdc7f951f62454546
 
-PODFILE CHECKSUM: ace2f34a7d73e2c1c9cdfc7471f86b2916e4cc49
+PODFILE CHECKSUM: c1b736f34980e21e7389fb17ca6896cf06d018ed
 
-COCOAPODS: 1.2.1
+COCOAPODS: 1.3.1
```
Sherlouk: Not part of this PR, but consider using NPM/bundler to ensure everyone is using same version of CP to prevent issues

Classes/Systems/NSAttributedStringSizing.swift
```diff
@@ -51,19 +50,24 @@ final class NSAttributedStringSizing: NSObject, ListDiffable {
...
+        showLineNumbers: Bool = false,
+        gutterDelegate: GutterLayoutManagerDelegate? = nil
         ) {
         self.attributedText = attributedText
         self.inset = inset
         self.screenScale = screenScale
         self.backgroundColor = backgroundColor
 
-        textContainer = NSTextContainer()
-        textContainer.exclusionPaths = exclusionPaths
```
Sherlouk: Was exclusionPaths ever used?

Classes/Repository/RepositoryCodeBlobViewController.swift
```diff
@@ -100,11 +117,21 @@ final class RepositoryCodeBlobViewController: UIViewController {
...
+        textView.attributedText = CreateColorCodedString(code: text, includeDiff: false, lineHandler: nil)
         let max = CGFloat.greatestFiniteMagnitude
         let size = textView.sizeThatFits(CGSize(width: max, height: max))
         textView.frame = CGRect(origin: .zero, size: size)
         scrollView.contentSize = size
     }
 
+    // MARK: GutterLayoutManagerDelegate implementation
+
+    func getMaximumGutterWidth(gutterLayoutManager: GutterLayoutManager) -> CGFloat {
```
Sherlouk: This looks identical to the implementation in GutterLayoutManager (ext. GutterLayoutManagerDelegate)

Classes/Issues/Files/IssuePatchContentViewController.swift
```diff
@@ -22,14 +111,79 @@ final class IssuePatchContentViewController: UIViewController {
...
+                    guard let lineInfo = chunkGutterRenderer.getLineDetails(for: index + 1) as LineMetadata? else { return .unknown }
+
+                    switch lineInfo.2 {
+                    case .addedLine: return .addedLine
+                    case .deletedLine: return .deletedLine
+                    default: return .unknown
+                    }
+                }
+
+                let text = NSAttributedStringSizing(
```
Sherlouk: A lot of these are default values, unnecessary to have them here

Classes/Views/GutterLayoutManager.swift
```diff
@@ -0,0 +1,256 @@
...
+//
+//  Created by Weyert de Boer on 15/10/2017.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+protocol GutterLayoutManagerDelegate {
+
+    func getLineForLine(gutterLayoutManager: GutterLayoutManager, lineNumber: Int, for line: String) -> String
```
Sherlouk: This name is a bit confusing 😅 Possible to revise?

Classes/Views/GutterLayoutManager.swift
```diff
@@ -0,0 +1,256 @@
...
+ GutterLayoutManager allows the drawing of gutter with line number on the left side of
+ a text view. A `GutterLayoutManagerDelegate` is availabla which is being used to query the
+ actual line number to be rendered on the screen. This useful when the to be rendered line number
+ differs from the internal line number in the text.
+
+ In the code base this is used to render the correct line numbers for a diff
+ **/
+class GutterLayoutManager: NSLayoutManager {
+
+    struct Constants {
```
Sherlouk: Should these be moved to the Constants file?

Classes/Views/GutterLayoutManager.swift
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    public var showLineNumbers: Bool = true {
+        didSet {
+            guard let textStorage = self.textStorage else { return }
+            self.invalidateDisplay(forCharacterRange: NSMakeRange(0, textStorage.length))
+        }
+    }
+
+    public var gutterTextColor = UIColor.fromHex("#A4A4A4") {
```
Sherlouk: Default value should probably be in the Styles -> Colors enum

Classes/Views/GutterLayoutManager.swift
```diff
@@ -0,0 +1,256 @@
...
+        didSet {
+            guard let textStorage = self.textStorage else { return }
+            self.invalidateDisplay(forCharacterRange: NSMakeRange(0, textStorage.length))
+        }
+    }
+
+    public var gutterFont = UIFont.systemFont(ofSize: 11.0) {
```
Sherlouk: Should probably be in Styles -> Fonts

Classes/Views/GutterLayoutManager.swift
```diff
@@ -0,0 +1,256 @@
...
+        didSet {
+            guard let textStorage = self.textStorage else { return }
+            self.invalidateDisplay(forCharacterRange: NSMakeRange(0, textStorage.length))
+        }
+    }
+    public var gutterDelegate: GutterLayoutManagerDelegate?
```
Sherlouk: If we can, maybe make the protocol require a class (`: class`) and make this weak to prevent against any retain cycles?

Classes/Views/Styles.swift
BasThomas: `Menlo` is monospaced, yep! Used by Apple prior to `SF Mono` in Xcode I think.

BasThomas: Which reminds me... why, in that case, would we go for `Menlo` over `SF Mono`?

weyert: I just thought it looked nice then Courier :) Shall I change it to SF Mono?

Classes/Issues/Files/IssuePatchContentViewController.swift
weyert: Yes, I didn't want to change the order of the existing parameters :)

Podfile.lock
weyert: Ooh, whoops. I will see if I can change back to `1.2.1`

Sherlouk: Don't worry about it, I'm pretty sure Ryan is on 1.3.1 also (It's me at fault!)

weyert:
Anyone had a chance to get it all compiled and had a look? I would love to know if I am heading the right direction and invest more time in improving the `GutterLayoutManager` @Sherlouk ?

Sherlouk:
Lot of changes to go through and haven't got tonnes of time just yet I'll try and have a proper look ASAP

Just compiled it though and I'm noticing that the longest line in every file is wrapped onto two lines. I'm going to hazard a guess and say the width of the gutter is not being added to the width of the scrollView? This'll need fixing

It'd also be nice if the gutter had a slight background to it, so you can visually see the difference vs just straight code - maybe take the colors of both the background and text from GitHub's web interface

Obviously we've already spoken about making full lines have the background, not just the text - think this would be a huge improvement! Would also be nice if the background extended over the line numbers as well.

Also as per the GitHub web interface, the line numbers (in patches) don't show for the add/delete lines on either the left/right respectively. This prevents having the same number on one side for large blocks of text, making it easier to read. I'd love this!

I like the method of chunking with the gap, I think with the aforementioned color change it would make it a bit more clear though as it's easy to think it's just a line with whitespace unless you look at the gutter!

I get we're using Highlightr for the main styling, so not sure how much control we have over it - but if we could tone the font size down a tad I think it would work better!

Classes/Utility/Parsers/DiffParser.swift
```diff
@@ -0,0 +1,223 @@
...
+//
+//  Created by Weyert de Boer on 14/10/2017.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+extension String {
+
+    func components(from matches: [NSTextCheckingResult]) -> [[String]] {
+        let line = self
+        return matches.map { result in
```
Sherlouk: I'm sorry I genuinely can't read the code in the function, Could we break it up a little to make a bit more sense please?

Classes/Issues/Files/IssuePatchContentViewController.swift
Sherlouk: We don't need to change the order of parameters, just remove the ones which are not the default implementation in this call to the init. Swift allows you to skip parameters where they have default values.

Classes/Views/DiffString.swift
```diff
@@ -7,6 +7,49 @@
 //
 
 import UIKit
+import Highlightr
+
+func CreateColorCodedString(code: String, includeDiff: Bool=false, lineHandler: ((String, Int) -> LineChangeType)? = nil) -> NSAttributedString {
+    guard let highlightr = Highlightr() else { return NSAttributedString(string: code) }
+
+    highlightr.setTheme(to: "github")
```
Sherlouk: Note for the future, not part of this card: Would be nice to have the option to switch this theme -- I personally find it easier to read code on a dark background. 🤷‍♂️

Classes/Views/GutterLayoutManager.swift
```diff
@@ -0,0 +1,257 @@
...
+//
+//  Created by Weyert de Boer on 15/10/2017.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+protocol GutterLayoutManagerDelegate: class {
+
+    func getLineForLine(gutterLayoutManager: GutterLayoutManager, lineNumber: Int, for line: String) -> String
```
Sherlouk: Possible to rename this delegate method to something that reads a bit better and makes more sense immediately?

Classes/Views/GutterLayoutManager.swift
```diff
@@ -0,0 +1,257 @@
...
+
+    func getLineForLine(gutterLayoutManager: GutterLayoutManager, lineNumber: Int, for line: String) -> String {
+        return "\(lineNumber)"
+    }
+
+    func getMaximumGutterWidth(gutterLayoutManager: GutterLayoutManager) -> CGFloat {
+        guard let textStorage = gutterLayoutManager.textStorage else { return 0 }
+        let numberOfLines = textStorage.string.components(separatedBy: .newlines).count
+        let maxLength = "\(numberOfLines)".count
+        let fillerString = String(repeating: "9", count: maxLength)
```
Sherlouk: If the font is monospace, is there a better/nicer way of calculating this?

weyert:
> Just compiled it though and I'm noticing that the longest line in every file is wrapped onto two lines. I'm going to hazard a guess and say the width of the gutter is not being added to the width of the scrollView? This'll need fixing

Hah, that's a good catch you might be correct!! The idea I had was to indent the renderable area for a line by adding the space of the gutter to it because `exclusionPaths` was failing for the `NSAttributedStringSizing`-class (using them caused a blank image). I will do some experimenting.

> It'd also be nice if the gutter had a slight background to it, so you can visually see the difference vs just straight code - maybe take the colors of both the background and text from GitHub's web interface

In the previous pull request I had a background colour for the gutter. This time I tried to match the look of Working Copy. @rnystrom, what kind of look, would you prefer?

> Obviously we've already spoken about making full lines have the background, not just the text - think this would be a huge improvement! Would also be nice if the background extended over the line numbers as well.

Yes, I wanted to wait until I got some feedback from this PR so I know whether this is heading to the right direction and then try to solve this problem. Currently, my idea is to get the line fragment rect for the line and then adapt it so the rectangle with is the full with of the text view and then draw it on the screen. I have been wondering if I could a new `CALayer` for this and draw it behind the text but I have to do some experimenting with this. It's the final goal :) 

> Also as per the GitHub web interface, the line numbers (in patches) don't show for the add/delete lines on either the left/right respectively. This prevents having the same number on one side for large blocks of text, making it easier to read. I'd love this!

Yes, I have been slacking off in this regard. I had this working in the playground but Xcode crashed before I could save it and lost it 🙄 It's the original reason why I have the `lineNumberCache` going for each duplicate line number I want to return nil so it renders nothing.

> I like the method of chunking with the gap, I think with the aforementioned color change it would make it a bit more clear though as it's easy to think it's just a line with whitespace unless you look at the gutter!

Definitely that would look nice, I have been thinking of creating a `ChunkElementView` which contains the two views, a header view and an `AttributedStringSize` view rendering the actual code.

> I get we're using Highlightr for the main styling, so not sure how much control we have over it - but if we could tone the font size down a tad I think it would work better!

This would need a change to `Styles.Fonts.code` to the size and/or font itself

Classes/Views/GutterLayoutManager.swift
weyert: This is the best approach I could come up. If anyone has a better idea, please don't hesitate to share!

Classes/Views/GutterLayoutManager.swift
weyert: If you have any suggestions 👼 What about `getVisualLineNumberForGutter(gutterLayoutManager: GutterLayoutManager, lineIndex: Int, for line: String)`?

Classes/Views/DiffString.swift
weyert: Adding a new parameter should this trick, right 💃

Classes/Utility/Parsers/DiffParser.swift
weyert: Gets the strings from all the groups of a regular expression match :)

weyert:
> We don't need to change the order of parameters, just remove the ones which are not the default implementation in this call to the init. Swift allows you to skip parameters where they have default values.

Wow, I really didn't know this was possible! I will give it a shot tomorrow 🆒

weyert:
Thanks @Sherlouk for your feedback! @BasThomas  and @rnystrom, I would love to hear from you too, whether it's heading the right direction and I should continue working on it.

weyert:
Did anyone else had a chance to have a look at it?

rnystrom:
Gonna revisit this later, things have diverged a lot and I just added Highlighr to the code view. Would love to get line numbers eventually, but not super high pri yet.

weyert:
To add the gutter we only need to start using the Gutter* related classes in the code view

PR #600, opened 2017-10-17
🐛 Disable selection in IssueCommentHtmlCell

### LINKED ISSUE #608: Make HTML template a Swift 4 String
We can do multiline strings like python now, right? [This is gross](https://github.com/rnystrom/GitHawk/blob/master/Classes/Issues/Comments/Html/IssueCommentHtmlCell.swift#L25-L53).

viktorgardart:
Now you can select and scroll the "Sent with Githawk" footer text. 

I only tried this with the Signature and "Bug report dump" not sure where else we're using the `IssueCommentHtmlCell`. 

Also Closes #608

### DIFF
--- a/Classes/Issues/Comments/Html/IssueCommentHtmlCell.swift
+++ b/Classes/Issues/Comments/Html/IssueCommentHtmlCell.swift
@@ -21,37 +21,46 @@ final class IssueCommentHtmlCell: UICollectionViewCell, ListBindable, UIWebViewD
 
     private static let WebviewKeyPath = #keyPath(UIWebView.scrollView.contentSize)
 
-    private static let htmlHead = [
-        "<!DOCTYPE html><html><head><style>",
-        "body{",
-        // html whitelist: https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/sanitization_filter.rb#L45-L49
-        // lint compiled style with http://csslint.net/
-        "font-family: -apple-system; font-size: \(Styles.Sizes.Text.body)px;",
-        "color: #\(Styles.Colors.Gray.dark);",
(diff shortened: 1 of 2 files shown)


rnystrom:
Is this bad tho?

viktorgardart:
@rnystrom I thought about it before creating the PR. 
It sure isn't bad, but feels weird to only be able to select the HTML parts of a comment. I would say either disable it in the HTML or allow it on native text as well! 🙂

Classes/Issues/Comments/Html/IssueCommentHtmlCell.swift
```diff
@@ -123,6 +123,8 @@ final class IssueCommentHtmlCell: UICollectionViewCell, ListBindable, UIWebViewD
     }
 
     func webViewDidFinishLoad(_ webView: UIWebView) {
+        webView.stringByEvaluatingJavaScript(from: "document.documentElement.style.webkitUserSelect='none'")
+        webView.stringByEvaluatingJavaScript(from: "document.documentElement.style.webkitTouchCallout='none'")
```
rnystrom: Shouldn't this be added into the HTML template?

https://github.com/rnystrom/GitHawk/blob/master/Classes/Issues/Comments/Html/IssueCommentHtmlCell.swift#L25-L53

viktorgardart:
@rnystrom Made the template into a multiline string and moved the javascript to the bottom of the body 👌

rnystrom:
@viktorgardart heads up that this broke some styles. `<detail>`s look like they are Times New Roman or something now.

viktorgardart:
@rnystrom that’s odd, I might have time to look into it on the plane later today! Otherwise tomorrow! 

PR #601, opened 2017-10-18
Add Bookmarks tab 

rizwankce:
~~WORK IN PROGRESS - PLEASE DON'T MERGE~~

Sending this PR for early review and to get some suggestions.

- Adds bookmarks tab in between search and settings tab
- Adds alert action to bookmark from issues and repo view controller
- Adds bookmark view controller which displays bookmarked issues, pull request and repo
- iPad friendly

Open for suggestion and improvements.

- Is it okay to have a tab?
- Is the name "Bookmarks" or "Add To Bookmark" make sense?
- Archiving bookmarks to documents directory for now. Do we need to use Defaults/ Core Data/ Realm?
- Any improvements on Bookmarks cell? 

ToDo
====  

- [X] Search bookmark
- [X] Swipe to delete 
- [X] Clear All
- [X] Accessiblity titles

Screenshot 
-------------------
bookmarks


rnystrom:
Ooommmgggg I really want this lol

rizwankce:
I'm done with my checklist here. Anything to improve here? Maybe I have to create custom bookmark cell?

Sherlouk:
Some other ideas, maybe they can be carded up separately?:
- Swipe a notification to "Save for Later" which bookmarks it?
- Some sort of indication that a bookmarked item has been updated since you last looked
- Bookmark action should be hidden (or changed to "Remove Bookmark") if already bookmarked
- Issue/PR should have color representation for merged/open/closed (easy indication)

Classes/Bookmark/BookmarkModel.swift
```diff
@@ -0,0 +1,114 @@
...
+//  BookmarkModel.swift
+//  Freetime
+//
+//  Created by Rizwan on 18/10/17.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+final class BookmarkModel: NSObject, NSCoding, Filterable {
```
Sherlouk: Project is in Xcode 9 now, would we be able to use `Codable` here to be "more swifty"?

Classes/Bookmark/BookmarkModel.swift
```diff
@@ -0,0 +1,114 @@
...
+        aCoder.encode(number, forKey: Keys.number)
+        aCoder.encode(name, forKey: Keys.name)
+        aCoder.encode(type.rawValue, forKey: Keys.type)
+        aCoder.encode(title, forKey: Keys.title)
+        aCoder.encode(hasIssueEnabled, forKey: Keys.hasIssueEnabled)
+    }
+    
+    override var hashValue: Int {
+        get {
+            return (name + owner + "\(number)").hashValue
```
Sherlouk: Interpolation over concatenation applicable here? (Minor compile speed increase)

Classes/Bookmark/BookmarkModel.swift
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+    
+    override func isEqual(_ object: Any?) -> Bool {
+        let bookmark = (object as? BookmarkModel)
```
Sherlouk: As per IGListKit diffing, it's usually good to check if `object === self` first, then switch the casting to `BookmarkModel` into a guard (returning false if they don't match)

Fail fast!

Classes/Bookmark/BookmarkModel.swift
```diff
@@ -0,0 +1,114 @@
...
+        return hashValue == bookmark?.hashValue
+    }
+    
+    // MARK: Filterable
+    
+    func match(query: String) -> Bool {
```
Sherlouk: Love the flexibility here!

Classes/Bookmark/BookmarkModel.swift
```diff
@@ -0,0 +1,114 @@
...
+    // MARK: Helper
+    
+    func getBookmarkImageIcon() -> UIImage? {
+        var imageName = ""
+        switch type {
+            case .repo: imageName = "repo"
+            case .issue: imageName = "issue-opened"
+            case .pullRequest: imageName = "git-pull-request"
+        }
+        return UIImage.init(named: imageName)?.withRenderingMode(.alwaysTemplate)
```
Sherlouk: I'm finding this bit of logic is being used in quite a few places around the app now (I know I had to do it for Repo view) - I wonder if there's a nicer interface which enables reuse?

Classes/Bookmark/BookmarkStore.swift
```diff
@@ -0,0 +1,62 @@
...
+import Foundation
+
+final class BookmarksStore {
+    static let shared = BookmarksStore()
+    
+    private let archivePath: String = {
+        return FileManager.default
+            .urls(for: .documentDirectory, in: .userDomainMask)
+            // crash intentionally on startup if doesn't exist
+            .first!
```
Sherlouk: Does this intentional crash ever happen? Or is it just in an eventuality where the app fails for some reason?

Would be nice if we didn't have a known crash 😅

Classes/Bookmark/BookmarkStore.swift
```diff
@@ -0,0 +1,62 @@
...
+    }
+    
+    var bookmarks: [BookmarkModel] {
+        return Array(_bookmarks)
+    }
+
+    // MARK: Private API
+    
+    func archive() {
+        NSKeyedArchiver.archiveRootObject(_bookmarks, toFile: self.archivePath)
```
Sherlouk: Out of interest, what was the reason for choosing this approach of storing the information?

Been a lot of very interesting methods of storing complex objects using Codable and User Defaults. Just curious on your thought process!

Classes/Bookmark/BookmarksViewController.swift
```diff
@@ -0,0 +1,218 @@
...
+    
+    func updateRightBarItem() {
+        navigationItem.rightBarButtonItem?.isEnabled = bookmarkStore.bookmarks.count > 0
+    }
+    
+    @objc private func onClearAll(sender: UIBarButtonItem) {
+        let alert = UIAlertController.configured(
+            title: NSLocalizedString("Bookmarks", comment: ""),
+            message: NSLocalizedString("Clear all bookmarks?", comment: ""),
+            preferredStyle: .alert
```
Sherlouk: Wondering if we should try match pattern with other confirmations, e.g. sign out?

"Are you sure?"
"All of your bookmarks will be lost. Do you want to continue?"

Classes/View Controllers/RootViewControllers.swift
```diff
@@ -49,3 +49,13 @@ func newSearchRootViewController(client: GithubClient) -> UIViewController {
     nav.tabBarItem.selectedImage = UIImage(named: "tab-search-selected")
     return nav
 }
+
+func newBookmarksRootViewController(client: GithubClient) -> UIViewController {
+    let title = NSLocalizedString("Bookmarks", comment: "nil")
```
Sherlouk: Nil not necessary in comment here, just empty string!

Classes/View Controllers/RootViewControllers.swift
```diff
@@ -49,3 +49,13 @@ func newSearchRootViewController(client: GithubClient) -> UIViewController {
...
+    let controller = BookmarksViewController(client: client)
+    controller.title = title
+    let nav = UINavigationController(rootViewController: controller)
+    nav.tabBarItem = UITabBarItem.init(tabBarSystemItem: .bookmarks, tag: 3)
```
Sherlouk: Explicit init unnecessary?

FreetimeTests/BookmarkStoreTests.swift
```diff
@@ -0,0 +1,59 @@
...
+//  FreetimeTests
+//
+//  Created by Rizwan on 18/10/17.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import XCTest
+@testable import Freetime
+
+final class BookmarkStoreTests: XCTestCase {
```
Sherlouk: Woo for tests, thanks! 🙌

Classes/Bookmark/BookmarksViewController.swift
```diff
@@ -0,0 +1,218 @@
...
+        let titleLabel = "\(bookmark.owner)/\(bookmark.name)"
+        cell?.textLabel?.text = bookmark.type == .repo ? titleLabel : titleLabel + " #\(bookmark.number)"
+        cell?.detailTextLabel?.text = bookmark.title
+        cell?.textLabel?.numberOfLines = 0
+        cell?.detailTextLabel?.numberOfLines = 0
+        cell?.accessibilityTraits |= UIAccessibilityTraitButton
+        cell?.isAccessibilityElement = true
+        cell?.accessibilityLabel = cell?.contentView.subviews
+            .flatMap { $0.accessibilityLabel }
+            .reduce("", { $0 + ".\n" + $1 })
```
Sherlouk: Interpolation over concatenation (compile issue), but yay for accessibility!

Classes/Bookmark/BookmarkStore.swift
BasThomas: I think it *shouldn't* ever. Doing a `fatalError` with a message might be better here, too?

Classes/Bookmark/BookmarkStore.swift
```diff
@@ -0,0 +1,62 @@
...
+    
+    private let archivePath: String = {
+        return FileManager.default
+            .urls(for: .documentDirectory, in: .userDomainMask)
+            // crash intentionally on startup if doesn't exist
+            .first!
+            .appendingPathComponent("bookmarks")
+            .path
+    }()
+    private var _bookmarks: Set<BookmarkModel>
```
BasThomas: Why the `_`?

Classes/Bookmark/BookmarkStore.swift
```diff
@@ -0,0 +1,62 @@
...
+    
+    // MARK: Init
+    
+    init() {
+        if let bookmarks = NSKeyedUnarchiver.unarchiveObject(withFile: archivePath) as? Set<BookmarkModel> {
+            _bookmarks = bookmarks
+        }
+        else {
+            _bookmarks = Set<BookmarkModel>()
```
BasThomas: You can just initialize the empty Set with `[]`.

Classes/Bookmark/BookmarksViewController.swift
```diff
@@ -0,0 +1,218 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    // MARK: View Life Cycle
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        searchBar.delegate = self
+        searchBar.placeholder = NSLocalizedString("Search", comment: "")
```
BasThomas: Can use `Constants.Strings.search`.

Classes/Bookmark/BookmarksViewController.swift
BasThomas: Yep and yep! Thanks for taking care of accessibility!

Classes/Bookmark/BookmarksViewController.swift
```diff
@@ -0,0 +1,218 @@
...
+        
+        present(alert, animated: true)
+    }
+    
+    func filter(query: String?) {
+        if let query = query {
+            filterdBookmarks = filtered(array: bookmarkStore.bookmarks, query: query)
+        }
+        else {
+            filterdBookmarks = nil
```
BasThomas: Does it make sense for the filteredBookmarks to be nullable? If we have a `nil` query, shouldn't `filteredBookmarks` be an empty array? What would happen now if the array is in fact empty instead of `nil`?

Classes/Bookmark/BookmarksViewController.swift
```diff
@@ -0,0 +1,218 @@
...
+        cell?.detailTextLabel?.numberOfLines = 0
+        cell?.accessibilityTraits |= UIAccessibilityTraitButton
+        cell?.isAccessibilityElement = true
+        cell?.accessibilityLabel = cell?.contentView.subviews
+            .flatMap { $0.accessibilityLabel }
+            .reduce("", { $0 + ".\n" + $1 })
+
+        cell?.imageView?.image = bookmark.getBookmarkImageIcon()
+        cell?.imageView?.tintColor = Styles.Colors.Blue.medium.color
+        return cell!
```
BasThomas: This implementation feels weird to me. Especially keeping the cell optional for so long and unwrapping at the end. We should also never have to call https://github.com/rnystrom/GitHawk/pull/601/files/a6266a248f339c6820d04f39d1e474e91c09f04e#diff-1e305760accc5a494351ba0602fa475fR130 - it's handled by the system in the line above I think?

At least, I think we should fail fast if the cell is invalid, and afterwards we can then just use `cell` as a non-optional.

Classes/Bookmark/BookmarksViewController.swift
```diff
@@ -0,0 +1,218 @@
...
+    }
+    
+    override func tableView(_ tableView: UITableView, canEditRowAt indexPath: IndexPath) -> Bool {
+        return filterdBookmarks != nil ? false : true // avoid swipe to delete when search bar is active
+    }
+    
+    override func tableView(_ tableView: UITableView,
+                            commit editingStyle: UITableViewCellEditingStyle,
+                            forRowAt indexPath: IndexPath) {
+        if (editingStyle == UITableViewCellEditingStyle.delete) {
```
BasThomas: Can omit the `UITableViewCellEditingStyle ` part.

Classes/Bookmark/BookmarksViewController.swift
```diff
@@ -0,0 +1,218 @@
...
+
+        case .issue, .pullRequest:
+            let issueModel = IssueDetailsModel(
+                owner: bookmark.owner,
+                repo: bookmark.name,
+                number: bookmark.number
+            )
+            let controller = IssuesViewController(client: client, model: issueModel)
+            let navigation = UINavigationController(rootViewController: controller)
+            showDetailViewController(navigation, sender: nil)
```
BasThomas: You can create a `let viewController` and `let navigation` before the switch and the `showDetail` after to avoid some duplication.

Classes/Bookmark/BookmarkStore.swift
rizwankce: Actually, i got inspired from this [line 17](https://github.com/rnystrom/GitHawk/blob/c30156c5a62a9d3899dc898f20ccb3a778bab549/Classes/Notifications/NotificationsDataSource.swift#L17) from `NotificationsDataSource`.

But i agree with doing a `fatalError` with a message.

Classes/Bookmark/BookmarkStore.swift
rizwankce: I had the same thought of using User Defaults and did a google search for `User Defaults vs NSKeyedArchiver` and found out from Stack overflow [answer](https://stackoverflow.com/questions/27644769/where-to-store-local-variables-nskeyedarchiver-or-nsuserdefaults) the below.

>NSUserDefaults is best for a small amount of data. If data is small the better option is NSUserDefaults else use NSKeyedArchiver.
 
Since we don't have any relationship between models, I went with archiving option.

Sherlouk: For me I just feel the NSKeyedArchiver is very ObjC, example [StackOverflow answer](https://stackoverflow.com/a/24664331).

We're having to use NSCopying and NSObject to get this support where we could be using the Swift version (Codable) and using JSONDecoders straight into UserDefaults

Classes/Bookmark/BookmarkModel.swift
```diff
@@ -0,0 +1,114 @@
...
+        return hashValue == bookmark?.hashValue
+    }
+    
+    // MARK: Filterable
+    
+    func match(query: String) -> Bool {
+        let lowerQuery = query.lowercased()
+        
+        if type.rawValue.contains(lowerQuery) { return true }
+        if String(number).contains(lowerQuery) {return true }
```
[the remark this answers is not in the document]
rizwankce: Wow. Nice find +1

Classes/Bookmark/BookmarkStore.swift
rizwankce: Inspired from `SearchRecentStore`. Just to differentiate between on memory bookmarks.

rnystrom:
Haven't had time to review, just noting that I'm planning on including this in 1.12 🚀

rizwankce:
@rnystrom super excited with 1.12. I'm gonna fix/incorporate review comments as soon as possible.

Classes/Bookmark/BookmarkModel.swift
rnystrom: I'm not too opinionated here. Familiar w/ `NSCoding` and idk the ups or downs of `Codable`.

Classes/Bookmark/BookmarkModel.swift
```diff
@@ -0,0 +1,94 @@
...
+//
+//  Created by Rizwan on 18/10/17.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+final class BookmarkModel: NSObject, NSCoding, Filterable {
+    
+    struct Keys {
```
rnystrom: `private` also lets change it to an `enum` so it can't ever be initialized

Classes/Bookmark/BookmarkStore.swift
rnystrom: `NSKeyedArchiver` seems like the right call here b/c we have control over the path, when it gets purged, saved, etc. `NSUserDefaults` syncs when the OS decides (or have to call `.synchronize()`). Plus we have to support other accounts, so we'll have to namespace each setting.

Also all of UD is loaded into memory on start, so bigger == slower.

Spent a LOT of time in UD + friends in Instagram.

Classes/Bookmark/BookmarkStore.swift
```diff
@@ -0,0 +1,62 @@
...
+//  Freetime
+//
+//  Created by Rizwan on 18/10/17.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+final class BookmarksStore {
+    static let shared = BookmarksStore()
```
rnystrom: This will share bookmarks between accounts, wont it? Is that how we want it to work? Might be a little weird swapping between accounts and forgetting who you're commenting as b/c the content all looks the same.

I'd also prefer not using singletons as much as physically possible. Dependency injection ftw.

Classes/Bookmark/BookmarksViewController.swift
```diff
@@ -0,0 +1,220 @@
...
+//  Freetime
+//
+//  Created by Rizwan on 18/10/17.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import UIKit
+
+class BookmarksViewController:
+    UITableViewController,
```
rnystrom: Any reason we're not using IGListKit here like we are everywhere else?

edit: `UITableView` editing stuff looks pretty handy here. Probably ok to deviate.

Classes/Bookmark/BookmarksViewController.swift
```diff
@@ -0,0 +1,220 @@
...
+            style: .plain,
+            target: self,
+            action: #selector(BookmarksViewController.onClearAll(sender:))
+        )
+    }
+    
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        rz_smoothlyDeselectRows(tableView: tableView)
+        tableView.reloadData()
```
rnystrom: Wont this `reloadData` interupt the smooth deselect?

Classes/Bookmark/BookmarksViewController.swift
```diff
@@ -0,0 +1,220 @@
...
+        searchBar.delegate = self
+        searchBar.placeholder = NSLocalizedString(Constants.Strings.search, comment: "")
+        searchBar.tintColor = Styles.Colors.Blue.medium.color
+        searchBar.backgroundColor = .clear
+        searchBar.searchBarStyle = .minimal
+        searchBar.sizeToFit()
+        tableView.tableHeaderView = searchBar
+        
+        navigationItem.rightBarButtonItem = UIBarButtonItem(
+            title: NSLocalizedString("Clear All", comment: ""),
```
rnystrom: This string exists already, would be great to move them to `Strings`

Classes/Bookmark/BookmarksViewController.swift
```diff
@@ -0,0 +1,220 @@
...
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        searchBar.delegate = self
+        searchBar.placeholder = NSLocalizedString(Constants.Strings.search, comment: "")
+        searchBar.tintColor = Styles.Colors.Blue.medium.color
+        searchBar.backgroundColor = .clear
+        searchBar.searchBarStyle = .minimal
+        searchBar.sizeToFit()
+        tableView.tableHeaderView = searchBar
```
rnystrom: We'll probably move to iOS 11 nav bars, so might want to play w/ iOS 11 and `navigationItem.searchController`

Classes/Bookmark/BookmarksViewController.swift
```diff
@@ -0,0 +1,220 @@
...
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        rz_smoothlyDeselectRows(tableView: tableView)
+        tableView.reloadData()
+        updateRightBarItem()
+    }
+    
+    override func viewWillLayoutSubviews() {
+        super.viewWillLayoutSubviews()
+    }
```
rnystrom: Remove?

Classes/Bookmark/BookmarksViewController.swift
```diff
@@ -0,0 +1,220 @@
...
+        cell?.detailTextLabel?.numberOfLines = 0
+        cell?.accessibilityTraits |= UIAccessibilityTraitButton
+        cell?.isAccessibilityElement = true
+        cell?.accessibilityLabel = cell?.contentView.subviews
+            .flatMap { $0.accessibilityLabel }
+            .reduce("", { $0 + ".\n" + $1 })
+
+        cell?.imageView?.image = bookmark.type.icon?.withRenderingMode(.alwaysTemplate)
+        cell?.imageView?.tintColor = Styles.Colors.Blue.medium.color
+        return cell!
```
rnystrom: Should we just create a subclass cell to incapsulate all this setup? Not a big deal since its only used here, but just an idea

Classes/Bookmark/BookmarksViewController.swift
```diff
@@ -0,0 +1,220 @@
...
+    }
+    
+    // MARK: TabNavRootViewControllerType
+    
+    func didSingleTapTab() {
+        
+    }
+    
+    func didDoubleTapTab() {
+        // TODO: Scroll to top for table view
```
rnystrom: nit: this should be in single-tap

Classes/Bookmark/BookmarksViewController.swift
```diff
@@ -0,0 +1,220 @@
...
+        if (editingStyle == .delete) {
+            bookmarkStore.remove(bookmark: getBookmarks()[indexPath.row])
+            tableView.deleteRows(at: [indexPath], with: .automatic)
+        }
+    }
+    
+    // MARK: UITableViewDelegate
+    
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: true)
```
rnystrom: Shouldn't do this if you want the smooth deselect stuff

Classes/View Controllers/RootViewControllers.swift
```diff
@@ -49,3 +49,13 @@ func newSearchRootViewController(client: GithubClient) -> UIViewController {
     nav.tabBarItem.selectedImage = UIImage(named: "tab-search-selected")
     return nav
 }
+
+func newBookmarksRootViewController(client: GithubClient) -> UIViewController {
+    let title = NSLocalizedString("Bookmarks", comment: "")
+    let controller = BookmarksViewController(client: client)
+    controller.title = title
+    let nav = UINavigationController(rootViewController: controller)
+    nav.tabBarItem = UITabBarItem(tabBarSystemItem: .bookmarks, tag: 3)
```
rnystrom: Hmm, I'd love to use GitHub's official iconography. There's a `bookmark` image already. Let's use that! Then we don't have to init a new item either.

rnystrom:
Alright gonna merge and give it a whirl!

rizwankce:
Oops. I was away for the whole weekend so thought of doing some improvements on Monday. 

But it's fine now. I'm gonna add new issues for book marks for my remaining Todo's. 

PR #602, opened 2017-10-18
Adds sharing to code browsing

### LINKED ISSUE #589: Add option to share file from code browser
<details>
<pre>
Version 1.11.0 (2091)
TestFlight: false
</pre>
</details>

BasThomas:
If in the future we want to be able to share a URL to the code / open the code in Safari / etc., let's create separate tickets for that and add an actionsheet at that point. This also doesn't add support to share images and other files yet, but we can easily do so once we support showing those file types.

Fixes #589

### DIFF
--- a/Classes/Repository/RepositoryCodeBlobViewController.swift
+++ b/Classes/Repository/RepositoryCodeBlobViewController.swift
@@ -17,6 +17,16 @@ final class RepositoryCodeBlobViewController: UIViewController {
     private let textView = UITextView()
     private let feedRefresh = FeedRefresh()
     private let emptyView = EmptyView()
+    private var sharingPayload: Any?
+    private lazy var sharingButton: UIBarButtonItem = {
+        let barButtonItem = UIBarButtonItem(
+            barButtonSystemItem: .action,
+            target: self,
+            action: #selector(RepositoryCodeBlobViewController.onShare(sender:)))
+        barButtonItem.isEnabled = false
+        
+        return barButtonItem
+    }()
 
     init(client: GithubClient, repo: RepositoryDetails, path: String) {
         self.client = client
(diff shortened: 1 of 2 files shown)


Classes/Repository/RepositoryCodeBlobViewController.swift
```diff
@@ -69,11 +81,25 @@ final class RepositoryCodeBlobViewController: UIViewController {
...
 
     @objc
     func onRefresh() {
         fetch()
     }
+    
+    @objc
+    func onShare(sender: UIBarButtonItem) {
+        guard let payload = sharingPayload else { return }
+        let activityController = UIActivityViewController(activityItems: [payload], applicationActivities: nil)
```
rnystrom: Should we just be sharing `textView.text` so we aren't duplicating the "source of truth" here?

Classes/Repository/RepositoryCodeBlobViewController.swift
```diff
@@ -17,6 +17,16 @@ final class RepositoryCodeBlobViewController: UIViewController {
     private let textView = UITextView()
     private let feedRefresh = FeedRefresh()
     private let emptyView = EmptyView()
+    private var sharingPayload: Any?
```
rnystrom: Why `Any?`?

BasThomas: Because the payload (`activityItems`) for the `UIActivityViewController` accepts `Any`; in the future this may not be `String` but eg. `UIImage` too.

Classes/Repository/RepositoryCodeBlobViewController.swift
BasThomas: Check above. How would we deal with this if the “source” is an image?

BasThomas:
Do we have the resource url somewhere? I’d love to be able to open in Safari too!

rnystrom:
It doesn’t look like it, we don’t even have the url in the GQL model. We probably have to make another builder that takes the resource path + branch to create a URL.

PR #607, opened 2017-10-18
Cancelled networking when the search cancel button is pressed

### LINKED ISSUE #549: Cancel networking when cancelling search before the request completes
Steps to reproduce:

1. Search something
2. Tap cancel
3. The "Recent searches" are shown
4. The request finishes, and the view is replaced with the results

4 shouldn't happen when you cancelled already

heshamsalman:
Tested this through a throttled Charles Proxy to ensure network requests are indeed being cancelled. 

Closes #549

### DIFF
--- a/Classes/Search/GithubClient+Search.swift
+++ b/Classes/Search/GithubClient+Search.swift
@@ -7,6 +7,7 @@
 //
 
 import Foundation
+import Apollo
 
 extension GithubClient {
     
@@ -14,12 +15,18 @@ extension GithubClient {
         case error
         case success(String?, [SearchRepoResult])
     }
-    
-    func search(query: String, before: String? = nil, containerWidth: CGFloat, completion: @escaping (SearchResultType) -> ()) {
+
+    @discardableResult
+    func search(
+        query: String,
+        before: String? = nil,
+        containerWidth: CGFloat,
+        completion: @escaping (SearchResultType) -> ()
+        ) -> Cancellable {
         let query = SearchReposQuery(search: query, before: before)
-        
-        fetch(query: query) { (result, error) in
-            guard error == nil, result?.errors == nil else {
+
+        return fetch(query: query) { (result, error) in
+            guard (error == nil || (error! as NSError).code == -999), result?.errors == nil else {
                 ShowErrorStatusBar(graphQLErrors: result?.errors, networkError: error)
                 completion(.error)
                 return
--- a/Classes/Search/GithubClient+Search.swift
+++ b/Classes/Search/GithubClient+Search.swift
@@ -26,7 +26,7 @@ extension GithubClient {
         let query = SearchReposQuery(search: query, before: before)
 
         return fetch(query: query) { (result, error) in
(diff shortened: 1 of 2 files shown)


Classes/Search/GithubClient+Search.swift
```diff
@@ -7,19 +7,26 @@
...
+        containerWidth: CGFloat,
+        completion: @escaping (SearchResultType) -> ()
+        ) -> Cancellable {
         let query = SearchReposQuery(search: query, before: before)
-        
-        fetch(query: query) { (result, error) in
-            guard error == nil, result?.errors == nil else {
+
+        return fetch(query: query) { (result, error) in
+            guard (error == nil || (error! as NSError).code == -999), result?.errors == nil else {
```
Sherlouk: Could we refactor this slightly to drop the force unwrap? (I get that it's safe because of the nil check before, but would prefer not to have it if we can easily avoid it!)

Classes/Search/SearchViewController.swift
```diff
@@ -101,12 +102,13 @@ SearchResultSectionControllerDelegate {
     func search(term: String) {
         recentStore.add(recent: term)
 
-        state = .loading
         update(animated: false)
 
-        client.search(query: term, containerWidth: view.bounds.width) { [weak self] resultType in
+        let request = client.search(query: term, containerWidth: view.bounds.width) { [weak self] resultType in
+            guard let state = self?.state, case .loading(_) = state else { return }
```
Sherlouk: Do you still need the `(_)` here bit? I know in switches, if all parameters are unused you can just not specify them! `case .loading = state`

heshamsalman: Good catch! 👍

Classes/Search/GithubClient+Search.swift
heshamsalman: @Sherlouk 
How do you feel about global-scope private functions? 

```swift
extension GitHubClient { 
  ...
  guard errorIsNilOrCancelled(error), ...
  ...
}

private func errorIsNilOrCancelled(_ error: Error?) -> Bool {
    guard let error = error else { return true }
    return (error as NSError).code == -999
}
```

Sherlouk: I don't terribly mind this, alternatively just using optionals: `guard (error == nil || (error as? NSError)?.code == -999), result?.errors == nil else {` should have the same effect (I think)?

Leaving it up to you, as long as we have no pesky exclamation marks 😉

heshamsalman: @Sherlouk I'd love to use an optional here but `Error?` doesn't bridge to `NSError` -- the compiler asks that we use a hard-cast `as!` for that conversion

Classes/Search/SearchViewController.swift
```diff
@@ -186,10 +188,14 @@ SearchResultSectionControllerDelegate {
     }
 
     func searchBarCancelButtonClicked(_ searchBar: UISearchBar) {
+        if case let .loading(request) = state {
```
rnystrom: TIL!

PR #614, opened 2017-10-19
CFBundleVersion set to commit count

rnystrom:
Following [this post](http://www.mokacoding.com/blog/automatic-xcode-versioning-with-git/). Adding commit hash to a custom key in Info.plist so we can display + link to it later in settings alongside the version and build #.

### DIFF
--- a/Resources/Info.plist
+++ b/Resources/Info.plist
@@ -17,7 +17,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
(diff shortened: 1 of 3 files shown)


Resources/Info.plist
```diff
@@ -32,7 +32,7 @@
 		</dict>
 	</array>
 	<key>CFBundleVersion</key>
-	<string>2114</string>
+	<string>SEE BUILD PHASE</string>
```
Sherlouk: 😄

PR #615, opened 2017-10-19
Resign first responder when hiding keyboard

### LINKED ISSUE #610: Resign first responder on iPad when dismissing kb (when using an external keyboard)
<details>
<pre>
Version 1.11.0 (2101)
TestFlight: true
</pre>
</details>

BasThomas:
Fixes #610 

Ended up doing this when the keyboard will hide, so it happens at that point immediately. Thought that would make more sense then doing so when the view will disappear.

### DIFF
--- a/Classes/Search/SearchViewController.swift
+++ b/Classes/Search/SearchViewController.swift
@@ -44,6 +44,7 @@ SearchResultSectionControllerDelegate {
     init(client: GithubClient) {
         self.client = client
         super.init(nibName: nil, bundle: nil)
+         NotificationCenter.default.addObserver(self, selector: #selector(SearchViewController.willHideKeyboard), name: .UIKeyboardWillHide, object: nil)
     }
 
     required init?(coder aDecoder: NSCoder) {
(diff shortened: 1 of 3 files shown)


Classes/Search/SearchViewController.swift
```diff
@@ -44,6 +44,7 @@ SearchResultSectionControllerDelegate {
     init(client: GithubClient) {
         self.client = client
         super.init(nibName: nil, bundle: nil)
+         NotificationCenter.default.addObserver(self, selector: #selector(SearchViewController.willHideKeyboard), name: .UIKeyboardWillHide, object: nil)
```
rnystrom: nit: prefer params on their own line, makes it more readable

BasThomas:
@rnystrom how would I do that target-action with a notificationcenter observer?

rnystrom:
It’s just an objc selector, so you should be able to send a param-less method like `resignFirstResponder`. So `searchBar` as the target and `UISearchBar.resignFirstResponder` as the action? Something like that.

BasThomas:
Trying to do that, I can't seem to get around the `Argument of '#selector' does not refer to an '@objc' method, property, or initializer`, whatever I do.

BasThomas:
Ah, cool. Fixed.

rnystrom:
Super clean, love it.

BasThomas:
Thanks for the help as well! :D

Issue #597, opened 2017-10-17
Optimistic State Manager


Sherlouk:
A lot of actions have now got optimistic states, where we assume the network request will be successful and update the UI accordingly and then revert the change if we receive an error

The current solution isn't really scalable and is pretty easy to make mistakes with (Found that out the hard way 😅)

In most cases we only have one case of storage for a change, for example locking/closing. If you close and then open a issue you only see one view added to the timeline (open) because that's all we keep track of. If you refresh the feed, they all appear.

Just wondering if there's a better way to be doing this?

Some considerations:
- When we know it was successful, actually commit the change to the main model?
- The same model might be added multiple times (new comments, open/close, etc)
  - They might pass/fail in a different order than they were triggered
- Scalability for new actions

BasThomas:
> When we know it was successful, actually commit the change to the main model?

This seems like a great start.

rnystrom:
I wonder if we can merge this into a master in-memory storage, similar to this blog post:

http://khanlou.com/2017/10/the-flat-cache/

Apollo has a memory cache w/ signals about changes, but that wont work for us b/c of how we intermingle the V3 (REST) and V4 (GraphQL) GitHub APIs. A great example is comment editing: read w/ GQL, mutate w/ V3... 🤢

If we had a centralized store, we could add mutations to the store, the store emits changes w/ new, immutable models to listeners, and UIs can update.

> When we know it was successful, actually commit the change to the main model?

As long as we can do this in a way that doesn't introduce mutable models. There are already some mutable states in the app that I'm a little on the fence about (`NSAttributedStringSizing` being the biggest). _Some_ mutation is obviously unavoidable, but we def want to minimize.

rnystrom:
Further good reading from that article

http://redux.js.org/docs/recipes/reducers/NormalizingStateShape.html

https://facebook.github.io/relay/docs/thinking-in-graphql.html

rnystrom:
Started tinkering with this. My biggest dilemma is dealing with changes at the list-level.

Lets say the notifications VC listens for changes on `NotificationViewModel`s. When you change one to read = true, we mutate the cache, it sends an event, and the VC updates. Great.

But what about when you reload notification data? That would send an individual update event for every single one. We don’t want to spam with updates. In that case, we’d want to batch.

Maybe we can build update listeners in a way that they are individual or batches?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

weyert:
[...]

PR #623, opened 2017-10-20
Fix: persist local changes

afonsograca:
Whenever on a PR/Issue, when you lock it, the lock badge appears, if afterwards you close the issue, the open badge changes to close and we lose the locked badge. 

This happens because every time there's a lock or status changed event, the local changes are reset to the fetched state, which is not ideal, the local changes should be cumulative.

Steps to reproduce:

1. Lock a PR
2. Close a PR (the lock badge will disappear)
3. Refresh PR (both the close and the lock badge appear)

This solves it.

### DIFF
--- a/Classes/Issues/IssuesViewController.swift
+++ b/Classes/Issues/IssuesViewController.swift
@@ -330,7 +330,7 @@ IssueCommentSectionControllerDelegate {
     }
 
     func setStatus(close: Bool) {
-        guard let currentStatus = current?.status else { return }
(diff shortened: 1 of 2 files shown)


rnystrom:
Good reminder for us to get #597 working too. Would be great to just update models in a shared store instead of tracking all this local state.

afonsograca:
Sure, probably it has to do with the new version of [apollo-codegen](https://github.com/apollographql/apollo-codegen). Reverting it now.

PR #625, opened 2017-10-21
Add tap to dismiss keyboard in Notification ViewController

kimdv:
I have added tap to dismiss keyboard in the keyboard. Right now you have to drag to dismiss it. 😄

### DIFF
--- a/Classes/Notifications/NotificationsViewController.swift
+++ b/Classes/Notifications/NotificationsViewController.swift
@@ -59,6 +59,8 @@ TabNavRootViewControllerType {
         resetRightBarItem()
 
         navigationController?.tabBarItem.badgeColor = Styles.Colors.Red.medium.color
+
+        setupTapToDismissKeyboard()
     }
 
     override func viewWillAppear(_ animated: Bool) {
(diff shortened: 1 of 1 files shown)


rnystrom:
What does this do when tapping on a notification to view its content?

I haven’t run into this being an issue yet. UX bug or preference?

kimdv:
I don't think I understand the 

> What does this do when tapping on a notification to view its content?

I will not call it a bug. It's more a UX enhancement I think. Because it is possible to dismiss the keyboard by dragging the "No new notification view". Not it is possible to tap anywhere to dismiss the keyboard 😄 

simulator screen shot - iphone 8 - 2017-10-21 at 15 44 45

rnystrom:
@kimdv I mean, what happens when there is content in the collection view? I'm assuming that the collection view absorbs the tap and this event doesn't fire? Just making sure this doesn't interfere w/ any other interactions.

kimdv:
@rnystrom I pushed a fix for that. So now it should not cancel other touches 😄 

kimdv:
I can make a video with the use case ? 

rnystrom:
Probably ok w/out the video. I def see what you mean. Let's hold for a minute tho b/c of all the refactoring that just happened w/ `NotificationsViewController`. Even more incoming.

PR #643, opened 2017-10-22
Cancelled debounce action if explicit search

### LINKED ISSUE #638: Searching from empty state never finishes
Some bug with the search changes maybe. On a clear search it doesn’t finish loading and just spins forever. Once you cancel recent results display then everything works fine.

<details>
<pre>
Version 1.12.0 (1063)
TestFlight: false
</pre>
</details>

heshamsalman:
Closes #638

Classic race condition.

Apollo `FetchQueryOperation` is a class. 
Auto-search while typing is queuing up tasks to be fired after so many milliseconds. If the user hits enter to explicitly trigger a search during that time, we immediately start executing the search. In the mean-time, the debounced signal from the type-while-search fires, notices we're in a loading state and ends up cancelling the explicitly placed search. _because_ apollo queries are classes, and it's making the same network call, it ends up cancelling its own operation.

### DIFF
--- a/Classes/Search/SearchViewController.swift
+++ b/Classes/Search/SearchViewController.swift
@@ -202,6 +202,7 @@ SearchResultSectionControllerDelegate {
 
     func searchBarSearchButtonClicked(_ searchBar: UISearchBar) {
         searchBar.resignFirstResponder()
+        debouncer.cancel()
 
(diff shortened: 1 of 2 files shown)


rnystrom:
Thanks!!

Issue #647, opened 2017-10-22
Make Storyboard labels localizable in code


BasThomas:
Would help when translating all strings to support other languages instead of having to do so in a storyboard strings file

rnystrom:
Any tips for this? Storyboard n00b here.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BasThomas:
After thinking a bit I think it wouldn't hurt that bad if we but an extra `.strings` there, separating it. Another option would be to `@IBOutlet` everything, but that's also gross.

BasThomas:
Or remove the storyboard completely?

heshamsalman:
@BasThomas What I’m accustomed to doing is outletting them as private outlets and having a `didSet` block set their text/font/etc. 

Ditching it wouldn’t be the worst thing either haha.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #649, opened 2017-10-22
Add Dutch Localization

BasThomas:
@weyert, can you review please?

simulator screen shot - iphone x - 2017-10-22 at 10 03 21

The Settings aren't localized yet: #647


Sherlouk:
Wut! This is so cool, even if I don't understand a word of it 😅

weyert:
@BasThomas Sure, I will go through it :+

BasThomas:
@Sherlouk don't lie, you've got all the translations right next to it ;)

weyert:
Some comments about the translation. I also notice that you often would translate "View Files" as "Bekijk Bestanden". I think "Bekijk bestanden" without the capitalisation.

" added to milestone " = " toevegoed aan mijlpaal ";
> toegevoegd aan mijlpaal

" unassigned" = " afgewezen";
> Shouldn't this be “Niet-toegewezen”?

"%@ edited this issue %@" = "%1$@ heeft dit issue aangepast %2$@";
>I know that `dit issue` is correct too but it sounds a bit odd in this case. What about `de issue` or `deze issue`??

"Add Account" = "Voeg Account Toe”;
> “Nieuwe account” or maybe “Voeg een account toe” ?

> “Add reaction" = "Voeg reactie toe";
“Nieuwe reactie” or maybe “Voeg een nieuwe reactie toe’ ?

"All of your accounts will be signed out. Do you want to continue?" = "Al je accounts zullen uitgelogd worden. Wil je doorgaan?";
> What about: `Al je accounts zullen worden uitgelogd. Wil je doorgaan?`  ??

“Wis Allemaal”
> Wis allemaal? 

"Mark All" = "Markeer alle";
> Markeer alles ??

"Mark Alle Gelezen
> Markeer alles gelezen ?

“Geen Resultaten Gevonden”
> Geen resultaten gevonden ??

"Recent Searches" = "Recente Zoekopdrachten";
> Recente zoekopdrachten

"Sign In" = "Log In"
"Sign out" = "Log uit";
> Aanmelden / Afmelden ??

"Switch Account" = "Wissel Account";
> `Account wisselen` or `Wissel van account` ?

"There was an error adding another account. Please try again." = "Er trad een fout op bij het toevoegen van een ander account. Probeer opnieuw.";
> `Er trad een fout op tijdens het toevoegen van een ander account. Probeer opnieuw.`

BasThomas:
Oh damn, didn't realise that reviewing these sucks because the files aren't rendered by GitHub. 👎

Thanks for catching the typos and suggesting improvements.

On the capitalization: I haven't checked all I think, but most of those are used in alerts. From the [HIG](https://developer.apple.com/ios/human-interface-guidelines/controls/buttons/.):

> Use title-case for titles. Capitalize every word except articles, coordinating conjunctions, and prepositions of four or fewer letters.

BasThomas:
> Shouldn't this be “Niet-toegewezen”? I would translate Dismissed as Afgewezen (in this context, as we aren't firing people here).

Tricky. You assign and then unassign someone, but "niet toewijzen" means something else than unassigning. I'll keep it for now.

weyert:
> Really? I'd have to disagree here. Also, correctness > "sounds better" imo

I think they are all correct, see: https://www.welklidwoord.nl/deze-of-dit/issue

> On the capitalization: I haven't checked all I think, but most of those are used in alerts. From the HIG:
> > Use title-case for titles. Capitalize every word except articles, coordinating conjunctions, and prepositions of four or fewer letters.

Hmm, fair point. If I go to Settings > Accounts to add a new mail account it's showing 'Nieuwe account'  in the navigation bar. Shouldn't this be `Nieuwe Account` then? In my opinion this would be incorrect Dutch spelling.

Sherlouk:
Out of interest, how are we going to handle this moving forward? Strings file doesn’t auto-generate past first usage and with new strings being added all the time I feel like every release will have less and less localised text - I suggested a tool in the original ticket which would help us identify the strings and offer a solution without having to have separate PRs all the time but wonder what people think?

rnystrom:
Soooooo cool!!

sberrevoets:
FWIW, the HIG is specifically tailored toward (American) English. I think "Nieuw account" (without e after nieuw, too) is more stylistically correct than "Nieuw(e) Account". English uses capitalization much more often than Dutch (see also months, days, acronyms & initialisms), and I think this is one of those cases where title case is recommended in English, but not in Dutch.

BasThomas:
Ah, good catch @sberrevoets. Just checked and it indeed seems like Apple is not uppercasing everything in the OS when the language is set to Dutch. I'll change 👍

BasThomas:
Re: Out of interest, how are we going to handle this moving forward?

Good question. I think we should look into one of the options you provided and pick one that would work for us. @Sherlouk

weyert:
I agree that HIG mainly refers to the English language for these things. @sberrevoets @BasThomas  I have to admit I am not sure if it should be `nieuwe` or `nieuw` account. As far as I know it's one of those words which is both `de` en `het` word so we could use both `nieuwe` and `nieuw` and pick the most stylistic one. What shall we then follow the Dutch iOS language of system UI or go our own way?

BasThomas:
So iOS is saying `nieuwe account`? That seems strange to me. Because `het nieuwe account` is correct, but we are actually saying `een nieuw account` (but dropping the `een`). I’d say `nieuw`.

weyert:
Ja, ze gebruiken echt nieuwe account, zie ook hier:
https://support.apple.com/kb/PH6651?locale=en_GB&viewlocale=nl_NL 
https://support.apple.com/nl-nl/HT201320

BasThomas:
Oh wow. They’re using it all over the place as well. I’m torn.

Sherlouk:
I'm voting `nieuw` because in my naive attitude I'm pronouncing `nieuwe` like "no way" in a slightly welsh accent.

🤷‍♂️

BasThomas:
OK. Still totally conflicted, but in the name of constancy and "default behavior", let's go for "nieuwe account":

> In principe zijn Engelse leenwoorden mannelijke de-woorden, behalve als er een reden is om daarvan af te wijken.
> - een woord dat eindigt op *-isme* of *-ment*: *het hooliganisme*, *het agreement*, *het assessment*, *het management*;
> - namen van talen: *het slang*, *het Pascal*;
> - verzamelnamen: *het panel*, *het platform*;
> - stofnamen: *het denim*, *het plastic*, *het teak*;
> - woorden waarvan de betekenis (gedeeltelijk) overeenkomt met een Nederlands het-woord: *het schort* - *het shirt*;
> - woorden waarvan de betekenis overlapt in het Nederlands en Engels, en die ook een vormovereenkomst hebben: *het pocketbook* vanwege *het boek* (maar *de/het notebook*), *het copyright* vanwege *het recht*.

\*sigh\*
🤷‍♂️

weyert:
Okay, let's go for that then. Let's just keep it closely as Apple using it :)

weyert:
Looks good to me, @BasThomas @sberrevoets

Sherlouk:
Sorry for the commit spam, was hoping that using [Linguist](https://github.com/github/linguist) overrides we would be able to render the strings file.

Tried various different flags that I saw in my searches over GitHub, I've had success with it in different ways before so I'm not sure why it hates .strings so much! Might be because it's nested in a lproj which it also doesn't know. I'll explore in a different branch at another point in time 😄

rnystrom:
Decided to hold on this atm. Vast majority are US/English users, and once we sign up to support other languages we will need to support it longterm (new strings, etc). Want to merge this, but I need to come up w/ a better strategy re: l18n.

PR #652, opened 2017-10-22
Added insets handling for SearchViewController

Austinate:
Hi!
This PR adds keyboard handling so users will be able to scroll to the bottom of the list when keyboard is on the screen. Like here: 
screen shot 2017-10-22 at 20 18 41

### DIFF
--- a/Classes/Search/SearchViewController.swift
+++ b/Classes/Search/SearchViewController.swift
@@ -73,6 +73,20 @@ SearchResultSectionControllerDelegate {
         searchBar.backgroundColor = .clear
         searchBar.searchBarStyle = .minimal
         navigationItem.titleView = searchBar
+
+        let nc = NotificationCenter.default
+        nc.addObserver(
+            self,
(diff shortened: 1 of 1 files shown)


rnystrom:
Oh dang! Didn’t realize we weren’t adjusting for this. Awesome stuff.

Classes/Search/SearchViewController.swift
```diff
@@ -89,6 +103,24 @@ SearchResultSectionControllerDelegate {
...
+        let inset = UIEdgeInsets(top: 0, left: 0, bottom: bottomInset, right: 0)
+        collectionView.contentInset = inset
+        collectionView.scrollIndicatorInsets = inset
+    }
+
+    @objc
+    func onKeyboardWillHide(notification: NSNotification) {
+        collectionView.contentInset = .zero
+        collectionView.scrollIndicatorInsets = .zero
+    }
```
rnystrom: Note to self: let's create a `UIScrollView` extension that does this automatically

PR #678, opened 2017-10-22
Add issue type as title in action sheet

### LINKED ISSUE #651: Add issue type in alert message/title instead of separate actions
See https://github.com/rnystrom/GitHawk/pull/559#issuecomment-338466101

BasThomas:
Fixes #651

Suggestions welcome. It (still?) feels like some context is missing atm imo.

simulator screen shot - iphone x - 2017-10-22 at 22 26 45


Austinate:
@BasThomas wouldn't be nice to include at least issue number to the title (i bet that name can be too long)

Sherlouk:
Agree we should add the number so it's `Issue #123`. Completely unrelated, should the first option just be "Bookmark".

BasThomas:
@Austinate title can overflow, so that's not an issue. I'll add the `#`, though it's a bit unnecessary too (as you'll pretty much always see the nav with the number), but I guess it's better.

simulator screen shot - iphone x - 2017-10-22 at 22 37 38
(Meta!!!)

@Sherlouk: Agree, think this was discussed in the Bookmark ticket itself but overlooked. I'll create a PR 👍

PR #692, opened 2017-10-23
Adds Large title for bookmarks with search controller

rizwankce:
Tried out with the large title for IOS 11 w/ search controller. Feel free to close if this isn't what you expected. @rnystrom 

  
screen shot 2017-10-23 at 10 39 15 am


rizwankce:
@rnystrom here you go! conflicts free branch now.

Classes/Bookmark/BookmarksViewController.swift
```diff
@@ -37,7 +45,7 @@ TabNavRootViewControllerType {
 
         configureSearchBar()
         configureTableView()
-
+
```
rnystrom: Please change your whitespace settings 😆 

screen shot 2017-10-24 at 9 38 56 pm

rnystrom:
Woof, that view on smaller devices actually don't look very good. I'd rather be consistent though, let's try it out. Will make a judgement call before shipping 1.12 if we want to keep them.

BasThomas:
Re: smaller devices: we can tuck the under the nav bar by default, which I think would make sense. That'd create some space :)

rizwankce:
for now Im setting ```
hidesSearchBarWhenScrolling
``` to `false`. May be if we set it to `true` it would give some room for contents 

Classes/Bookmark/BookmarksViewController.swift
rizwankce: Including whitespace-only lines for turned off for me. Thx for the info 👍

PR #681, opened 2017-10-22
Rename "Add to Bookmark" to simply "Bookmark"

BasThomas:
As discussed [here](https://github.com/rnystrom/GitHawk/pull/678#issuecomment-338507125) and [here](https://github.com/rnystrom/GitHawk/pull/601#issue-266518955).

PR #696, opened 2017-10-23
Adds options to remove bookmark

### LINKED ISSUE #684: Hide bookmark option / toggle to "Remove Bookmark"
From https://github.com/rnystrom/GitHawk/pull/601#issuecomment-338176128

rizwankce:
Closes #684 

Probably overrides #681 

- Checks whether bookmark is present or not 
- Display destructive or default action with `Bookmark` or `Remove Bookmark` title
- Some sweet test for store


Classes/Views/Constants.swift
```diff
@@ -42,5 +42,6 @@ enum Constants {
         static let inbox = NSLocalizedString("Inbox", comment: "")
         static let upload = NSLocalizedString("Upload", comment: "")
         static let bookmark = NSLocalizedString("Bookmark", comment: "")
+        static let removeBookmark = NSLocalizedString("Remove Bookmark", comment: "")
```
rnystrom: Is this used anywhere else?

FreetimeTests/BookmarkStoreTests.swift
```diff
@@ -56,4 +56,12 @@ final class BookmarkStoreTests: XCTestCase {
 
         XCTAssert(BookmarksStore.shared.bookmarks.count == 0)
     }
+    
+    func test_containsBookmark() {
+        let b1 = BookmarkModel(type: .repo, name: "GitHawk", owner: "rizwankce")
+        
+        BookmarksStore.shared.add(bookmark: b1)
```
rnystrom: Gotta get rid of this singleton 😉

Classes/Utility/AlertAction.swift
```diff
@@ -143,8 +143,18 @@ struct AlertAction {
     }
 
     static func bookmark(_ bookmark: BookmarkModel) -> UIAlertAction {
-        return UIAlertAction(title: NSLocalizedString("Add To Bookmark", comment: ""), style: .default) { _ in
-            BookmarksStore.shared.add(bookmark: bookmark)
+        let isNewBookmark = !BookmarksStore.shared.contains(bookmark: bookmark)
+        let title = isNewBookmark ? Constants.Strings.bookmark : Constants.Strings.removeBookmark
+        return UIAlertAction(
+            title: title,
+            style: isNewBookmark ? .default : .destructive
```
rnystrom: Actually can we just make this `.default`? Seeing the destructive "red" next to other actions like "Close" and "Lock" is a little weird b/c this doesn't carry _nearly_ the same weight. cc @BasThomas for thoughts

rnystrom: Actually can we just make this `.default`? Seeing the destructive "red" next to other actions like "Close" and "Lock" is a little weird b/c this doesn't carry _nearly_ the same weight. cc @BasThomas for thoughts

BasThomas: I think it would make sense, but I think we’d like to move the action away from lock/close. Also, both Bookmark and Remove Bookmark have “side effects”, so (as we’re using it right now) both actions should be “destructive”. 

I feel like it’s time to explore something completely different than the (now very long) action sheets...

Classes/Views/Constants.swift
rizwankce: Nope. I just don't like to have random string in AlertActions

Classes/Utility/AlertAction.swift
rizwankce: "Remove bookmark" does not carry the same weight as other destructive buttons. But IMO it would avoid accidentally tapping. 

From Apple doc
> case destructive
>Apply a style that indicates the action might change or delete data.

Also, I do agree with @BasThomas. Action sheet is very long now. After 1.12.0 we should rethink and adopt a better way to handle this action sheet for sure

Classes/Views/Constants.swift
BasThomas: Although I get it, I think it would be more in line to keep single-use strings "hardcoded" as to not clutter the Constants *even more*.

FreetimeTests/BookmarkStoreTests.swift
rizwankce: Yes Yes. I just don't want to give a big PR with all changes. Wanted to give a separate/small PR after I fix the initial round of bug fixes. So it would be easy for us to review/conflicts free PR.

Classes/Utility/AlertAction.swift
BasThomas: What do you think about moving the action (more) to the bottom? I think it doesn't make sense that it is next (and in this case even above) some probably more used, destructive actions.

BasThomas: On the discussion regarding "destructive": https://github.com/rnystrom/GitHawk/pull/532

rizwankce: You mean "Remove bookmark" action to the bottom of the action sheet? 
Yes, I'm down with that.

BasThomas: Yep, the add + remove bookmark :)

rizwankce: Moved bookmark actions to bottom of the action sheet. 👍

rnystrom:
re: action sheet growing, I'm brainstorming ways to move all edit/action functionality into some sort collapsed feature set. Same w/ editing labels/milestones/etc. Would be nice to put it all in one place that is tucked away.

PR #699, opened 2017-10-23
3D touch short cut item for bookmarks tab

### LINKED ISSUE #694: 3D touch short cut to Bookmark Tab

rizwankce:
Closes #694

### DIFF
--- a/Classes/Systems/ShortcutHandler.swift
+++ b/Classes/Systems/ShortcutHandler.swift
@@ -11,9 +11,11 @@ import Foundation
 struct ShortcutHandler {
 
     static private let searchViewControllerIndex = 1
+    static private let bookmarksViewControllerIndex = 2
 
     private enum Items: String {
         case search
+        case bookmarks
         case switchAccount
     }
 
@@ -28,6 +30,9 @@ struct ShortcutHandler {
         switch itemType {
         case .search:
             navigationManager.selectViewController(atIndex: searchViewControllerIndex)
+            return true
+        case .bookmarks:
+            navigationManager.selectViewController(atIndex: bookmarksViewControllerIndex)
             return true
         case .switchAccount:
             if let index = item.userInfo?["sessionIndex"] as? Int {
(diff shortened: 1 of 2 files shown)


Classes/Views/Constants.swift
```diff
@@ -41,5 +41,6 @@ enum Constants {
         static let delete = NSLocalizedString("Delete", comment: "")
         static let inbox = NSLocalizedString("Inbox", comment: "")
         static let upload = NSLocalizedString("Upload", comment: "")
+        static let bookmark = NSLocalizedString("Bookmark", comment: "")
```
rnystrom: Is this used anywhere else yet? Rule of thumb is to only put something in `Constants.swift` when its used >1.

Classes/Systems/ShortcutHandler.swift
```diff
@@ -46,10 +51,17 @@ struct ShortcutHandler {
...
                                                    localizedTitle: Constants.Strings.search,
                                                    localizedSubtitle: nil,
-                                                   icon: searchIcon,
-                                                   userInfo: nil)
+                                                   icon: searchIcon)
         items.append(searchItem)
 
+        // Bookmarks
+        let bookmarkIcon = UIApplicationShortcutIcon(templateImageName: "bookmarks")
+        let bookmarkItem = UIApplicationShortcutItem(type: Items.bookmarks.rawValue,
```
rnystrom: nit: will fix this up myself, but should put all params on newline [like this](https://github.com/rnystrom/GitHawk/blob/master/Classes/Issues/IssuesViewController.swift#L118-L125)

Classes/Systems/ShortcutHandler.swift
```diff
@@ -11,9 +11,11 @@ import Foundation
 struct ShortcutHandler {
 
     static private let searchViewControllerIndex = 1
+    static private let bookmarksViewControllerIndex = 2
```
rnystrom: yikes, these are going to get out of hand. very very easy to break.

PR #700, opened 2017-10-23
Settings view source redirects to Repository view Code Tab

rizwankce:
Since we do have code browsing, It's nice to redirect our settings "view source" button to repository view code tab. In case user still wants to see repo in safari, they can use more button on top right in repo screen.

### DIFF
--- a/Classes/Repository/RepositoryViewController.swift
+++ b/Classes/Repository/RepositoryViewController.swift
@@ -12,6 +12,13 @@ import Pageboy
 import TUSafariActivity
 import SafariServices
 
+enum RepositoryPageIndex: Int {
+    case overview    = 0
+    case issues      = 1
+    case pullRequest = 2
+    case code        = 3
+}
+
 class RepositoryViewController: TabmanViewController,
 PageboyViewControllerDataSource,
 NewIssueTableViewControllerDelegate {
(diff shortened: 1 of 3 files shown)


Classes/Repository/RepositoryViewController.swift
```diff
@@ -12,18 +12,27 @@ import Pageboy
 import TUSafariActivity
 import SafariServices
 
+enum RepositoryPageIndex: Int {
+    case overview    = 0
+    case issues      = 1
```
Sherlouk: Issues is not always available (see the `controllers` array in the init) so these numbers will immediately be out in such a repository

rizwankce: 🤒  forgot about that totally! Gonna make some change

rizwankce: I'm stuck 😢  @Sherlouk do you have any way to tackle this w/ or w/out `enum`?

Sherlouk: RepositoryPageIndex could have a fileprivate var which is the view controller type (not an initialised version) e.g.  `RepositoryIssuesViewController.self`. Then `defaultPage(for:)` is the index of the first item in the `controllers` array where item is of type `defaultPageIndex.viewController`

rizwankce: but we use `RepositoryIssuesViewController` for both issue and pull request. So when we want to default index to be a pull request it will fail. isn't?

Sherlouk: Fair point, I didn't realise that - Maybe a protocol then which specifies the PageType so you can iterate through the controllers array and find the first item that matches PageType with the enum you pass through for default.

Classes/Repository/RepositoryViewController.swift
```diff
@@ -12,18 +12,27 @@ import Pageboy
...
+    case pullRequest = 2
+    case code        = 3
```
rnystrom: What's to stop this from getting out of sync w/ the actual pages? This will be really easy to break and the compiler wont help us.

rnystrom:
I actually think we should pause this for now. The code browser atm is super limited, and there's no way to open Safari from it. I'd rather ppl be able to view the source, open Safari, and share stuff.

Ok w/ closing for now? We can revive once the code browsing is much more robust.

PR #711, opened 2017-10-23
Adds light gray text color on selection of LoadMoreCell

joerocca:
Gives some indication that load more has been pressed.


rnystrom:
Took some liberties, I like the `light` gray style on the button 😋

PR #716, opened 2017-10-24
Added labels equatable check to ListDiffable isEqual method

joerocca:
Fixes #715

PR #725, opened 2017-10-25
Faster label summary equality for diffing

rnystrom:
Followup from #716


joerocca:
Nice!
We should be able to remove the Equatable conformance from RepositoryLabel using this method?

Classes/Repository/RepositoryIssueSummaryModel.swift
```diff
@@ -38,6 +41,7 @@ class RepositoryIssueSummaryModel: ListDiffable {
         self.status = status
         self.pullRequest = pullRequest
         self.labels = labels
+        self.labelSummary = labels.reduce("", { $0 + $1.name })
```
BasThomas: *Maybe* `{ "\($0)\($1.name)" }`? No big deal though.

PR #743, opened 2017-10-25
Removed Equatable comformance from RepositoryLabel

joerocca:
No longer necessary after #725


BasThomas:
What do we win by removing this though?

joerocca:
Nothing
Just cleaning up since it was only added for the equality check used in #716

PR #748, opened 2017-10-26
Improve code browsing accessibility

BasThomas:
Ref #118


Sherlouk:
Should we also be adding hints for some of these things? "Opens Directory" "Opens File"?

Issue #754, opened 2017-10-26
“Bookmark” title for 3D Touch action should be “Bookmarks”


BasThomas:
<details>
<summary>Bug Report Dump (Auto-generated)</summary>
<pre>
Version 1.12.0 (1106)
Device: iPhone 7 Plus (iOS 11.0.3)
TestFlight: true
</pre>
</details>

PR #766, opened 2017-10-26
Update Bookmark quick action title

ajfigueroa:
Fixes https://github.com/rnystrom/GitHawk/issues/754

- Updates the Bookmark quick action title to the plural `"Bookmarks"` vs `"Bookmark"` 

### Before

simulator screen shot - iphone 8 - 2017-10-26 at 17 44 15

### After
simulator screen shot - iphone 8 - 2017-10-26 at 17 46 30

### DIFF
--- a/Classes/Systems/ShortcutHandler.swift
+++ b/Classes/Systems/ShortcutHandler.swift
@@ -57,7 +57,7 @@ struct ShortcutHandler {
         // Bookmarks
         let bookmarkIcon = UIApplicationShortcutIcon(templateImageName: "bookmarks")
         let bookmarkItem = UIApplicationShortcutItem(type: Items.bookmarks.rawValue,
-                                                   localizedTitle: Constants.Strings.bookmark,
+                                                   localizedTitle: Constants.Strings.bookmarks,
                                                    localizedSubtitle: nil,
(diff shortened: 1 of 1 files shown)


Classes/Systems/ShortcutHandler.swift
```diff
@@ -57,7 +57,7 @@ struct ShortcutHandler {
         // Bookmarks
         let bookmarkIcon = UIApplicationShortcutIcon(templateImageName: "bookmarks")
         let bookmarkItem = UIApplicationShortcutItem(type: Items.bookmarks.rawValue,
```
ajfigueroa: Should I update these variable names to match the plural form too?
That is, `bookmarksIcon` and `bookmarksItem`.

BasThomas: Don't think that's needed 👍

rnystrom:
Why is the icon a dot?

rizwankce:
@rnystrom When we give an asset name which can not be found in image assets. OS is adding DOT as image. 

ajfigueroa:
I think this will be fixed in https://github.com/rnystrom/GitHawk/pull/737

PR #780, opened 2017-10-27
Inject bookmarks, cleanup

rnystrom:
Followup from #738

- Dependency inject a single object
- Don't init if token doesn't exist
- Move action to helper func so no duped code

cc @rizwankce


rizwankce:
Lovely 😮  👍

Issue #771, opened 2017-10-27
Specify that image upload title and message are optional


BasThomas:
![GitHawk Upload by Sherlouk](https://i.imgur.com/IiQPWAV.jpg)

<details>
<summary>Bug Report Dump (Auto-generated)</summary>
<pre>
Version 1.12.0 (1106)
Device: iPhone 7 Plus (iOS 11.0.3)
TestFlight: true
</pre>
</details>

rnystrom:
Agree

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #785, opened 2017-10-28
Specify that image upload title and message are optional

DamascenoRafael:
Add “optional” in the title's placeholder.
And since the message is a TextView I added UITextViewDelegate methods to change text and textColor when needed.

ref #771

### DIFF
--- a/Classes/Image Upload/ImageUploadTableViewController.swift
+++ b/Classes/Image Upload/ImageUploadTableViewController.swift
@@ -7,6 +7,7 @@
 //
 
 import UIKit
+import SlackTextViewController
 
 protocol ImageUploadDelegate: class {
     func imageUploaded(link: String, altText: String)
@@ -21,7 +22,7 @@ class ImageUploadTableViewController: UITableViewController {
         }
     }
     @IBOutlet private var titleTextField: UITextField!
-    @IBOutlet private var bodyTextField: UITextView!
+    @IBOutlet private var bodyTextField: SLKTextView!
     
     private var bodyPlaceholder: String?
     private var bodyTextColor: UIColor?
@@ -73,13 +74,6 @@ class ImageUploadTableViewController: UITableViewController {
 
         // Set the left button item to cancel
         setLeftBarItem()
-
-        // Save placeholder and text color for body field
-        bodyPlaceholder = bodyTextField.text
-        bodyTextColor = bodyTextField.textColor
-        
-        // Set body field delegate so placeholder appears and disappears when needed
-        bodyTextField.delegate = self
         
         // Compress and encode the image in the background to speed up the upload process
         image.compressAndEncode { [weak self] result in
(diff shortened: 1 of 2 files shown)


Sherlouk:
Sorry, I may be missing something - Why are we not using the placeholder attribute that text fields already provide us with?

DamascenoRafael:
Hey @Sherlouk, I'm using the placeholder attribute in the title (titleTextField), but the message (bodyTextField) is actually a UITextView and not a UITextField. I can change the name of the outlet if you want, I think it would be better if we called bodyTextView.

Sherlouk:
Ahh nah I just didn't realise there was no placeholder on text view 😅 Assumed it would be there, my bad!

DamascenoRafael:
That's ok 😉

Classes/Image Upload/ImageUploadTableViewController.swift
```diff
@@ -186,13 +196,35 @@ class ImageUploadTableViewController: UITableViewController, UITextFieldDelegate
             }
         }
     }
+    
+}
 
-    // MARK: UITextFieldDelegate
+// MARK: UITextFieldDelegate
 
+extension ImageUploadTableViewController: UITextFieldDelegate {
```
rnystrom: Can we actually avoid adding extensions w/in the same file? Trying to keep the code style consistent.

Classes/Image Upload/ImageUploadTableViewController.swift
```diff
@@ -186,13 +196,35 @@ class ImageUploadTableViewController: UITableViewController, UITextFieldDelegate
...
         bodyTextField.becomeFirstResponder()
         return false
     }
+}
 
+// MARK: UITextViewDelegate
+
+extension ImageUploadTableViewController: UITextViewDelegate {
+    func textViewDidBeginEditing(_ textView: UITextView) {
+        if textView.text == bodyPlaceholder {
```
rnystrom: Maybe we should just start w/ `Optional` in the text field and add this later if totally necessary. Doing this text comparison is pretty inefficient. `SLKTextView` has an implementation for placeholders w/ `UITextView` already too.

Classes/Image Upload/ImageUploadTableViewController.swift
BasThomas: I did the same in #746 though, fwiw

Classes/Image Upload/ImageUploadTableViewController.swift
DamascenoRafael: I didn't know about `SLKTextView`. I'll take a look and try to do with it, ok?

BasThomas: See here: https://github.com/slackhq/SlackTextViewController/blob/master/Source/SLKTextView.h

Classes/Image Upload/ImageUploadTableViewController.swift
DamascenoRafael: @rnystrom Sure! But I've seen some style guides recommending this way - https://github.com/raywenderlich/swift-style-guide#protocol-conformance

DamascenoRafael: I changed the SLKTextView. There's still the UITextFieldDelegate extension, do you want me to put it without the extension?

rnystrom: I actually ran into compilation issues b/c of overuse of `extension` and ObjC interop, so I'm just moving away from them as much as possible 😋 .

PR #789, opened 2017-10-29
Constraint clean up

joerocca:


rnystrom:
Sweet! Much cleaner. I hate the VFL.

PR #792, opened 2017-10-29
added optional value handler #771

askari01:


BasThomas:
Hey @askari01 - just a heads up that #785 seems to try and accomplish the same.

rnystrom:
Merged #785

PR #795, opened 2017-10-29
Better FlatCache usage in notifications

rnystrom:
Mutations was a little gross. Cleans that up a bit so we're always fetching the latest model from the cache when something changes.

### DIFF
--- a/Classes/Notifications/NotificationsViewController.swift
+++ b/Classes/Notifications/NotificationsViewController.swift
@@ -23,8 +23,6 @@ FlatCacheListener {
     private let client: NotificationClient
     private let foreground = ForegroundHandler(threshold: 5 * 60)
     private let showRead: Bool
-
-    // todo
     private var notificationIDs = [String]()
 
     // set to nil and update to dismiss the rating control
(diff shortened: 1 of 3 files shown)


Classes/Notifications/NotificationsViewController.swift
```diff
@@ -214,30 +245,7 @@ FlatCacheListener {
...
-                    } else {
-                        models.append(model)
-                    }
-                }
-                self.update(models: models, animated: true)
-                self.resetLeftBarItem()
-            }
-        case .list: break
-        }
+        self.update(animated: true)
```
rnystrom: This needs to only update on `.item` otherwise it'll update for _every_ object when refreshed.

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -24,6 +24,9 @@ FlatCacheListener {
     private let foreground = ForegroundHandler(threshold: 5 * 60)
     private let showRead: Bool
 
+    // todo
```
BasThomas: What's up with this guy?

Classes/Repository/RepositoryIssuesViewController.swift
```diff
@@ -87,6 +88,10 @@ class RepositoryIssuesViewController: BaseListViewController<NSString>, BaseList
         return [searchKey]
     }
 
+    func models(listAdapter: ListAdapter) -> [ListDiffable] {
```
BasThomas: Why/where do we use this?

Classes/Notifications/NotificationsViewController.swift
rnystrom: Ahahaha, you know what, I have no idea.

Classes/Repository/RepositoryIssuesViewController.swift
rnystrom: See new required function for `BaseListViewControllerDataSource`

PR #797, opened 2017-10-30
Add Cell setup code to an StyledTableCell Extension

### LINKED ISSUE #749: Create UITableViewCell subclass for code browsing cells
We should move the setup for the code browsing cells to a separate subclass to not overcomplicate the ViewController.

Code is [here](https://github.com/rnystrom/GitHawk/blob/master/Classes/Repository/RepositoryCodeDirectoryViewController.swift#L102-L112).

Nixsm:
This fixes #749, sorry I've messed up the other Pull request

### DIFF
--- a/Classes/Repository/RepositoryCodeDirectoryViewController.swift
+++ b/Classes/Repository/RepositoryCodeDirectoryViewController.swift
@@ -99,28 +99,10 @@ final class RepositoryCodeDirectoryViewController: UIViewController, UITableView
     }
 
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        let cell = tableView.dequeueReusableCell(withIdentifier: cellIdentifier, for: indexPath)
+        let cell = tableView.dequeueReusableCell(withIdentifier: cellIdentifier, for: indexPath) as! StyledTableCell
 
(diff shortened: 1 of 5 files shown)


Classes/Repository/RepositoryCodeDirectoryViewController.swift
```diff
@@ -99,28 +99,10 @@ final class RepositoryCodeDirectoryViewController: UIViewController, UITableView
     }
 
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        let cell = tableView.dequeueReusableCell(withIdentifier: cellIdentifier, for: indexPath)
+        let cell = tableView.dequeueReusableCell(withIdentifier: cellIdentifier, for: indexPath) as! StyledTableCell
```
rnystrom: Eeeeep force-unwrap 💥☠️🚫 Can you handle this w/out forcing? Either `fatalError` or
```swift
if let cell = cell as? StyledTableCell {
  cell.setup(with: file)
}
```

Nixsm: using a if let would not work, because we can't return `nil`, so I decided to use `fatalError`

PR #807, opened 2017-10-31
Toggle hiding for details view controller

### LINKED ISSUE #799: Option to hide the master view on iPad
There are screens, It would be really helpful to hide/unhide the master view for our split view controller. So that detail view will be visible full screen on iPad. 

The screens I would like to hide master view are below
 - Code browsing 
 - Pull request file browser

Quickly tested something on iPad SIM:

screen shot 2017-10-30 at 11 01 18 am

rizwankce:
Closes #799 

After some experiments found out that we can use `displayModeButtonItem` from `splitViewController` to hide/unhide master view controller. So instead of adding this toggle button to a specific view controller, added this to top most Detail view controller. 

- This will not affect anything on iPhone
- Once master view is hidden, we will get back button to get back the master-detail view

Sample in action:

hide-split-view

### DIFF
--- a/Classes/Bookmark/BookmarksViewController.swift
+++ b/Classes/Bookmark/BookmarksViewController.swift
@@ -158,6 +158,7 @@ TabNavRootViewControllerType {
         default:
             return
         }
+        destinationViewController.navigationItem.configure(splitViewController?.displayModeButtonItem)
         let navigation = UINavigationController(rootViewController: destinationViewController)
         showDetailViewController(navigation, sender: nil)
     }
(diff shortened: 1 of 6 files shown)


BasThomas:
Just as a heads up: this does affect iPhone Plus Devices, but that should be expected. 

rizwankce:
Yes! Forgot to mention it. But it is expected. 

rnystrom:
I actually had this originally and pulled it out. I’m not really optimizing the design for the iPad on full screen, and think there’s plenty of usable room atm. Thanks tho!

PR #809, opened 2017-10-31
Adds double tap comment to 👍

### LINKED ISSUE #802: Double tap comment to 👍
<details>
<pre>
Version 1.12.0 (1106)
TestFlight: false
</pre>
</details>

joerocca:
Not sure if this is the best way to accomplish this. Please review!

Closes #802


Classes/Views/DoubleTappableCell.swift
```diff
@@ -0,0 +1,45 @@
...
+//
+
+import UIKit
+import Foundation
+
+protocol DoubleTappableCellDelegate: class {
+    func didDoubleTap(cell: DoubleTappableCell)
+}
+
+class DoubleTappableCell: UICollectionViewCell, UIGestureRecognizerDelegate {
```
heshamsalman: Spit-balling here, but wouldn't this also be useful as a protocol? 

Such that if we need a SwipeableDoubleTappableCell we can just implement a hypothetical `Swipeable` and `DoubleTappable`? 

```swift
protocol DoubleTappable: AnyObject, UIGestureRecognizerDelegate {
  var doubleTapGestureRecognizer: UITapGestureRecognizer { get set }
  var doubleTapDelegate: DoubleTappableCellDelgate? { get set }
}

extension DoubleTappable {
  func doubleTapAction() {
    doubleTapDelegate?.didDoubleTap(view: self)
  }

  func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldRecognizeSimultaneouslyWith otherrGestureRecognizer: UIGestureRecognizer) -> Bool {
    return true
  }
}
```

heshamsalman: Could even get cute with the gesture recognizer (though not sure it's of _that much value_):

```swift
func DoubleTapGestureRecognizer(containingView: DoubleTappable, delegate: DoubleTappableDelegate) -> UIGestureRecognizer {
    let gestureRecognizer = UITapGestureRecognizer()
    gestureRecognizer.numberOfTapsRequired = 2
    gestureRecognizer.delegate = delegate
    gestureRecognizer.addTarget(containingView, action: #selector(containingView.doubleTapAction))
    return gestureRecognizer
}
```

rnystrom:
Lol this is EXACTLY what I was thinking. Great stuff!

Classes/Views/DoubleTappableCell.swift
joerocca: @heshamsalman So I can't do this because Swift protocol extensions are invisible to ObjC. 

https://stackoverflow.com/a/40297075/4645381

heshamsalman:
@joerocca thanks for investigating and entertaining the notion 😁. Otherwise LGTM 

BasThomas:
Is this blocked by anything?

rnystrom:
Keep forgetting to comment. We need an option to disable these so they no-op in PR review comments. Then we’re good!

joerocca:
Should be good to go!

Classes/Issues/Comments/IssueCommentSectionController.swift
```diff
@@ -281,7 +281,8 @@ DoubleTappableCellDelegate {
             cell.delegate = self
         }
         
-        if let cell = cell as? DoubleTappableCell {
+        if !(viewController is PullRequestReviewCommentsViewController),
```
rnystrom: Ack, strongly discourage reaching into the VC like this. If the VC is renamed, this gets refactored, or used in another surface, it will silently break.

There's already a flag that we use to show reactions or not here:

https://github.com/rnystrom/GitHawk/blob/master/Classes/Issues/Comments/IssueCommentSectionController.swift#L204-L206

Can we just reuse that flag?

joerocca:
Oops! Didn't read the warning in the description.
Now it should be good!

Classes/Issues/Comments/IssueCommentSectionController.swift
```diff
@@ -281,7 +281,8 @@ DoubleTappableCellDelegate {
             cell.delegate = self
         }
         
-        if !(viewController is PullRequestReviewCommentsViewController),
+        if let object = self.object,
+            !object.asReviewComment,
```
rnystrom: lovely!

PR #720, opened 2017-10-24
Unified deletion swipes

heshamsalman:
- Pulled workaround for iOS 11 & UITableViewCells from SwipeCellKit (https://github.com/SwipeCellKit/SwipeCellKit/issues/48)  (ref #679 ?)
- Made destructive swipe behavior generic
- Changed destructive swipe animation
- Matched search and bookmark cell heights
- Rolled in changes from #717 and #698 

Closes #709 #710 #705 #659

PR #819, opened 2017-11-01
Bookmark cell dynamic size and fix swipe to delete crash

### LINKED ISSUE #812: BookmarksViewController cells should have dynamic size
When a cell has large content it can overflow out of the bounds of the cell.

img_0004

rizwankce:
Closes #812 

Guess #720 causes some issues with bookmark cell dynamic sizes. This will fix that and a bonus point it will fix crash on swipe to delete and a custom bookmark cell.

### DIFF
--- a/Classes/Bookmark/BookmarkCell.swift
+++ b/Classes/Bookmark/BookmarkCell.swift
@@ -27,7 +27,7 @@ class BookmarkCell: SwipeSelectableTableCell {
     }
 
     override init(style: UITableViewCellStyle, reuseIdentifier: String?) {
-        super.init(style: .subtitle, reuseIdentifier: reuseIdentifier)
+        super.init(style: style, reuseIdentifier: reuseIdentifier)
         accessibilityTraits |= UIAccessibilityTraitButton
         isAccessibilityElement = true
         titleLabel.numberOfLines = 0
--- a/Classes/Bookmark/BookmarkCell.swift
+++ b/Classes/Bookmark/BookmarkCell.swift
@@ -7,12 +7,18 @@
 //
 
 import UIKit
+import SnapKit
 
 class BookmarkCell: SwipeSelectableTableCell {
 
     static var cellIdentifier: String {
         return "bookmark_cell"
     }
+
+    private let reasonImageView = UIImageView()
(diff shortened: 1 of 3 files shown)


rizwankce:
@heshamsalman didn't see that coming! checking right now and will make changes

rizwankce:
IOS 11 has default `UITableViewAutomaticDimension` so self-sizing is enabled by default. However, I'm facing issues on iOS 10 for `UITableViewCellStyle` as `subTitle`. Setting `UITableViewAutomaticDimension` is not actually working.

rizwankce:
So default table view cell details label don't have self-sizing on < iOS 11. Since we support from iOS 10. I went ahead and created custom cell + layout. Tested on both iOS 11 and iOS 10 and it just works.

rnystrom:
@heshamsalman @rizwankce any concerns w/ how much bookmarks is deviating away from the way we build things in the rest of the app? This requires a complete 180 context shift to work on now 😕 

If we were using more specific `UITableView` features I wouldn't mind, but we aren't. e.g. swiping is done w/ SwipeCellKit, and we're not using editing mode...

There's also the fact that we could just reuse PR/issue and repo cells from other products and standardize things a bit more.

I'll land this as-is, just calling it out. Forwarning that I'm probably going to 100% rewrite this in the future 😛

heshamsalman:
@rnystrom yeah. Bit weird to go from IGLK mode to table mode. I actually have a branch with a re-write in IGLK but it's definitely gonna need a rebase.

The big thing that bugs me -- more than the context switch -- is pulling in the hack from SwipeCell to get it to work on tables in iOS 11 as well as the duplication around it.

Regards,
Hesham

rnystrom:
@heshamsalman that's a great point

rnystrom:
Couple followups re: sizing:

- min height needs to be larger
- v-alignment looks off (label should be vertically centered)

img_6610

rizwankce:
When we started bookmarks. It was simple table view cell with editing mode enabled. But down the line with self sizing and swipe to delete we kind of diviated a lot. I totally agree we should use Iglk. Since there are many ways we can reuse things within our code base. 

PR #825, opened 2017-11-01
Sorted/Ordered Bookmarks

### LINKED ISSUE #822: Weired Bookmarks behaviour when having more bookmarks
- Add more than 3 bookmarks (probably issue and repo) 
- constantly tap between search and bookmark tab 

Notice that bookmarks order is changing and on click it's going to wrong bookmark then what it displays.

rizwankce:
Closes #822 

Learnt some nifty details. Oversight from my side. Hope I'm not alone

### DIFF
--- a/Classes/Bookmark/BookmarkStore.swift
+++ b/Classes/Bookmark/BookmarkStore.swift
@@ -75,7 +75,7 @@ final class BookmarksStore {
 
     var bookmarks: [BookmarkModel] {
         refresh()
-        return Array(_bookmarks)
+        return Array(_bookmarks.sorted(by: { "\($0.owner)\($0.name)\($0.number)" > "\($1.owner)\($1.name)\($1.number)"}))
     }
 
     // MARK: Private API
--- a/Classes/Bookmark/BookmarkStore.swift
+++ b/Classes/Bookmark/BookmarkStore.swift
@@ -36,7 +36,7 @@ final class BookmarksStore {
     }
 
     private var token: String
-    private var _bookmarks: Set<BookmarkModel> = []
+    private var _bookmarks: NSMutableOrderedSet = []
     private var listeners: [ListenerWrapper] = []
 
     // MARK: Init
@@ -55,7 +55,7 @@ final class BookmarksStore {
     }
 
     func add(bookmark: BookmarkModel) {
-        _bookmarks.insert(bookmark)
+        _bookmarks.add(bookmark)
         archive()
     }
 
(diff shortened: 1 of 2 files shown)


Classes/Bookmark/BookmarkStore.swift
```diff
@@ -75,7 +75,7 @@ final class BookmarksStore {
 
     var bookmarks: [BookmarkModel] {
         refresh()
-        return Array(_bookmarks)
+        return Array(_bookmarks.sorted(by: { "\($0.owner)\($0.name)\($0.number)" > "\($1.owner)\($1.name)\($1.number)"}))
```
BasThomas: What about making `BookMarkModel` conform to `Comparable` and using `>`?

Sherlouk: Also is sorting the array _every_ time we fetch this value the best idea? I wonder if there's a better place we could do this?

rnystrom:
This will sort every single time `bookmarks` is called. Maybe instead it should sort when you mutate the store?

rizwankce:
Guess we will end up in the same situation when we get the bookmarks.  I'll give it a try though 

rizwankce:
After a bit of rethinking, We actually need a collection which can avoid duplicates and maintain the order. So end up giving away `Set<BookmarkModel>` and added `NSMutableOrderedSet`.

So by this way, we can avoid adding `sorted` method and some overkills comes along with it.

**Note** 
If this PR accepted. Possibly all your previous bookmarks will be lost.

Classes/Bookmark/BookmarkStore.swift
```diff
@@ -36,7 +36,7 @@ final class BookmarksStore {
     }
 
     private var token: String
-    private var _bookmarks: Set<BookmarkModel> = []
+    private var _bookmarks: NSMutableOrderedSet = []
```
rnystrom: wo TIL this syntax works

Classes/Bookmark/BookmarkStore.swift
```diff
@@ -69,19 +69,23 @@ final class BookmarksStore {
...
     }
 
     var bookmarks: [BookmarkModel] {
         refresh()
-        return Array(_bookmarks)
+        guard let bookmarks = _bookmarks.array as? [BookmarkModel] else {
+            return []
+        }
+
+        return bookmarks
```
rnystrom: great!

Issue #337, opened 2017-09-21
Pull request review threads GQL broken


rnystrom:
This has been broken on GitHub's end for almost 2 months. Been trying to [get an answer here](https://platform.github.community/t/pullrequestreviewthread-no-longer-showing-up-in-timeline-for-prs/2755) but it's not going well.

Probably have to play around w/ other ways to get PR review threads and contents. Bummer.

rnystrom:
Got a [response](https://platform.github.community/t/pullrequestreviewthread-no-longer-showing-up-in-timeline-for-prs/2755/16), tho its basically "known bug, no ETA".

BasThomas:
It’s better than nothing at least :)

Sherlouk:
How has this still not been fixed 🙄 😞 

The fact this isn't a feature really annoys me, more so because there's nothing we can do about it 😕

rnystrom:
I know I’m losing my mind not having these. Gonna try to hack a fix today... 😭

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Sherlouk:
I've already spent an hour trying to find every alternative I could, even looking at the v3 API previews 😅 Couldn't find anything

rnystrom:
Lol ya, its totally broken. Updated w/ a reply:

https://platform.github.community/t/pullrequestreviewthread-no-longer-showing-up-in-timeline-for-prs/2755/17

Hoping we can get **some** clarification on what is going on here. Its _impossible_ to get PR review threaded comments atm w/ the GQL API. Huge disappointment.

Sherlouk:
Okay so I don't think they're going to fix it... so I've found an alternative. @rnystrom 

https://api.github.com/repos/rnystrom/GitHawk/pulls/704/comments?access_token=revoked
https://github.com/rnystrom/GitHawk/pull/704

The above endpoint returns an array of comments, in chronological order. It appears to only include comments from a review (it doesn't include the root comment, or other comments in the feed from my testing). The nodes can contain a "in_reply_to_id" which is always the first comment in a "discussion".

So in the above PR I left a comment "Not sure if whatever rule is causing this..." the ID of which is `146309348`. The other three comments in that "discussion" have the "in_reply_to_id" field pointing to the same ID. We obviously then use the time to work out the order of messages!

[...]

PR #828, opened 2017-11-02
Open PR comments in another VC

rnystrom:
Using @Sherlouk's idea from #337. Some drawbacks:

- Reactions are totally broken
  - No list of users so detail view isn't visible
  - Don't have GQL id so mutations fail
  - No indication if the viewer has reacted
- Determining author is kind of gross, have to compare usernames
- No editing information
- Can't know if viewer can update

Things unfinished:

- Putting the reviews + comments in proper chronological order w/ the rest of the timeline
- Head/neck/tail position of comments (for borders)

Overall, the code is pretty gross but it _does work_!

simulator screen shot - iphone x - 2017-11-01 at 21 55 11


rnystrom:
Maybe we just strip reactions for the time being?

Sherlouk:
A quick idea à la Slack - what if you could tap on a comment to see the full thread with edit/reactions etc? (Use the comment ID to load full GQL response?)

rnystrom:
I like the tapping on comments idea, but still leaves us stuck w/ reactions being broken. I actually don't mind the inline comments, tho it would be nice to collapse ones that are outdated. Maybe we should explore that... 

For each PR review we could add a button that says "view comments" or something. We can query the total number of **root** comments on each review, so the UI could say "View comments (2)", or we just drop comments entirely and put a "View comments" button, omitting if 0.

@Sherlouk @BasThomas @heshamsalman thoughts?

rnystrom:
Ok I'm actually really digging this idea. It cleans up `IssueViewController` _a lot_ by removing the request and having to rebuild the timeline....

I think I'm going to build this today.

BasThomas:
Seems 👍

Sherlouk:
I'm glad it took you an entire minute to decide that was the right idea 😂 Sounds good!

Sherlouk:
It's worth remembering that GQL allows you to query the total count, even if you don't load them at no extra cost to rate limits. So we can give accurate information without hurting them, and they just tap what they actually wanna see ("buy")

rnystrom:
Ya the only thing misleading with showing the count is that it’s only a count of **root** comments. If there’s a PR with one thread of 100 comments, the total count still just says “1”.

rnystrom:
Gonna land so we can try this out in TF

Issue #441, opened 2017-09-27
Handle GH links natively


Sherlouk:
Extension to #146 

When a user inputs a full GitHub URL, we should modify the text to it's shorthand counter-part (where appropriate) and also handle it natively rather than opening in a Safari VC

Examples:

https://github.com/rnystrom/GitHawk/issues/441
`https://github.com/rnystrom/GitHawk/issues/441` (Should go to the issue)

https://github.com/rnystromtest/Githawk-Playground/issues/2
`https://github.com/rnystromtest/Githawk-Playground/issues/2` (Should go to the issue)

https://github.com/rnystrom/GitHawk
`https://github.com/rnystrom/GitHawk` (No shorthand; Should navigate to the repository view)

BasThomas:
The first one does not work (anymore) as the repository has been moved to the organization. GitHub itself reroutes it. Can we do that too?

rnystrom:
Also these links aren’t shortened anymore 🤔 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BasThomas:
Oh I think that might be because I put them in a list

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BasThomas:
https://github.com/rnystrom/GitHawk/issues/441
`https://github.com/rnystrom/GitHawk/issues/441` (Should go to the issue)

https://github.com/rnystromtest/Githawk-Playground/issues/2
`https://github.com/rnystromtest/Githawk-Playground/issues/2` (Should go to the issue)

https://github.com/rnystrom/GitHawk
`https://github.com/rnystrom/GitHawk` (No shorthand; Should navigate to the repository view)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BrianLitwin:
Looking [here](https://github.com/GitHawkApp/GitHawk/blob/master/Classes/Issues/Comments/Markdown/CMarkParsing.swift#L45), the shorlinkInfo method works fine; the context.inLink flag is false tho on all of them. 

They also all have leading spaces: `" https://github.com/rnystromtest/Githawk-Playground/issues/2" ` when you print them?

PR #849, opened 2017-11-04
Use repository default branch rather than master for file view

### LINKED ISSUE #824: Repository 'code' view shows master instead of default
Code view shows the state of `master` instead of the repository's default branch ie `dev` - not sure if intentional

Sherlouk:
Closes #824 

Stores the defaultBranch off a repository and uses that for the code view. We could fetch it once the user goes to the page, but that would then require two separate requests which I felt was a tad overkill for this.

I made the conscious choice to pass the branch through to the code view rather than assuming it as the default branch from the repository details model in order to allow future expansion and link to a certain branch (#441)

Precursor to adding functionality of changing branches on the fly, even if we did support that - this functionality should be there as per GitHub desktop.

### DIFF
--- a/Freetime.xcodeproj/project.pbxproj
+++ b/Freetime.xcodeproj/project.pbxproj
@@ -2025,7 +2025,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "if which swiftlint >/dev/null; then\n  swiftlint\nelse\n  echo \"warning: SwiftLint not installed, download from https://github.com/realm/SwiftLint\"\nfi";
+			shellScript = "\"${PODS_ROOT}/SwiftLint/swiftlint\"";
 		};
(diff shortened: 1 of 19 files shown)


Freetime.xcodeproj/project.pbxproj
```diff
@@ -2025,7 +2025,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "if which swiftlint >/dev/null; then\n  swiftlint\nelse\n  echo \"warning: SwiftLint not installed, download from https://github.com/realm/SwiftLint\"\nfi";
+			shellScript = "\"${PODS_ROOT}/SwiftLint/swiftlint\"";
```
Sherlouk: Oh and this had been mildly infuriating me for ages 😂 It keeps trying to use my local version of SwiftLint so I get about 300 violations and can't build anything! This uses the actual version from Pods

Issue #863, opened 2017-11-05
Removing last item from bookmarks doesn't return empty view


BasThomas:
Search works as expected, though.

rnystrom:
Gonna lean on #854 fixing this as well

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Issue #862, opened 2017-11-05
Bookmark Search bar can be dragged over the empty state


BasThomas:
v 1.12.0 (1149) on TF. Not sure if this is caused by the search bar or the empty state
![simulator screen shot - iphone 5s - 2017-11-05 at 15 22 06](https://user-images.githubusercontent.com/4190298/32415641-19a6c848-c23d-11e7-8dba-445a952ed137.png)
.

Sherlouk:
because it's just a background view on the table view as opposed to it's own cell which we use for search/notifications/repos/etc 

#854 would "fix" this

Issue #855, opened 2017-11-05
Unify store behavior


heshamsalman:
SearchRecentStore & BookmarksStore have different impl., behavior.

rizwankce:
Things to note here bookmark store needs to be name spaced per user account. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rnystrom:
No reason the search one can’t either. Ya we should unify.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Issue #854, opened 2017-11-05
IGListKit on Bookmarks


rizwankce:
Follow up from #819 

While starting we thought that bookmarks are going to be very simple table view cell with table view editing mode. Along the road, we had to do a lot more than that and added self-sizing, swipe to delete and support from iOS 10. 

Now if we step back and see, there are lots of ways we can improve or reuse using collection view cells in Bookmarks. So it would be really better to re-write bookmarks view controller to adopt IGListKit.

rizwankce:
@heshamsalman mind pushing your iglsk re-write for this and we can pull this one together?
Or Do you want me to take care?

heshamsalman:
@rizwankce yeah, i can push it up soon. should be ready for review ~Monday

PR #865, opened 2017-11-05
Chore/bookmark rewrite

heshamsalman:
TODO: 

- [x] Clear All
- [x] Remove old code
- [x] Remove Swipe hacks 
- [x] Search 

This is likely going to be a larger PR than I normally like, so I'm going to try and have a clear commit history here.

Ref #855 #854 #862 #863


rnystrom:
This is awesome. Idk how deep I'll go w/ a review on this one. Maybe just yolo land and test out in TF. I'd love to ship 1.12 soon and this is the last big blocker IMO.

heshamsalman:
@rnystrom Should be good. But running into one small issue that i wonder if you have insight on? Can't tell if it's true of on-device builds but it looks like the auto-sized bookmark cells (unchanged; just a collection cell now) have the occasional stutter on-update.

rnystrom:
Self-sizing cells + IGListKit notoriously don't play nicely. We should probably change them to use `NSAttributedStringSizing`. Will be much better perf and no weird self-sizing flickers.

heshamsalman:
@rnystrom good to go :+1:

Classes/Bookmark v2/BookmarkStore.swift
```diff
@@ -0,0 +1,42 @@
...
+    }
+    let defaults = UserDefaults.standard
+
+    let encoder = JSONEncoder()
+    let decoder = JSONDecoder()
+
+    let token: String
+
+    init(token: String) {
+        self.token = token
```
rizwankce: We need to use this user token for namespacing. So that each user account can have different bookmarks. Maybe we should move this to `Store`

heshamsalman:
Good catch re: name spacing! Can add that one tmrw in a separate PR. #783 should be handled already by the update on viewWillAppear and the way `bookmarks` is being stored in the `BookmarkStore`

rizwankce:
@heshamsalman I just tested on iPad SIM. #783 actually meant for iPad. When we toggle bookmark in detail view, The master view is not updated. Check #773 for more info.

heshamsalman:
@rizwankce should be updating listeners now, _a la_ #783

rnystrom:
🌮

Classes/Bookmark v2/Bookmark.swift
```diff
@@ -0,0 +1,62 @@
...
+        ) {
+        self.type = type
+        self.name = name
+        self.owner = owner
+        self.number = number
+        self.title = title
+        self.hasIssueEnabled = hasIssueEnabled
+        self.defaultBranch = defaultBranch
+    }
+}
```
rnystrom: We probably need to add `hashValue` eventually

Classes/Bookmark v2/Bookmark.swift
```diff
@@ -0,0 +1,62 @@
...
+
+extension Bookmark: Filterable {
+    func match(query: String) -> Bool {
+        let lowerQuery = query.lowercased()
+
+        if type.rawValue.contains(lowerQuery) { return true }
+        if String(number).contains(lowerQuery) { return true }
+        if title.lowercased().contains(lowerQuery) { return true }
+        if owner.lowercased().contains(lowerQuery) { return true }
+        if name.lowercased().contains(lowerQuery) { return true }
```
rnystrom: I wonder which technique is faster. This or simply:

```swift
return (type + String(number) + title + owner + name).lowercased().contains(lowerQuery)
```

Classes/Bookmark v2/BookmarkSectionController.swift
```diff
@@ -0,0 +1,69 @@
...
+
+    init(delegate: BookmarkSectionControllerDelegate) {
+        self.delegate = delegate
+        super.init()
+    }
+
+    override func sizeForItem(at index: Int) -> CGSize {
+        guard let width = collectionContext?.containerSize.width, let object = object else { fatalError("Missing context") }
+
+        return CGSize(width: width, height: max(object.text.textViewSize(width).height, Styles.Sizes.tableCellHeightLarge))
```
rnystrom: needs more newlines!

Classes/Bookmark v2/BookmarkStore.swift
```diff
@@ -0,0 +1,63 @@
...
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+protocol BookmarkListener: class {
+    func didUpdateBookmarks()
+}
+
+class BookmarkStore: Store {
```
rnystrom: nit: `final`

Classes/Bookmark v2/BookmarkStore.swift
```diff
@@ -0,0 +1,63 @@
...
+    func add(listener: BookmarkListener) {
+        let wrapper = ListenerWrapper(listener: listener)
+        listeners.append(wrapper)
+    }
+
+    func save() {
+        guard let data = try? encoder.encode(values) else { return }
+        defaults.set(data, forKey: key)
+        listeners.forEach { $0.listener?.didUpdateBookmarks() }
+    }
```
rnystrom: - What if stores have listeners as part of the design?
- Maybe add a completion block in the `Store` extension so that you don't have to basically copy/paste here?

Classes/Bookmark v2/BookmarkViewController.swift
```diff
@@ -0,0 +1,304 @@
...
+        case filtering(String)
+    }
+
+    var state: State = .idle
+
+    // MARK: - Initialization
+
+    init(client: GithubClient) {
+        self.client = client
+        guard let store = client.bookmarksStore else { fatalError("Client does not have a bookmark store") }
```
rnystrom: What happens on fresh app install? Do we not init the VC? Maybe the init should be failable so that is represented?

Classes/Bookmark v2/BookmarkViewController.swift
```diff
@@ -0,0 +1,304 @@
...
+            name: .UIKeyboardWillHide,
+            object: nil
+        )
+
+        nc.addObserver(
+            searchBar,
+            selector: #selector(UISearchBar.resignFirstResponder),
+            name: .UIKeyboardWillHide,
+            object: nil)
+    }
```
rnystrom: note to self: need a KB helper to handle this common stuff

Classes/Bookmark v2/BookmarkViewController.swift
```diff
@@ -0,0 +1,304 @@
...
+        let view = InitialEmptyView(
+            imageName: "bookmarks-large",
+            title: NSLocalizedString("Add Bookmarks", comment: ""),
+            description: NSLocalizedString("Bookmark your favorite issues,\npull requests, and repositories.", comment: "")
+        )
+        view.delegate = self
+        return view
+    }
+
+    // MARK: - UISearchBarDelegate
```
rnystrom: nit: `// MARK: ` style inconsistent w/ rest of app. No "-"

Classes/Bookmark v2/BookmarkViewModel.swift
```diff
@@ -0,0 +1,60 @@
...
+//  Freetime
+//
+//  Created by Hesham Salman on 11/5/17.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+import IGListKit
+
+class BookmarkViewModel: NSObject, ListDiffable {
```
rnystrom: nit: `final`

also you don't need to subclass `NSObject`

Classes/Bookmark v2/BookmarkViewModel.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    let bookmark: Bookmark
+    var text: NSAttributedStringSizing
+
+    init(bookmark: Bookmark, width: CGFloat) {
+        self.bookmark = bookmark
+        let repositoryText = NSMutableAttributedString(attributedString: RepositoryAttributedString(owner: bookmark.owner, name: bookmark.name))
```
rnystrom: newlines 🙏

Classes/Bookmark v2/BookmarkViewModel.swift
```diff
@@ -0,0 +1,60 @@
...
+        text = NSAttributedStringSizing(
+            containerWidth: width,
+            attributedText: repositoryText,
+            inset: BookmarkCell.titleInset
+        )
+    }
+
+    var icon: UIImage {
+        return bookmark.type.icon.withRenderingMode(.alwaysTemplate)
+    }
```
rnystrom: Do we access images from models anywhere else in the app? Looks like this is only used in the cell's configure method. Probably belongs in there.

Classes/Bookmark v2/BookmarkViewModel.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    // MARK: - ListDiffable
+
+    func diffIdentifier() -> NSObjectProtocol {
+        return "#\(bookmark.number) - \(bookmark.name) - \(bookmark.owner) - \(bookmark.title)" as NSObjectProtocol
```
rnystrom: for perf, compute + store this in init

Classes/Bookmark v2/BookmarkViewModel.swift
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    func isEqual(toDiffableObject object: ListDiffable?) -> Bool {
+        return true
```
rnystrom: I can see this breaking once we start refreshing bookmarks. IMO better to do a full equality check

Classes/Utility/Store.swift
```diff
@@ -0,0 +1,57 @@
...
+protocol Store: AnyObject {
+    associatedtype Model: Codable, Equatable
+
+    var key: String { get }
+    var defaults: UserDefaults { get }
+
+    var encoder: JSONEncoder { get }
+    var decoder: JSONDecoder { get }
+
+    var values: [Model] { get set }
```
rnystrom: Really weary about a public `var` here.

1. No thread safety
2. Anyone can do anything

Also we probably need to revisit this so that mutations and lookups are O(1)

Classes/Utility/Store.swift
```diff
@@ -0,0 +1,57 @@
...
+    func remove(_ value: Model)
+    func clear()
+
+    func save()
+    func contains(_ value: Model) -> Bool
+}
+
+extension Store {
+    func add(_ value: Model) {
+        guard !values.contains(value) else { return }
```
rnystrom: Hmph, O(_n_) lookup for every change?

Classes/Utility/Store.swift
```diff
@@ -0,0 +1,57 @@
...
+        values.insert(value, at: 0)
+        save()
+    }
+
+    func remove(_ value: Model) {
+        guard let offset = values.index(of: value) else { return }
```
rnystrom: Another O(_n_) 😕

Classes/Utility/Store.swift
```diff
@@ -0,0 +1,57 @@
...
+        save()
+    }
+
+    func save() {
+        guard let data = try? encoder.encode(values) else { return }
+        defaults.set(data, forKey: key)
+    }
+
+    func contains(_ value: Model) -> Bool {
+        return values.contains(value)
```
rnystrom: x3

Classes/Bookmark v2/Bookmark.swift
Sherlouk: Interpolation rather than concatenation of the strings, but interesting idea

Classes/Bookmark v2/BookmarkViewController.swift
Sherlouk: There's a SwiftLint rule to ensure marks all follow the same style. I think it chooses to include the `-` but it would enforce everyone to use same? Just an idea!

Classes/Bookmark v2/Bookmark.swift
rnystrom: @Sherlouk sorry what does this mean?

Sherlouk: Concatenating a bunch of strings like that has rather poor performance in Swift, if you interpolate the same thing like so it should be faster:

`"\(type)\(String(number))\(title)\(owner)\(name)".lowercased()...`

heshamsalman:
I'm down to split these up. I can start cracking on them tmrw.

Classes/Bookmark v2/BookmarkViewModel.swift
heshamsalman: Beyond limiting this class from growing out of control through extensions and subclasses, is there a perf. benefit to `final`?

PR #878, opened 2017-11-07
Stores now have listeners as a part of their design (ref #865 review)

heshamsalman:
Listeners are a built in part of the design of `Store` protocol.
Removed special-case code from the `BookmarkStore`.


rnystrom:
Yeeessssssss

Classes/Utility/Store.swift
```diff
@@ -8,6 +8,18 @@
 
 import Foundation
 
+protocol StoreListener: class {
```
rnystrom: ❤️

PR #879, opened 2017-11-07
Changes Bookmark action to be a nav bar item

### LINKED ISSUE #872: Change Bookmark action to be a nav bar item
- Free up alert action space
- Simpler, more intuitive UI

I'll prepare some assets and PR them if someone wants to hack at this. I think this is the last big thing I want to do before launching 1.12.0!

@rizwankce @heshamsalman @Sherlouk @BasThomas for thoughts

rizwankce:
Closes #872


Classes/Repository/RepositoryViewController.swift
```diff
@@ -81,6 +85,50 @@ NewIssueTableViewControllerDelegate {
...
+        }
+
+        if !store.contains(bookmark) {
+            store.add(bookmark)
+            sender.image = UIImage(named: "nav-bookmark-selected")
+        } else {
+            store.remove(bookmark)
+            sender.image = UIImage(named: "nav-bookmark")
+        }
+        Haptic.triggerNotification(.success)
```
rnystrom: A lot of this toggling code looks identical to the stuff in Issues. Think you could make a new file with a shared function both call sites can use? Similar to what I did when this was an alert.

rizwankce: Fair point 👍  Let me try!

rizwankce:
@rnystrom Made some changes for shared toggle bookmark.

heshamsalman:
The new bookmark nav bar item is _so sick_. 

PR #881, opened 2017-11-07
Hides "Mark All" button when there is no "Unread" message

### LINKED ISSUE #821: Inbox - mark all button visible all the time
When we have inbox zero, expecting the "Mark Al"l button to be disabled. But in current build its always visible. 

On click "Mark All", triggering request but it's not disabling itself.

<details>
<pre>
Version 1.12.0 (1137)
Device: iPhone 7 Plus (iOS 11.1)
TestFlight: true
</pre>
</details>

rizwankce:
Closes #821

### DIFF
--- a/Classes/Notifications/NotificationsViewController.swift
+++ b/Classes/Notifications/NotificationsViewController.swift
@@ -78,6 +78,7 @@ FlatCacheListener {
                 action: #selector(NotificationsViewController.onMarkAll(sender:))
             )
         }
+        updateUnreadState(count: notificationIDs.count)
     }
 
     private func updateUnreadState(count: Int) {


rizwankce:
A quick test seems to be working but not sure using `notificationIDs.count` to update the status is correct. Need some help in reviewing @rnystrom

rnystrom:
This is better than where I [added it before](https://github.com/rnystrom/GitHawk/blob/master/Classes/Notifications/NotificationsViewController.swift#L197), that's way too side-effect-ish. Maybe we can remove that entirely?

rizwankce:
Yup. we can totally remove that. Literally, I placed breakpoints all over and figured out this fix. 😆

BasThomas:
Thanks for this @rizwankce! Super annoying bug 😛

PR #884, opened 2017-11-07
More shared logic for bookmarks in nav item

rnystrom:
Followup from #879 for _even more_ shared code. Trying to avoid letting implementations diverge as much as possible.


Classes/Bookmark v2/BookmarkNavigationController.swift
```diff
@@ -0,0 +1,68 @@
...
+        return item
+    }
+
+    // MARK: Private API
+
+    @objc
+    func add(sender: UIBarButtonItem) {
+        Haptic.triggerSelection()
+        sender.action = #selector(BookmarkNavigationController.remove(sender:))
+        sender.image = UIImage(named: "nav-bookmark-selected")
```
rnystrom: btw @rizwankce, switching the image on the _sender_ is really clever 👌 I like it!

PR #888, opened 2017-11-07
@objc at same line as func

### LINKED ISSUE #882: Put all @objc on same line as func
Idk why I went w/ this style originally, but it looks clowny. Should be an easy project-wide find/replace.

rizwankce:
Closes #882


rizwankce:
@rnystrom a conflict-free PR for you!

PR #897, opened 2017-11-08
Add keyboard helpers to remove duped code

rnystrom:
Noticed some copy/paste. This should speed up anyone working on stuff that requires handling the kb. Also adds animation.


heshamsalman:
Any reason to not remove observers on deinit of the helper?

PR #905, opened 2017-11-08
Open GitHub Status when tapping cell

### LINKED ISSUE #904: API Status opens GitHub
Open in Safari: https://status.github.com/messages

BasThomas:
Fixes #904 

Any reason why that cell is not a `StyledTableCell`?


BasThomas:
Good I checked - it's not actually the StyledTableCell, but the selection was `none` instead of `default`.

When selected, the status indicator actually is not shown (like with the labels), can we live with that for now?

rnystrom:
I’m fine with that

PR #925, opened 2017-11-12
Remove Search subtitle from shortcut

rnystrom:
Making these consistent


BasThomas:
I would probably prefer having subtitles on all of them instead of removing. What do you think?

Issue #369, opened 2017-09-22
Add milestones to issue/pr 


rnystrom:
Version 1.10.0 (1897)
Device: iPhone 7 Plus (iOS 11.0)

PR #936, opened 2017-11-14
Move label editing to a managing control with collaborator request

### LINKED ISSUE #929: New editing controls
I think I solved our collaborator issues. We will def need to make another request to get collab information, but that’s only complicated now because these controls are placed with each unit (eg labels).

Instead, I propose we move all admin/collab controls to one place. It can start off as a single button that says “Edit” and a down carrot. Tap the cell and it unfolds to show all available editing controls.

- Edit Labels
- Edit Assignees
- etc

This gives us a place to keep expanding admin controls. Maybe eventually we move open/close here too?

Putting this in a. Single unit means we only need to update with a single new model once the collab endpoint returns, making this setup very simple.

I’ll post back with some drawings.

<details>
<pre>
Version 1.12.0 (1179)
Device: iPhone 7 Plus (iOS 11.1.1)
TestFlight: false
</pre>
</details>

rnystrom:
Fixes #929 and unblocks #369

I think this needs some design love (now or later). 

One idea is to maybe move the “Manage” to the top right of the screen, in-line with the open button. We don’t even need to merge cells or section controllers for this. I’ll give it a try.


rnystrom:
If someone wouldn't mind giving this a quick sanity check, I'll land and bump the version so we can get a new TestFlight out!

Sherlouk:
So yeah does seem to work, but I'm not a huge fan of the location.

Maybe we could move Close/Lock into that manage area too? Removes a bit of clutter from the overflow, and turns into a nice colloborator place!

Classes/Issues/GithubClient+Issues.swift
```diff
@@ -269,4 +269,52 @@ extension GithubClient {
...
+        completion: @escaping (Result<Bool>) -> Void
+        ) {
+        guard let viewer = userSession?.username else {
+            completion(.error(nil))
+            return
+        }
+
+        // https://developer.github.com/v3/repos/collaborators/#check-if-a-user-is-a-collaborator
+        request(Request(
+            path: "repos/\(owner)/\(repo)/collaborators/\(viewer)",
```
Sherlouk: We do everything in our power to prevent caching these requests, but should we add something so we don't constantly hit this endpoint? Thinking of our rate limits!

Classes/Issues/IssuesViewController.swift
```diff
@@ -321,6 +325,22 @@ FlatCacheListener {
...
+            client.fetchViewerCollaborator(
+                owner: model.owner,
+                repo: model.repo
+            ) { [weak self] (result) in
+                switch result {
+                case .success(let isCollab):
+                    self?.viewerIsCollaborator = isCollab
+                    // avoid finishLoading() so empty view doesn't appear
+                    self?.feed.adapter.performUpdates(animated: true)
+                case .error:
```
Sherlouk: Error states are never sent through, consider revision?

Classes/Issues/Managing/IssueManagingExpansionCell.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    // MARK: Public API
+
+    func animate(expanded: Bool) {
+        UIView.animate(
+            withDuration: (0.8),
```
Sherlouk: nit: Unnecessary brackets

BasThomas:
Yeah, feels clunky with just the edit labels there.

Regarding design: I feel like the arrow should not move in circles, but rather go back and forth (open clockwise, close anti-clockwise).

Also the funny animation (where the arrow goes a bit too far, then comes back) feels like it's leaving the button slightly slanted, though it isn't. Feels weird atm; maybe tweaking it will solve it.

rnystrom:
@Sherlouk @BasThomas re: space, keep in mind this is where "Edit Reviewers", "Edit Milestone", "Edit Assignees" will go too. Will def fill up!

Just had a stroke of genius (lol) w/ the design. I could make a new cell w/ icons and display them in a grid [sort of like this](https://dribbble.com/shots/3081604-Smart-Home-Login-and-Home-Profile/attachments/649958) (right image).

Sketched out an idea:

img_0008

Classes/Issues/Managing/IssueManagingExpansionCell.swift
rnystrom: copy paste fail, good eye

Classes/Issues/GithubClient+Issues.swift
rnystrom: I was thinking about caching this too. I think you're right, this is worth following up w/ maybe a singleton that caches your collab state in memory for the lifetime of the process. If you were removed as a collab we just return an error result. How's that sound?

Sherlouk: Sounds good, if we get an error result later in the flow then we just clear the cache for that repo so it updates?

Sherlouk:
Loving where the sketch is heading! Would be great to see that!

rnystrom:
Gonna hack at this tonight! I'm pretty excited about where this is going. Unlocks a lot of other stuff we can put in there and reduce clutter. Only concern atm is 320-wide devices on issues that are closed+locked (uses 2 pills in the top left). Also if we add German localization the labels will be like 1,000 pixels wide.

BasThomas:
Yep, that way of working on this makes sense. Re pills: maybe use a lock symbol when locked instead of text?

Sherlouk:
I feel like it should all fit on one line, otherwise just shift underneath either right or left aligned still?
Worst case we just always have the options visible?

rnystrom:
@BasThomas good idea! We have icons for locked and unlocked, that would be perfect.

Also thoughts on the action word "Manage" here? Maybe it should say "Collab"? That isn't as localizable though.

BasThomas:
I think “Manage” is much better than “Collab”, especially since collab would imho limit what we should put there (eg locking etc is not a collab task imho). “(Issue) Settings” could be an alternative maybe?

BasThomas:
And at that point we might be able to find a better ... in the top right? Only has share and view at that point...

rnystrom:
Here's the latest:

manage

Trying to figure out a clean way to "reverse" the rotation animation on the chevron. Any tips?

Sherlouk:
Looks sweet!

rnystrom:
Gonna pull this in so we can start iterating!

PR #948, opened 2017-11-14
Bump version to 1.12.1

rnystrom:
🚀


Sherlouk:
1.12.1? If it's just the quick patch to images?

rnystrom:
I wasn't going to patch the image button, but probably should now that you mention it. Will update and ship a new version today.

PR #966, opened 2017-11-17
Use collab permissions endpoint to check collab status

### LINKED ISSUE #963: Manage not available when part of organisation
I’m a member of MessageKit org but can’t manage any of the issues/prs even though I should be able to!

<details>
<pre>
Version 1.13.0 (1190)
Device: iPhone 6s Plus (iOS 11.2)
TestFlight: true
</pre>
</details>

rnystrom:
Fixes #963

Sending out a TF build from this branch

### DIFF
--- a/Classes/Issues/GithubClient+Issues.swift
+++ b/Classes/Issues/GithubClient+Issues.swift
@@ -294,10 +294,13 @@ extension GithubClient {
             path: "repos/\(owner)/\(repo)/collaborators/\(viewer)/permission",
             headers: ["Accept": "application/vnd.github.hellcat-preview+json"],
             completion: { (response, _) in
-                // documentation states that collab = 204
-                if let json = response.value as? [String: Any],
+                let statusCode = response.response?.statusCode
+                if statusCode == 200,
+                    let json = response.value as? [String: Any],
                     let permission = json["permission"] as? String {
-                    completion(.success(permission == "admin" || permission == "write"))
+                    completion(.success((permission == "admin" || permission == "write")))
+                } else if statusCode == 403 {
+                    completion(.success(false))
                 } else {
                     completion(.error(response.error))
                 }
--- a/Classes/Issues/GithubClient+Issues.swift
+++ b/Classes/Issues/GithubClient+Issues.swift
@@ -280,9 +280,9 @@ extension GithubClient {
     }
 
     enum CollaboratorPermission: String {
-        case admin = "admin"
-        case write = "write"
-        case read = "read"
+        case admin
+        case write
+        case read
         case none
 
         static func from(_ str: String) -> CollaboratorPermission {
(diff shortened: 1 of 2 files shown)


Classes/Issues/GithubClient+Issues.swift
```diff
@@ -289,13 +289,21 @@ extension GithubClient {
             return
         }
 
-        // https://developer.github.com/v3/repos/collaborators/#check-if-a-user-is-a-collaborator
+        // https://developer.github.com/v3/repos/collaborators/#review-a-users-permission-level
         request(Request(
-            path: "repos/\(owner)/\(repo)/collaborators/\(viewer)",
+            path: "repos/\(owner)/\(repo)/collaborators/\(viewer)/permission",
             headers: ["Accept": "application/vnd.github.hellcat-preview+json"],
```
Sherlouk: Does this endpoint still need the accepts value?

Sherlouk:
Yaas!

Classes/Issues/GithubClient+Issues.swift
```diff
@@ -289,13 +289,21 @@ extension GithubClient {
...
             completion: { (response, _) in
-                // documentation states that collab = 204
-                completion(.success(response.response?.statusCode == 204))
+                let statusCode = response.response?.statusCode
+                if statusCode == 200,
+                    let json = response.value as? [String: Any],
+                    let permission = json["permission"] as? String {
```
Sherlouk: Super minor but is it worth loading these permissions into an enum and passing that value through to be handled else where?

The enum could then have a `canManageIssue: Bool` which returns true if `self == .admin || self == .write`.

Just thinking ahead we may want to use the user's permission for other things and makes it a bit more flexible without locking this down to strictly this one use case?

Classes/Issues/GithubClient+Issues.swift
rnystrom: Good catch, nope!

Classes/Issues/GithubClient+Issues.swift
rnystrom: I was wondering this too. Doesn't hurt either way. I'll add this!

Classes/Issues/GithubClient+Issues.swift
```diff
@@ -279,23 +279,51 @@ extension GithubClient {
         }))
     }
 
+    enum CollaboratorPermission: String {
+        case admin = "admin"
+        case write = "write"
+        case read = "read"
```
Sherlouk: Values are unnecessary here, defaults to same as name

rnystrom: 😱 TIL!

BasThomas: It's awesome, been around since Swift 2 I think!

PR #974, opened 2017-11-18
Display a "no description given" message if the root comment is empty

### LINKED ISSUE #959: Add “no description given" for empty root comments
Eg. https://github.com/MessageKit/MessageKit/issues/354

<details>
<pre>
Version 1.12.1 (1185)
Device: iPhone X (iOS 11.1.1)
TestFlight: true
</pre>
</details>

BasThomas:
Fixes #959

Didn't feel completely right to put this in the Model, but couldn't find any other place where this would fit better.

### DIFF
--- a/Classes/Issues/Comments/IssueCommentModel.swift
+++ b/Classes/Issues/Comments/IssueCommentModel.swift
@@ -51,7 +51,7 @@ final class IssueCommentModel: ListDiffable {
         self.collapse = collapse
         self.threadState = threadState
         self.number = number
-        self.rawMarkdown = rawMarkdown
+        self.rawMarkdown = (rawMarkdown.isEmpty && isRoot) ? "*\(Constants.Strings.noDescriptionProvided)*" : rawMarkdown
         self.viewerCanUpdate = viewerCanUpdate
         self.viewerCanDelete = viewerCanDelete
         self.isRoot = isRoot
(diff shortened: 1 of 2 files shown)


Sherlouk:
Out of interest is there a reason we can't just do this as an "empty view" (of sorts) to the comment section controller?

We already have a section controller which has a list of all of the individual models it needs (for images, markdown, html, etc) and adds them all as one cell each. Can we not at that point basically say `if models.count == 0 { return NoDescriptionGivenCell() }`

I might be missing something!

rnystrom:
Wait, actually don't we already do basically this here:

https://github.com/rnystrom/GitHawk/blob/f213bac92431677962289ed6925f69797e351cd3/Classes/Issues/Comments/Markdown/CommentModelsFromMarkdown.swift#L42-L57

I like @Sherlouk's idea too, that's maybe a more appropriate representation of empty data. Atm this is sort of side-effecty.

BasThomas:
Oh huh, seems like we do. I think it wasn't showing as the root comment referenced wasn't exactly empty (it was something like `/r/r/n`) and this didn't kick in.

And yeah, I definitely prefer the way we're already doing this. Closing.

rnystrom:
Seems like there's still a bug tho. We should be trimming whitespace here:

https://github.com/rnystrom/GitHawk/blob/f213bac92431677962289ed6925f69797e351cd3/Classes/Issues/Comments/Markdown/CommentModelsFromMarkdown.swift#L64-L68

I think, right?

BasThomas:
Don't know if just removing whitespaces wouldn't break anything else...

PR #984, opened 2017-11-18
New comments use FlatCache mutation system

### LINKED ISSUE #947: Optimistic update puts close event above comment
Here I commented, then closed. Optimistic state is the other way around (it inserted the close event above the comment)

Also: check the top closed label being truncated for whatever reason

GitHawk Upload by BasThomas

<details>
<pre>
Version 1.12.0 (1179)
Device: iPhone X (iOS 11.1.1)
TestFlight: true
</pre>
</details>

rnystrom:
Fixes #947

### DIFF
--- a/Classes/Issues/IssuesViewController.swift
+++ b/Classes/Issues/IssuesViewController.swift
@@ -27,10 +27,10 @@ FlatCacheListener {
     private let model: IssueDetailsModel
     private let addCommentClient: AddCommentClient
     private let autocomplete = IssueCommentAutocomplete(autocompletes: [EmojiAutocomplete()])
-    private var hasScrolledToBottom = false
     private let viewFilesModel = "view_files" as ListDiffable
     private let textActionsController = TextActionsController()
     private var bookmarkNavController: BookmarkNavigationController? = nil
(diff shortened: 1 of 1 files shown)


Classes/Issues/IssuesViewController.swift
```diff
@@ -529,12 +523,13 @@ FlatCacheListener {
...
-        sentComments.append(comment)
 
-        let collectionView = feed.collectionView
-        feed.adapter.performUpdates(animated: false, completion: { _ in
-            collectionView.slk_scrollToBottom(animated: true)
-        })
+        needsScrollToBottom = true
+
+        let newResult = previous.updated(
+            timelinePages: previous.timelinePages(appending: [comment])
```
BasThomas: This (the `timelinePages(appending:)`) looked really weird and had to look up what it was actually doing. Can this be renamed to more clearly state the side effect?

At least we can add a similar method that accepts just one value instead of having to wrap it in an array.

rnystrom: What would you rename it to? FWIW there is no side effect, it returns new pages appending the array you give it.

PR #1001, opened 2017-11-19
Add Crashlytics and Answers

### LINKED ISSUE #998: Basic Analytics
From experience iTC doesn't really give you the drill-down on analytics that we'd really like

Following the [localization discussion](https://twitter.com/_ryannystrom/status/932301382043996160) on Twitter, it's maybe worth adding a couple of our own analytics calls?

Namely: Device Language & Accessibility Preferences

If we did, should probably be opt-in only

(Obviously this wouldn't help with trying to enter new markets, but understanding the needs of the people who already use it)

rnystrom:
Fixes #998


Sherlouk:
Is it worth doing a custom event on app launch to track things like accessibility and localization? Or card up something else to do that?

rnystrom:
@Sherlouk ya let's followup w/ a new issue for that. Agree tho, we should start capturing app launches w/ VoiceOver or other AX features enabled.

PR #1007, opened 2017-11-19
Fix unit tests

rnystrom:
Tests were failing to build after merging #809


Sherlouk:
@rnystrom We should probably move to using `@testable` to prevent these issues again! 😅

rnystrom:
@Sherlouk what’s that?

Sherlouk:
It imports the main project as a module like you could with any Pod, but the `@testable` attribute allows you to have access to all internal functions

Means you don't need to target files to the tests class!

BasThomas:
Also what about running the unit tests on CI? 😄

rnystrom:
@Sherlouk oh snap that's awesome!

@BasThomas they already do! [example](https://dashboard.buddybuild.com/apps/59d904ff8588d60001729c4e/build/5a12496995807f00017ce81d#tests)

BasThomas:
🤔 why wasn't this caught then?

rnystrom:
I merged the double tap Pr without looking 😅

PR #1011, opened 2017-11-20
Add and remove assignees

### LINKED ISSUE #560: Assign issues and pull requests

rnystrom:
Fixes #560 

img_0075

img_0076

### DIFF
--- a/Classes/Issues/Assignees/IssueAssigneeSummaryModel.swift
+++ b/Classes/Issues/Assignees/IssueAssigneeSummaryModel.swift
@@ -14,11 +14,13 @@ final class IssueAssigneeSummaryModel: ListDiffable {
     let title: String
     let urls: [URL]
     let expanded: Bool
+    private let urlHash: String
 
     init(title: String, urls: [URL], expanded: Bool) {
         self.title = title
         self.urls = urls
         self.expanded = expanded
+        self.urlHash = urls.reduce("") { $0 + $1.path }
     }
 
     // MARK: ListDiffable
@@ -31,6 +33,7 @@ final class IssueAssigneeSummaryModel: ListDiffable {
         if self === object { return true }
         guard let object = object as? IssueAssigneeSummaryModel else { return false }
         return expanded == object.expanded
+        && urlHash == object.urlHash
     }
 
 }
--- a/Classes/Issues/GithubClient+Issues.swift
+++ b/Classes/Issues/GithubClient+Issues.swift
@@ -389,4 +389,87 @@ extension GithubClient {
         })
     }
 
+    func addAssignees(
+        previous: IssueResult,
+        owner: String,
+        repo: String,
+        number: Int,
+        assignees: [IssueAssigneeViewModel]
+        ) {
+        guard let actor = userSession?.username else { return }
+
+        let oldAssigness = Set<String>(previous.assignee.users.map { $0.login })
+        let newAssignees = Set<String>(assignees.map { $0.login })
+
+        var newEvents = [IssueRequestModel]()
+        var added = [String]()
+        var removed = [String]()
+
+        for old in oldAssigness {
+            if !newAssignees.contains(old) {
+                removed.append(old)
+                newEvents.append(IssueRequestModel(
+                    id: UUID().uuidString,
+                    actor: actor,
+                    user: old,
+                    date: Date(),
+                    event: .unassigned,
+                    width: 0 // will be inflated when asked
+                    ))
+            }
+        }
+        for new in newAssignees {
+            if !oldAssigness.contains(new) {
+                added.append(new)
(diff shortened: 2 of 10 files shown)


Classes/Issues/Assignees/IssueAssigneeSummaryModel.swift
```diff
@@ -14,11 +14,13 @@ final class IssueAssigneeSummaryModel: ListDiffable {
     let title: String
     let urls: [URL]
     let expanded: Bool
+    private let urlHash: String
```
rnystrom: Necessary to get the summary nested view to update

Classes/Issues/GithubClient+Issues.swift
```diff
@@ -389,4 +389,87 @@ extension GithubClient {
...
+            request(GithubClient.Request(
+                path: path,
+                method: .post,
+                parameters: ["assignees": added]
+            ) { (response, _) in
+                handler(201, response.response?.statusCode)
+            })
+        }
+
+        // https://developer.github.com/v3/issues/assignees/#remove-assignees-from-an-issue
```
rnystrom: These endpoints are lame. Unlike labelling, adding and removing are in **separate** endpoints. There's definitely the case where one of these fails and the other succeeds, then you reset the data back to the _original_ state (not including success or failure), but I don't really care at this point. Should be rare.

rnystrom:
Also in case you're curious about the iconography between "Assignees" and "Reviewers", I whipped up this icon tonight for reviewers:

reviewer 3x

Classes/Issues/GithubClient+Issues.swift
Sherlouk: That is such a stupid limitation, what...

Classes/People/PeopleViewController.swift
```diff
@@ -0,0 +1,134 @@
...
+
+    func configure(
+        selections: [String],
+        type: PeopleType,
+        client: GithubClient,
+        delegate: PeopleViewControllerDelegate,
+        owner: String,
+        repo: String
+        ) {
+        self.selections = Set<String>(selections)
```
Sherlouk: Not sure where the best place to make the change would be (maybe an option could get passed through to the `configure` method -- but we should always have the signed-in user as the first option to quickly assign yourself!

Classes/People/PeopleViewController.swift
```diff
@@ -0,0 +1,134 @@
...
+    private var users = [User]()
+    private let feedRefresh = FeedRefresh()
+    private var type: PeopleType!
+    private var selections = Set<String>()
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        tableView.refreshControl = feedRefresh.refreshControl
+        feedRefresh.refreshControl.addTarget(self, action: #selector(LabelsViewController.onRefresh), for: .valueChanged)
```
Sherlouk: `LabelsViewController` here. Fun fact you don't need to specify the class `#selector(onRefresh)` is just as valid :)

Classes/People/PeopleViewController.swift
```diff
@@ -0,0 +1,134 @@
...
+        if let cell = cell as? PeopleCell, let avatarURL = URL(string: user.avatar_url) {
+            let login = user.login
+            cell.configure(avatarURL: avatarURL, username: login, showCheckmark: selections.contains(login))
+        }
+        return cell
+    }
+
+    // MARK: UITableViewDelegate
+
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
```
Sherlouk: https://developer.apple.com/documentation/uikit/uitableviewdelegate/1614943-tableview

We could implement `willSelectRowAt` and return `nil` if the user would exceed their selection count (preventing the visual selecting of the cell?

Another alternative would be to do a quick error modal saying "Exceeded Selection Limit"?

rnystrom: I like the `willSelectRowAt` idea, that'll be clean. The error message might be a bit much, hoping people can pick up on the "3/10" label in the top left. Maybe we can do a scale "pop" animation on that label to draw their attention? IMO details we can add later.

Classes/People/PeopleViewController.swift
rnystrom: TIL

Classes/People/PeopleViewController.swift
rnystrom: GREAT idea. I'll look into adding a "Claim" button at the very top

rnystrom:
Btw should have mentioned, I plan on reusing this for adding Reviewers to PRs since its almost exactly the same (notice the `PeopleType` enum).

rnystrom:
Going to add "claim" as a followup blocked by #1026. Will make these much easier to manage.

PR #1014, opened 2017-11-20
Update Tabman to 1.0.7

msaps:
Updated Tabman and Pageboy pods to latest available versions. This also removes the need for the local Tabman fork as the insetting issues from #923 have been fixed in an update.


rnystrom:
Awesome, thanks @msaps! I'll get better at contributing back to Tabman upstream too 😇

Sherlouk:
@rnystrom For what it's worth, I'm only seeing 20 changes of this PR via the app. There's clearly a tonne more I'm missing!

You seeing the same?

rnystrom:
@Sherlouk ugh yup same here. We should be using the paging API on this. However the files VC is made w/ a `UITableViewController`, we should migrate it to use IGListKit and `BaseListViewController` since that has paging built in.

In fact we should probably be using that for everything.......

PR #1034, opened 2017-11-21
Background and border for managing actions

rnystrom:
I'm not sure I'm sold on this. What do you think?

simulator screen shot - iphone 8 - 2017-11-21 at 15 59 03

### DIFF
--- a/Classes/Issues/Managing/IssueManagingActionCell.swift
+++ b/Classes/Issues/Managing/IssueManagingActionCell.swift
@@ -18,6 +18,8 @@ final class IssueManagingActionCell: UICollectionViewCell, ListBindable {
     override init(frame: CGRect) {
         super.init(frame: frame)
 
+        contentView.backgroundColor = .white
+
         let tint = Styles.Colors.Blue.medium.color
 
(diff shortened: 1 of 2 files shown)


Sherlouk:
I'd say it definitely needs some more padding outside of the container for example above the title. But I quite like it! Especially if we get a row 2 with things like lock & close! I think it looks nice!

BasThomas:
I think what’s definitely off atm is the spacing from the bottom to the title. Think it would help if that was the same spacing as the bottom of the title to the comment. 

joerocca:
+1 on the spacing
Maybe remove the borders?
I️ also feel like it is kind of weird to split the issue/PR status from the other details. May look better if the status were to slide down with the other details somehow?

BasThomas:
Also now seeing that manage isn’t vertically centered with the merge status. 🙈

rnystrom:
I'm gonna let this simmer for a bit, really not crazy about it. Agree w/ the point about moving the status, but that makes things unnecessarily complicated.

rnystrom:
Not feeling this

PR #1040, opened 2017-11-22
Add initial accessibility support for text cell

### LINKED ISSUE #1018: Add comment contents as AX content on IssueCommentTextCell
From #934

BasThomas:
Fixes #1018.

This adds some simple support for accessibility in the cell. It doesn't yet check if the cell is expandable and offer the option to expand it.
screen shot 2017-11-22 at 10 25 57
screen shot 2017-11-22 at 10 25 55

### DIFF
--- a/Classes/Issues/Comments/Text/IssueCommentTextCell.swift
+++ b/Classes/Issues/Comments/Text/IssueCommentTextCell.swift
@@ -18,6 +18,7 @@ final class IssueCommentTextCell: DoubleTappableCell, ListBindable, CollapsibleC
 
     override init(frame: CGRect) {
         super.init(frame: frame)
+        isAccessibilityElement = true
 
         backgroundColor = .white
         contentView.clipsToBounds = true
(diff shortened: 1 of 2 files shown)


Classes/Issues/Comments/Text/IssueCommentTextCell.swift
```diff
@@ -36,6 +37,17 @@ final class IssueCommentTextCell: DoubleTappableCell, ListBindable, CollapsibleC
         textView.reposition(width: contentView.bounds.width)
     }
 
+    // MARK: Accessibility
+
+    override var accessibilityLabel: String? {
+        get {
+            return contentView.subviews
```
BasThomas: I am wondering if we should make this some extension on `UICollectionViewCell`? Something like `generatedAccessibilityLabel` or similar?

rnystrom: Yes please!

PR #1041, opened 2017-11-22
Disable the right item editing a comment to an empty one

### LINKED ISSUE #976: Add native error for empty comments (that are not root)
We should either show a local error toast containing the message (like with the empty title when creating an issue) (this might be fragile if GitHub changes this in the future though, allowing empty comments for some reason).

Or we can check if, based on the error code / headers, we can display something more useful to the user (instead of the generic "Something went wrong.")

BasThomas:
Fixes #976.

Followed the same thing we do when [creating a new issue](https://github.com/rnystrom/GitHawk/blob/master/Classes/New%20Issue/NewIssueTableViewController.swift#L202-L204).

### DIFF
--- a/Classes/Issues/EditComment/EditCommentViewController.swift
+++ b/Classes/Issues/EditComment/EditCommentViewController.swift
@@ -14,7 +14,7 @@ protocol EditCommentViewControllerDelegate: class {
     func didCancel(viewController: EditCommentViewController)
 }
 
-class EditCommentViewController: UIViewController {
+class EditCommentViewController: UIViewController, UITextViewDelegate {
 
     weak var delegate: EditCommentViewControllerDelegate?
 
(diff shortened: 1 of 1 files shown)


Classes/Issues/EditComment/EditCommentViewController.swift
```diff
@@ -164,4 +165,9 @@ class EditCommentViewController: UIViewController {
         ToastManager.showGenericError()
     }
 
+    func textViewDidChange(_ textView: UITextView) {
+        guard textView === self.textView else { return }
```
BasThomas: Just as a sanity check in case we add another textview, but can remove if redundant.

rnystrom: Ya bit redundant atm

PR #1045, opened 2017-11-22
Add magic tap to saving comment/sending issue

### LINKED ISSUE #1042: Add magic tap to save / submit comment or issue
From #934

BasThomas:
Fixes #1042


Sherlouk:
~~Not even sure if it's possible, probably thinking about this *wayy* too much, but could we get it to read out the comment first as like a confirmation? "Sending: [what ever they put], are you sure?" magic tap again to send 🤔~~

Looks good to me!

BasThomas:
Probably could, but I think that would only complicate things.

PR #1052, opened 2017-11-22
Generate accessibility labels for cells

### LINKED ISSUE #1051: Create a type / extension that generates an ax label for a cell
See https://github.com/rnystrom/GitHawk/pull/1040#discussion_r152507935

BasThomas:
Fixes #1051.

This adds an `AccessibilityHelper` that can be used to generate an accessibility label for either a `UICollectionViewCell` or `UITableViewCell`.

### DIFF
--- a/Classes/Notifications/NotificationCell.swift
+++ b/Classes/Notifications/NotificationCell.swift
@@ -106,9 +106,8 @@ final class NotificationCell: SwipeSelectableCell {
         textLabel.attributedText = viewModel.title.attributedText
         dateLabel.setText(date: viewModel.date)
         reasonImageView.image = viewModel.type.icon.withRenderingMode(.alwaysTemplate)
-        accessibilityLabel = contentView.subviews
-            .flatMap { $0.accessibilityLabel }
-            .reduce("", { "\($0).\n\($1)" })
+        accessibilityLabel = AccessibilityHelper
+            .generatedLabel(forCell: self)
             .appending(".\n\(viewModel.type.localizedString)")
     }
 
--- a/Classes/Utility/Accessibility.swift
+++ b/Classes/Utility/Accessibility.swift
@@ -1,0 +1,35 @@
+//
+//  Accessibility.swift
+//  Freetime
+//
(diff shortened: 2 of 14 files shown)


Classes/Notifications/NotificationCell.swift
```diff
@@ -106,9 +106,8 @@ final class NotificationCell: SwipeSelectableCell {
         textLabel.attributedText = viewModel.title.attributedText
         dateLabel.setText(date: viewModel.date)
         reasonImageView.image = viewModel.type.icon.withRenderingMode(.alwaysTemplate)
-        accessibilityLabel = contentView.subviews
-            .flatMap { $0.accessibilityLabel }
-            .reduce("", { "\($0).\n\($1)" })
+        accessibilityLabel = AccessibilityHelper
+            .generatedLabel(forCell: self)
             .appending(".\n\(viewModel.type.localizedString)")
```
BasThomas: As a label is just a string, we can still append anything as we did before.

Classes/Utility/Accessibility.swift
```diff
@@ -0,0 +1,35 @@
...
+private protocol ContentViewContaining {
+    var contentView: UIView { get }
+}
+
+extension UICollectionViewCell: ContentViewContaining {}
+extension UITableViewCell: ContentViewContaining {}
+
+enum AccessibilityHelper {
+
+    static func generatedLabel(forCell cell: UICollectionViewCell) -> String {
```
Sherlouk: What's the thought behind having separate methods for the cells when we could just call the `forContentViewContainer` method directly without changing parameters?

BasThomas: Hmm, because I thought it could be confusing to call this and see it would expect something other than a cell; in this case the protocol. Right now it would be clear to the caller what type they can provide.

BasThomas:
@Sherlouk what do you think the test should cover?

Sherlouk:
That given an array of views it can correctly concatenate them in the right order, if you're really looking for a fun time this Wednesday afternoon you could create one of the cells with a repo model (for example) and check it transfers all the right information

Issue #1061, opened 2017-11-22
Can't build master at the moment


Sherlouk:
![image](https://user-images.githubusercontent.com/15193942/33149424-9eba7182-cfc7-11e7-8348-b29289f5389c.png)

@BasThomas @rnystrom 

😢

BasThomas:
Up to date with master? This is what was happening before #1039

BasThomas:
Or at least, this was happening before I recloned the repo I think?

Sherlouk:
Did a fresh fetch and that's what's happening

Generate Secrets config is before the CP stuff, so that fix has gone in

🤔

BasThomas:
But the xcconfig isn't generated yet looking at your screenshot?

BasThomas:
Did you build or run?

Sherlouk:
Run, I just did a clean and build and it went through, So maybe false alarm

Though maybe it is an issue where you have to build twice because of the order of events

BasThomas:
I *think* you have to specifically build the first time, as for some reason run doesn't invoke the script. Not sure.

Sherlouk:
Buddybuild still broke as well, so something's definitely up!

BasThomas:
Ugh, you're right

rnystrom:
God this is so frustrating. Tempted to just revert all of the changes so I can actually focus on building things instead of fighting blackbox tools.

So its working locally, even cloned a new repo and got it to work out of box. I assume that buddybuild is using their own config or something.

I'm pretty stuck at the moment. Any ideas?

Sherlouk:
Genuinely I'm just unsure how the whole ObjC bit fits in, I feel we could use `ProcessInfo.processInfo.environment` to get the environmental variables. They can be set via the scheme that way, and BB uses env variables AFAIK so should be fine!

Removes all this complexity with .xcconfig etc, but yea I'm easy

rnystrom:
@Sherlouk mind putting up a PR to change that?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rnystrom:
@Sherlouk just checking, are you able to help with this? Would be really appreciated. I tried using the env vars and got stuck.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Sherlouk:
Sorry yea will pick this up today! (DIdn't actually get a notification for the first comment 🤔)

Sherlouk:
[...]

PR #1062, opened 2017-11-22
Add better error messages for Image Upload fail states

### LINKED ISSUE #1047: Image Upload failure always says Rate Limit reached
If the imgur request fails because we don't define the client ID or something else, it'll always say "Rate Limit Reached" which obviously isn't strictly true.

We should show a more appropriate error here so we know the difference!

Sherlouk:
Tested on an older version of master before I realised it wasn't the latest code, can't test on latest version of master because #1061 😢 

Fairly simple logic but if we get an error from Imgur we show it straight up with no magic, if it's an error enforced by us we do the rate limit error. If it's an unexpected error we just use a generic one. Atleast we know which path it went down before it failed!

"Invalid client_id" is the current one, seems fair!

fixes #1047

### DIFF
--- a/Classes/Image Upload/ImageUploadTableViewController.swift
+++ b/Classes/Image Upload/ImageUploadTableViewController.swift
@@ -155,10 +155,21 @@ class ImageUploadTableViewController: UITableViewController {
         }
 
         // Check that we have enough "tokens" to actually upload the image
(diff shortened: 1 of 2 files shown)


Sherlouk:
Tested on master, works :)

PR #1063, opened 2017-11-22
Update config script to create keys when missing

rnystrom:
Will wait on buddy


Sherlouk:
Failed 😢 We should also add a unit test which checks that the value (assuming we can get it past the compile stage) is not equal to CHANGEME. The reason I say this is for the TestFlight builds done via BB, we want it to ensure that it's actually defining it and not just leaving them as default!

PR #1064, opened 2017-11-22
Setup xcconfig instructions

rnystrom:


rnystrom:
Holding until we sort out #1061 b/c it might change

PR #1071, opened 2017-11-23
Use AttributedStringCell for IssueLabeledCell

### LINKED ISSUE #1057: Migrate IssueLabeledCell to AttributedStringCell
See #864 for how-to

Requires adding custom drawing for rounded background colors on labels. See:

- https://stackoverflow.com/questions/21857408/how-to-set-nsstrings-background-cornerradius-on-ios7
- https://stackoverflow.com/questions/16362407/nsattributedstring-background-color-and-rounded-corners

Sherlouk:
Closes #1057 

Would've liked to make the glyph taller but looks like we're limited to the line height 😞 I did add a shadow though? 😄

image


Sherlouk:
Gonna go ahead and mark this as WIP until I do more checks of how it handles wrapping onto multiple lines and multiple labels in a single string etc!

Classes/Issues/Labeled/IssueLabeledCell.swift
```diff
@@ -9,95 +9,78 @@
...
+            bottom: centerVertical,
+            right: Styles.Sizes.eventGutter
+        )
+        
+        set(attributedText: NSAttributedStringSizing(
+            containerWidth: 0,
+            attributedText: attributedString,
+            inset: inset,
+            backgroundColor: Styles.Colors.Gray.lighter.color
+        ))
```
rnystrom: Eeep, creating these is going to be very expensive. Check out the [source](https://github.com/rnystrom/GitHawk/blob/master/Classes/Systems/NSAttributedStringSizing.swift) of `NSAttributedStringSizing`, it'll setup all TextKit objects and find the size. Plus this will happen _every time_ `configure(...)` is called 😱 (and the width here is `0`, so the calc'd size wont even be used).

Can we move this into `IssueLabeledModel` so its only created w/ the model? See #1068 as a reference.

Classes/Issues/Labeled/IssueLabeledCell.swift
```diff
@@ -9,95 +9,78 @@
...
+        
+        attributedString.append(NSAttributedString(string: "  \(model.date.agoString)", attributes: dateAttributes))
+        
+        // Set
+        
+        let centerVertical = (Styles.Sizes.labelEventHeight - Styles.Fonts.secondaryBold.lineHeight) / 2
+        let inset = UIEdgeInsets(
+            top: centerVertical,
+            left: Styles.Sizes.eventGutter,
+            bottom: centerVertical,
```
rnystrom: When we move string generation into the model, let's just use insets for the top/bottom so the cell can be sized according to the string size:

```swift
UIEdgeInsets(
  top: Styles.Sizes.inlineSpacing,
  left: Styles.Sizes.eventGutter,
  bottom: Styles.Sizes.inlineSpacing,
  right: Styles.Sizes.eventGutter
)
```

(we can move this inset to a const since it'll be used so much)

Classes/Views/AttributedStringView.swift
```diff
@@ -8,10 +8,11 @@
...
 
-protocol AttributedStringViewDelegate: class {
-    func didTapURL(view: AttributedStringView, url: URL)
-    func didTapUsername(view: AttributedStringView, username: String)
-    func didTapEmail(view: AttributedStringView, email: String)
+@objc protocol AttributedStringViewDelegate: class {
+    @objc optional func didTapURL(view: AttributedStringView, url: URL)
+    @objc optional func didTapUsername(view: AttributedStringView, username: String)
+    @objc optional func didTapEmail(view: AttributedStringView, email: String)
+    @objc optional func didTapLabel(view: AttributedStringView, label: String)
```
rnystrom: eep, why add `optional`? I guess w/ Swift the compiler forces you to put the `?`, but now if a new delegate forgets to implement a method, the compiler wont catch that bug. Can we leave these required?

Classes/Systems/NSAttributedStringSizing.swift
```diff
@@ -183,3 +183,44 @@ final class NSAttributedStringSizing: NSObject, ListDiffable {
     }
 
 }
+
+class LabelLayoutManager: NSLayoutManager {
```
rnystrom: nit: `final`

Classes/Systems/NSAttributedStringSizing.swift
```diff
@@ -183,3 +183,44 @@ final class NSAttributedStringSizing: NSObject, ListDiffable {
...
+        context?.setShouldAntialias(true)
+        context?.setLineWidth(cornerRadius * 2)
+        context?.setLineJoin(.round)
+        
+        // Draw the label background
+        context?.addPath(path.cgPath)
+        context?.drawPath(using: .fillStroke)
+        
+        // Add a small inset box-shadow to the label
+        UIColor.black.withAlphaComponent(0.2).set()
```
rnystrom: Could we do a device-pixel-thick border and use `Styles.Colors.Gray.border` instead so its the same style [as this](https://github.com/rnystrom/GitHawk/blob/master/Classes/Labels/LabelTableCell.swift#L17-L19)?

Classes/Systems/NSAttributedStringSizing.swift
```diff
@@ -183,3 +183,44 @@ final class NSAttributedStringSizing: NSObject, ListDiffable {
...
+        context?.fill(CGRect(x: rect.minX, y: rect.maxY + (cornerRadius / 2), width: rect.width, height: 1))
+        
+        // Reset color as per Apple docs
+        color.set()
+        
+    }
```
rnystrom: This is surprisingly simple! Very cool!

Classes/Views/AttributedStringView.swift
Sherlouk: I actually added it for that exact reason, as we add more detection for additional types we don't necessarily want to enforce every listener to handle it (slash end up with a whole bunch of empty implementations). I can remove it yea, but was intentional!

Classes/Issues/Labeled/IssueLabeledCell.swift
Sherlouk: The issue is the top/bottom has to be 7 otherwise the spacing is different to the other events. I was testing with the issue attached so if the sizing wasn't the same then you could visible see different padding (and my need for consistency was hurting).

I'll work out something else

Classes/Systems/NSAttributedStringSizing.swift
Sherlouk: Opted to do the same styling as GitHub desktop where it's alpha based on the background so it's slightly darker rather than a fixed grey border.

Happy to make the change though

Classes/Systems/NSAttributedStringSizing.swift
Sherlouk: I intentionally skipped a lot of the unnecessary stuff in the guides you linked, though when I do additional testing for line wrapping etc it may need to be added back in. Was hoping for our needs specifically though we should over do it!

Classes/Systems/NSAttributedStringSizing.swift
rnystrom: Ya consistency with app > everything else

Sherlouk:
Made the changes, we'll probably have to change this as new requirements come in (e.g. adding images etc) but as a starting point it should be solid!

PR #1077, opened 2017-11-25
Search cell shares common height

### LINKED ISSUE #1037: Recent search vs Bookmark cell different heights
Exact same content in both cells (rnystrom/GitHawk) but the bookmark one is a little bit bigger

Only something somebody with mild OCPD would notice but I do think we should try and be as consistent as possible for these things. I would go as far as to say as they should be using the exact same cell. "RepositorySectionController" which is used by both of them

rnystrom:
Fixes #1037


Sherlouk:
@rnystrom So genuine question, why don't we have a single section controller which the soul purpose is displaying a Repository and when you tap on it, you go to that repository?

At the moment we have quite a lot of similar duplication (empty states are another; but getting better with a bit of reuse)

rnystrom:
Ya we need that. Historically search was made, then bookmarks used table view, then we refactored it to use IGLK. I think they’ve just drifted together at this point. Would be great to just use the same section controller!

PR #1078, opened 2017-11-25
Render checkmarks

### LINKED ISSUE #1074: Use NSTextAttachment to render checkboxes
- [x] Need unchecked asset
- [x] Need checked asset
- [ ] Find/replace `[ ]` and `[x]` (does GH support things like `[ x]`?)
- [ ] Create `NSTextAttachment` probably in `NSAttributedStringSizing` alongside find/replace

Prereq for finishing #50 (making interactive)

rnystrom:
- Changes appearance when viewer can and cannot update
- Custom assets mimicking GitHub checkmarks on Chrome

Fixes #1074


rnystrom:
Handling read and unread states:

simulator screen shot - iphone 8 - 2017-11-24 at 23 03 59

simulator screen shot - iphone 8 - 2017-11-24 at 23 04 06

rnystrom:
This was shockingly easy

rnystrom:
Reminder for myself on how to tweak the y position

https://stackoverflow.com/a/28355210

Sherlouk:
Now make them re-orderable :trollface: 😂 

Looks amazing!

Classes/Issues/Comments/Markdown/CommentModelsFromMarkdown.swift
```diff
@@ -404,20 +431,56 @@ func shortenGitHubLinks(attributedString: NSAttributedString,
     return mutableAttributedString
 }
 
+private let checkedIdentifier = ">/CHECKED>/"
+private let uncheckedIdentifier = ">/UNCHECKED>/"
+private let checkmarkRegex = try! NSRegularExpression(pattern: "^- (\\[([ |x])\\])", options: [.anchorsMatchLines])
```
Sherlouk: FWIW you can use checkmarks in numbered lists also, this regex will ignore those!

Classes/Issues/Comments/Markdown/CommentModelsFromMarkdown.swift
```diff
@@ -404,20 +431,56 @@ func shortenGitHubLinks(attributedString: NSAttributedString,
...
+private func annotateCheckmarks(markdown: String) -> String {
+    let matches = checkmarkRegex.matches(in: markdown, options: [], range: markdown.nsrange)
+    let result = NSMutableString(string: markdown)
+    for match in matches.reversed() {
```
Sherlouk: `match.range(at:)` crashes if it doesn't have the range, for a bit of added security should we add a where clause to the for loop in order to prevent this?

`for match in matches.reversed() where match.numberOfRanges == 2 {`

Classes/Issues/Comments/Markdown/CommentModelsFromMarkdown.swift
rnystrom: Hmm, am I doing something wrong [here](https://github.com/rnystrom/GitHawk/issues/1074#issuecomment-346913669)? I tried testing numbered lists.

Classes/Issues/Comments/Markdown/CommentModelsFromMarkdown.swift
rnystrom: Tho by definition of the regex, nothing can match w/out 2 groups (nested)

Classes/Issues/Comments/Markdown/CommentModelsFromMarkdown.swift
Sherlouk: I've updated the comment, and it looks like if you add a word after the checklist with a number then it works - but it's a bit buggy because if you look it splits them up so it's not one list.

This actually might be a GitHub bug!

Classes/Issues/Comments/Markdown/CommentModelsFromMarkdown.swift
Sherlouk: Fair point, really not a fan of the regex API in Swift tbh! Not very safe and often have to cast to NS[String] 😅

Classes/Issues/Comments/Markdown/CommentModelsFromMarkdown.swift
rnystrom: Ya, looks bizarre. I'm going to skip for now.

Classes/Issues/Comments/Markdown/CommentModelsFromMarkdown.swift
rnystrom: Ya definitely showing its age... all the `NSRange` too 🙄

rnystrom:
Note to my future self, if we need to attach custom attributes, we have to do a workaround like this:

```swift
var attributes: [NSAttributedStringKey: Any] = [.attachment: attachment]
attributes[checked ? MarkdownAttribute.checked : MarkdownAttribute.unchecked] = 1
// workaround b/c you can't make an empty attributed string w/ attachment attributes
let attributedString = NSMutableAttributedString(attributedString: NSAttributedString(attachment: attachment))
attributedString.setAttributes(attributes, range: attributedString.string.nsrange)
```

rnystrom:
A little concerned w/ the complexity required to actually check things b/c we have to track the **original** position of the checkmark (before all of our processing). That's going to be pretty tricky.

PR #1087, opened 2017-11-25
Add syntax highlighting for comments

### LINKED ISSUE #59: Syntax highlighting
Not sure how to do this yet. Wonder if I can do some fancy find & replace of https://github.com/isagalaev/highlight.js and convert parts of it to Swift. Looks like its really just loads of regex? Might be easy. All I need to do is turn it into an `NSAttributedString`.

rnystrom:
Fixes #59

simulator screen shot - iphone 8 - 2017-11-25 at 16 07 29


Podfile
```diff
@@ -40,4 +41,14 @@ end
 
 post_install do |installer|
   system("sh tools/generateAcknowledgements.sh")
+
+  # convert incompatible pods back to Swift 3.2
```
Sherlouk: Worth a local fork of this to fix this issue? I'm saying this as it might also be worth updating the local highlight.js dependency (https://github.com/raspu/Highlightr/tree/master/Pod/Assets/Highlighter) to ensure we have the latest styles and syntax!

(Project hasn't been updated in like a year!)

Sherlouk: The version in Highlighter is 9.9.0, in the latest (9.12.0) has fixes to Swift

https://highlightjs.org/

So yeah should probs update! 😄

Sherlouk:
Where's my custom https://github.com/rnystrom/Metro-Lights theme also? 😂 

Not part of this, but would be nice in the future if we could allow basic theming (tl;dr I want a dark theme okay) since it's basically built in! Keep simple for now tho

Podfile
rnystrom: Good call. Let's do that in a followup, want to collect some feedback from beta testers and make sure this thing is stable. Going to merge as-is, but then we can create a local pod and manually fix as we see fit.

PR #1124, opened 2017-12-02
[WIP] Make base url injectable

### LINKED ISSUE #1119: Make base url injectable
Don’t hardcode `github.com` everywhere. Maybe we set this in `NSUserDefaults`?

This will help unlock enterprise support.

See #406

<details>
<pre>
Version 1.15.0 (SEE BUILD PHASE)
Device: iPhone X (iOS 11.1.2)
TestFlight: false
</pre>
</details>

Sherlouk:
Closes #1119 
Dependency on enterprise support

Opening PR for visibility and to easily track the changes I've made. As an initial step I've gone through the app and found all references to `github.com` that we _need_ to address (obviously, skipping comment links to docs etc)


Classes/Systems/GithubAPIDateFormatter.swift
```diff
@@ -10,6 +10,7 @@ import Foundation
 
 private let dateFormatter = ISO8601DateFormatter()
 func GithubAPIDateFormatter() -> ISO8601DateFormatter {
+    // TODO: REVISE (Should be static)
     // https://developer.github.com/v3/#schema
     return dateFormatter
```
Sherlouk: Not sure if I'm going insane here, I can't remember if you have a let at root scope like this if it's the same as static? Might be fine, but we definitely shouldn't be creating multiple date formatters!

rnystrom: Yup! Same as static. You can't static w/out some sort of namespacing (`enum`, `struct`, `class`). This only creates a single instance.

BasThomas:
Love how you've started tackling this, @Sherlouk!

Classes/Systems/GithubAPIDateFormatter.swift
BasThomas: Bit off topic, but recently saw something I really liked in our codebase at work, looking something like this:

```swift
enum A {
  static let
  hi = "hi",
  bye = "bye"
}
```

Using just "one" `static let`.

rnystrom:
Add flakes Falkland fakes don’t flaked lkas fjahksdf has Delia’s dfkjadhfaskdjf akjsdh 

Classes/Systems/Client.swift
```diff
@@ -0,0 +1,98 @@
...
+    }
+    
+    var statusAPIUrl: String? {
+        switch type {
+        // https://status.github.com/api
+        case .github: return "https://status.\(baseUrl)/api/status.json"
+        
+        // TODO: Remove "API Status" as an option in Settings
+        // Unsupported
+        case .githubEnterprise: return nil
```
Sherlouk: This'll work will need to be done as part of the other ticket, but keeping track and laying the basic foundations to make that work easier

Classes/Login/GithubClient+AccessToken.swift
```diff
@@ -27,8 +27,9 @@ extension GithubClient {
         let headers = [
             "Accept": "application/json"
         ]
+        
         request(Request(
-            url: "https://github.com/login/oauth/access_token",
+            url: Client().loginOauthUrl, // Assumes GitHub.com due to login flow
```
Sherlouk: There are some places where I'm creating a new `Client()` this is all around the login flow where the `GitHubClient` doesn't have a reference to the `userSession` and therefore hasn't got a client set.

These will have to be updated once we actually have enterprise login working, but I can't make any assumptions on that work at this stage!

Sherlouk:
@BasThomas @rnystrom 

Okay so the "easy" stuff is done, where I have simple references to the userSession so I can just grab the client and construct URLs that way.

With most of the others we seem to be making assumptions on the structure, the biggest example is where we always assume `hostname.com/Sherlouk` is the correct URL for a profile, but when it's a bot (in this example) it's actually `hostname.com/apps/Sherlouk` - For this I actually wanna make a bit of a change where we add a couple new values to the GQL responses to actually retrieve the URLs from there? That way we basically don't need to do any URL building ourselves, we just send them to the URL GitHub gives us!

Are we okay with the changes made so far, and the direction I'm planning on taking this? Speak now or <s><sub>forever hold your peace</sub></s> don't and just decline the PR at the end 😂

rnystrom:
One push back against this is the fact that enterprise support won’t ever really happen given that we sync our GraphQL schema against GitHub.com. Talked about this with the GitHub team. Would have to update your enterprise version to whatever min version we use, and then we have to communicate every time we update 😭

Sherlouk:
I still think this is an important change to be honest, it moves us away from all of these fixed URLs which causes issues like bot posts opening invalid URLs

Irrespective if this moves us closer to enterprise, which I still think can be done with some work and limited functionality!

BasThomas:
You are right @rnystrom, that is an issue. We tried in [this branch](https://github.com/rnystrom/GitHawk/tree/make_enterprise_work_hack) (which builds on top of #1549) and we can login and see a list of issues (v3 api).

As soon as we open an issue, it blows up, where we run into GraphQL errors like these:

```
(lldb) po result?.errors
▿ Optional<Array<GraphQLError>>
  ▿ some : 3 elements
    ▿ 0 : Field 'changedFiles' doesn't exist on type 'PullRequest'
      ▿ object : 2 elements
        ▿ 0 : 2 elements
          - key : "locations"
          ▿ value : 1 element
            ▿ 0 : 2 elements
              ▿ 0 : 2 elements
                - key : line
                - value : 581
              ▿ 1 : 2 elements
                - key : column
                - value : 9
        ▿ 1 : 2 elements
          - key : "message"
          - value : Field 'changedFiles' doesn't exist on type 'PullRequest'
    ▿ 1 : Field 'issues' doesn't exist on type 'Milestone'
      ▿ object : 2 elements
        ▿ 0 : 2 elements
          - key : "locations"
          ▿ value : 1 element
            ▿ 0 : 2 elements
              ▿ 0 : 2 elements
                - key : line
                - value : 645
              ▿ 1 : 2 elements
                - key : column
                - value : 3
        ▿ 1 : 2 elements
          - key : "message"
          - value : Field 'issues' doesn't exist on type 'Milestone'
    ▿ 2 : Field 'issues' doesn't exist on type 'Milestone'
      ▿ object : 2 elements
        ▿ 0 : 2 elements
          - key : "locations"
          ▿ value : 1 element
            ▿ 0 : 2 elements
              ▿ 0 : 2 elements
                - key : line
                - value : 649
              ▿ 1 : 2 elements
                - key : column
                - value : 3
        ▿ 1 : 2 elements
          - key : "message"
          - value : Field 'issues' doesn't exist on type 'Milestone'
```

Which makes sense, look at [the enterprise docs](https://developer.github.com/enterprise/2.12/v4/object/milestone/) and [the GitHub.com docs](https://developer.github.com/v4/object/milestone/).

And making things optional (if that's even possible with GraphQL?) doesn't quite scale, especially when you can't really test it either.

rnystrom:
Cleaning up, this is pretty behind

PR #1127, opened 2017-12-02
Don't collide gestures on AttributedStringView

### LINKED ISSUE #1125: Only recognize taps in AttributedStringView when interactive attributes detected
Noticed when fixing #1096 that if you tap the author, a profile VC _and_ the cell's `didSelect` action are pushed onto the nav stack. This is b/c the gesture doesn't cancel touches in view.

We need to add the gesture delegate and not allow the gesture to begin when an attribute isn't detected.

rnystrom:
Fixes #1125

### DIFF
--- a/Classes/Issues/Comments/Markdown/MarkdownAttribute.swift
+++ b/Classes/Issues/Comments/Markdown/MarkdownAttribute.swift
@@ -18,4 +18,16 @@ enum MarkdownAttribute {
     static let details = NSAttributedStringKey(rawValue: "com.freetime.Markdown.details")
     static let label = NSAttributedStringKey(rawValue: "com.freetime.Markdown.label")
     static let commit = NSAttributedStringKey(rawValue: "com.freetime.Markdown.commit")
+
+    static let all: Set<NSAttributedStringKey> = Set<NSAttributedStringKey>([
+        url,
+        email,
+        username,
+        usernameDisabled,
+        linkShorteningDisabled,
+        issue,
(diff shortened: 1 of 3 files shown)


Classes/Issues/Comments/Markdown/MarkdownAttribute.swift
```diff
@@ -18,4 +18,16 @@ enum MarkdownAttribute {
     static let details = NSAttributedStringKey(rawValue: "com.freetime.Markdown.details")
     static let label = NSAttributedStringKey(rawValue: "com.freetime.Markdown.label")
     static let commit = NSAttributedStringKey(rawValue: "com.freetime.Markdown.commit")
+
+    static let all: Set<NSAttributedStringKey> = Set<NSAttributedStringKey>([
```
BasThomas: Isn't the `: Set<NSAttributedStringKey` obsolete?

Classes/Views/AttributedStringView.swift
```diff
@@ -26,6 +26,8 @@ final class AttributedStringView: UIView {
     weak var issueDelegate: AttributedStringViewIssueDelegate?
 
     private var text: NSAttributedStringSizing?
+    private var tapGesture: UITapGestureRecognizer?
```
BasThomas: Why are these optional?

rnystrom: Takes `self` as a target so I can't set on init

PR #1129, opened 2017-12-02
Prevent SLKTextViewController cache from overwriting text

### LINKED ISSUE #1095: Comment image uploading broken when already typed something
To reproduce:

Start a comment
Upload an image

The image markdown is never added to your WIP comment

<details>
<pre>
Version 1.14.0 (1230)
Device: iPhone X (iOS 11.1.2)
TestFlight: true
</pre>
</details>

rnystrom:
@Sherlouk was spot on!

Fixes #1095

Also changing image insertion to be at the cursor

### DIFF
--- a/Classes/Views/TextActionsController.swift
+++ b/Classes/Views/TextActionsController.swift
@@ -74,7 +74,7 @@ final class TextActionsController: NSObject,
     // MARK: ImageUploadDelegate
 
     func imageUploaded(link: String, altText: String) {
-        textView?.replace(left: "![\(altText)](\(link))", right: nil, atLineStart: true)
+        textView?.replace(left: "![\(altText)](\(link))", right: nil, atLineStart: false)
     }
 
 }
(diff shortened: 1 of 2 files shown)


BasThomas:
Oh ha, I should've assigned the issue. Was working on something similar.

Not sure if this would be exactly what we want in all situations? This is what I came up with:

```swift
self?.dismiss(animated: true) { [weak delegate = self?.delegate] in
    delegate?.imageUploaded(link: link, altText: name)
}
```

Classes/Views/TextActionsController.swift
```diff
@@ -74,7 +74,7 @@ final class TextActionsController: NSObject,
     // MARK: ImageUploadDelegate
 
     func imageUploaded(link: String, altText: String) {
-        textView?.replace(left: "![\(altText)](\(link))", right: nil, atLineStart: true)
+        textView?.replace(left: "![\(altText)](\(link))", right: nil, atLineStart: false)
```
BasThomas: Nice catch, saw this too and was going to ask.

PR #1131, opened 2017-12-02
[Cleanup] Remove nil-default completion parameters

BasThomas:


Classes/Image Upload/ImageUploadTableViewController.swift
```diff
@@ -189,14 +189,12 @@ class ImageUploadTableViewController: UITableViewController {
 
                 switch result {
                 case .error:
-                    print("error")
                     ToastManager.showGenericError()
                     self?.setRightBarItemIdle()
 
                 case .success(let link):
-                    print("success")
```
BasThomas: Also removed these prints, I felt like if there was a good reason to keep them they should at least be more descriptive. (cc @Sherlouk)

Sherlouk: Debug messages, forgot to remove!

BasThomas: That's what I thought, cool!

Classes/Repository/RepositoryIssuesViewController.swift
```diff
@@ -84,7 +84,7 @@ SearchBarSectionControllerDelegate {
     // MARK: SearchBarSectionControllerDelegate
 
     func didChangeSelection(sectionController: SearchBarSectionController, query: String) {
-        filter(query: query.trimmingCharacters(in: .whitespacesAndNewlines), animated: true)
+        filter(query: query, animated: true)
```
BasThomas: Removed as this is redundant; it already happens in [`filtered()`](https://github.com/rnystrom/GitHawk/blob/master/Classes/Utility/Filterable.swift#L17).

PR #1135, opened 2017-12-02
Improve scroll perf of DotListView

rnystrom:
Noticed the `reloadData` was eating a ton of perf. Also added a border so labels similar to background color are noticeable.


joerocca:
@rnystrom This may not be necessary, but removing this prevents recalculating the number of dots that fit on the current orientation on orientation change.

rnystrom:
Oh interesting. We could follow up with checking if the collection view frame is changing, _then_ reload. @joerocca wanna PR that?

joerocca:
@rnystrom Yes!

PR #1137, opened 2017-12-02
Repo code highlighting

rnystrom:
Slick

img_0233


jwandrews:
Ahh yes! This was one of the things I’ve been wanting in this app! Big 👍 would be nice to allow a user preference option for setting a theme from available options

Issue #1134, opened 2017-12-02
"Project has no Readme" when offline


BasThomas:
When not connected to the internet, browsing to a repo will have the overview stating there is no readme, even though there could be one.

Sherlouk:
I wonder if we should re-approach caching? At the moment we basically don't cash _anything_ - READMEs don't change often, so would be nice to cache that until we get a new copy or somethin'

BasThomas:
After I created this one I was thinking about the whole offline thing - most of the time it probably doesn't make sense to have stuff offline, so maybe we should skip it completely.

PR #1140, opened 2017-12-03
Actually show an empty state on base list vc

rnystrom:
I'm not sure why this was never the case.... 😱


rnystrom:
Required w/ #1141 to fully fix #1134

PR #1167, opened 2017-12-03
[Accessibility] Disable animations when reduce motion enabled

BasThomas:
This disables animations if the user has reduce motion enabled.

Ref #118


BasThomas:
Gtg after #1208 👍

PR #1198, opened 2017-12-06
Log when prompting for rating

### LINKED ISSUE #1182: Add logging event when prompting for rating
Should add for both types. Can see how often they trigger (even if we don’t know for sure if the native one appears).

Maybe couple with how often the rating controller is _asked_ to prompt, that way we can see a ratio and frequency.

<details>
<pre>
Version 1.15.0 (1512339646)
Device: iPhone X (iOS 11.1.2)
TestFlight: true
</pre>
</details>

rnystrom:
Fixes #1182

### DIFF
--- a/Classes/Systems/Rating/RatingController.swift
+++ b/Classes/Systems/Rating/RatingController.swift
@@ -9,6 +9,7 @@
 import Foundation
 import StoreKit
 import IGListKit
+import Crashlytics
 
 final class RatingToken: NSObject, ListDiffable {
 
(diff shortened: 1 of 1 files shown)


Classes/Systems/Rating/RatingController.swift
```diff
@@ -57,10 +58,17 @@ final class RatingController {
     class func prompt(_ type: Prompt) {
         guard canPrompt(.system) else { return }
 
+        let typeString: String
         switch type {
-        case .system: requestSystem()
-        case .inFeed: openAppStore()
+        case .system:
+            typeString = "system"
```
BasThomas: What about adding these typeStrings to the enum itself?

rnystrom: Ugh that’s a way better idea. I’ll update.

PR #1203, opened 2017-12-08
Added wrapping labels view

### LINKED ISSUE #951: Redesign labels cell as h-scroll
Instead of the dots + tap-expand, what if we just make an h-scroll unit to display all the labels? Might be a better use of available whitespace, and easier to use.

Thoughts?

One constraint is that cells must be dynamically sized (prob not estimated size b/c of issues, but maybe it'd work here?). Makes it slightly more complicated.

joerocca:
Closes #951

img_0051

img_0052


rnystrom:
Omg @joerocca that looks so cool!!

Classes/Issues/Labels/IssueLabelSummaryModel.swift
```diff
@@ -11,12 +11,18 @@ import IGListKit
...
 
-    let title = NSLocalizedString("Labels:", comment: "")
-    let colors: [UIColor]
+    let labels: [RepositoryLabel]
     let _diffIdentifier: String
-
-    init(colors: [UIColor]) {
-        self.colors = colors
+    
+
```
rnystrom: nit: too much whitespace

Classes/Issues/Labels/IssueLabelSummaryModel.swift
```diff
@@ -11,12 +11,18 @@ import IGListKit
 
 final class IssueLabelSummaryModel: ListDiffable {
 
-    let title = NSLocalizedString("Labels:", comment: "")
-    let colors: [UIColor]
+    let labels: [RepositoryLabel]
```
rnystrom: can we make this `private`?

Classes/Issues/Labels/IssueLabelSummaryModel.swift
```diff
@@ -11,12 +11,18 @@ import IGListKit
...
     let _diffIdentifier: String
-
-    init(colors: [UIColor]) {
-        self.colors = colors
+    
+    
+    // quicker comparison for diffing rather than scanning the labels array
+    private let labelSummary: String
```
rnystrom: great!!

Classes/Issues/Labels/IssueLabelsSectionController.swift
```diff
@@ -10,10 +10,8 @@ import UIKit
...
 
 final class IssueLabelsSectionController: ListBindingSectionController<IssueLabelsModel>,
-    ListBindingSectionControllerDataSource,
-ListBindingSectionControllerSelectionDelegate {
+    ListBindingSectionControllerDataSource {
 
-    private var expanded = false
     private let issueModel: IssueDetailsModel
     private let client: GithubClient
     private var labelsOverride: [RepositoryLabel]?
```
rnystrom: This actually isn't used anymore, wanna remove it?

PR #1206, opened 2017-12-09
Dynamic content size support

### LINKED ISSUE #1016: Dynamic font support
- [Style](https://github.com/rnystrom/GitHawk/blob/master/Classes/Views/Styles.swift) font sizes should be functions that return a relative font based on system settings
  - e.g. baseline `body` is size 16
- Listen for `UIContentSizeCategoryDidChangeNotification` and:
  - Rebuild all models since `NSAttributedStringSizing` will have the font stored in its `NSAttributedString`
  - Call `listAdapter.reloadData()` to rebuild section controllers

rnystrom:
Fixes #1016

### DIFF
--- a/Classes/Settings/Settings.storyboard
+++ b/Classes/Settings/Settings.storyboard
@@ -23,7 +23,7 @@
                             <rect key="frame" x="0.0" y="561.5" width="375" height="44"/>
                             <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMaxY="YES"/>
                             <subviews>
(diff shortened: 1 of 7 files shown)


BasThomas:
If I understand this correctly a user would have to kill and relaunch the app for the text size changes to go into effect, right?

rnystrom:
@BasThomas at the moment ya, I started to respond to content size notifications, but b/c of the way text models are created from content (and are immutable), it'll require a BIG refactor to support. Still uncertain how to do it. I'm _thinking_ about a new text-builder pattern to make it work, but not sure yet.

I think shipping w/ basic support is a great v1 tho!

rnystrom:
Alrighty, gonna merge this, I think it's good enough to start testing. Hopefully everyone can help find some unsupported stuff and we can patch! Really excited for 1.16 now!

BasThomas:
Definitely a good v1, didn't want to suggest otherwise! Love this

PR #1208, opened 2017-12-09
Don't scroll beyond the top offset

rnystrom:
Stops `SLKTextViewController` from scrolling beyond the top offset when showing the keyboard. There was an issue for this I think? I'll dig it up.


rnystrom:
#1068

BasThomas:
Might also unblock #1167, I will check

Issue #1222, opened 2017-12-10
NSAttributedStringSizing Refactor


rnystrom:
In order to unlock a range of features, I'm proposing we refactor this text class into a bunch of smaller components. This is required before stuff like:

- Dynamic content size changes w/out app restart
- Tapping checkboxes to change state
- Performance enhancements (warm up bitmap caches, memoize string building)
- Scaling text sizes

Some thoughts on things I think we could do:

- [ ] Put all threaded params into a single `struct`. This will make extending params much easier.
- [ ] Move size caching into a singleton completely decoupled from `NSAttributedStringSizing`.
- [ ] Move bitmap caching into a singleton completely decoupled from `NSAttributedStringSizing`.
- [ ] Convert `Styles` into an actual `enum` with functions to get dynamic size and font. Required before we create builders
- [ ] Create `struct StyledString` that has properties:
  - `let attributes: [NSAttributedStringKey: Any]`
  - `let string: String`
  - `let font: Styles.Font` (or wherever that lands)
  - `let markdown: MarkdownAttributes`
- [ ] Create common style helpers (body, bold, username, etc)
- [ ] Create builder that let's you combine `StyledString`s with "render" function that outputs `NSAttributedString`. Can be memoized.
  - Use this when content size changes, memoization/cache should miss and return new `NSAttributedString`

lixing520:
- 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #1231, opened 2017-12-11
[WIP] Exploring styled text refactor

rnystrom:
⚠️ Very WIP ⚠️ 

Keep reviews brief, high chance that this will change drastically, frequently.

Part of #1222


rnystrom:
W/ heavy use of hash, can start to memoize string creation:

https://gist.github.com/maximveksler/3d4deca1912870bdb142a1ce27adcd71

rnystrom:
Current WIP, getting pretty excited about the builder direction:

https://gist.github.com/rnystrom/02a8508b8840d4121e487f4d3fa37253

Local Pods/StyledText/StyledTextTests/StyledTextBuilderTests.swift
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    func test_whenAddingAttributes_withSavingState_thenRestoring() {
+        let render = StyledTextBuilder(styledText: StyledText(text: "foo", style: TextStyle(traits: .traitBold)))
+            .save()
+            .add(styledText: StyledText(text: " bar", style: TextStyle(traits: .traitItalic)))
+            .restore()
+            .add(text: " baz")
+            .render(contentSizeCategory: .large)
+        XCTAssertEqual(render.string, "foo bar baz")
```
rnystrom: @BasThomas @Sherlouk quick peek at how this API will work. I'm pretty happy w/ the direction.

- Moves the `UIApplication` API to the render function so we could call it on main, getting rid of the warning.
- Save/restore attribute state (e.g. push/pop)
- `NSAttributedString` rendered w/ passed-in content size so that the builder could be maybe once and output different sizes, like after someone changes their settings

rnystrom:
This PR is ballooning, but making some awesome stuff:

- New `LRUCache` object, probably going to split this into its own local pod. Thread safe!
- String builder API that is fully immutable but w/ a clean, builder pattern
- Renderer object backed by snapshot tests
- Custom view object

rnystrom:
I rolled back changes to `Classes` b/c it was getting out of hand. Instead I'm going to land this new framework and we can start refactoring things piece-by-piece.

One big discovery I made is that we can patch MMarkdown to do scanning for:

- Usernames
- Auto link by repo+owner+issue
- Checkboxes (including state)

We can also move URL shortening away from being regex to just when we put the link in the builder.

This is a big deal b/c it'll finally help us map parsed-checkbox location to its range in the original markdown. Huge, we'll be able to tap checkboxes and send edits!

This is going to take a _while_ to finish, but it'll all be worth it! In the end we'll have:

- A new text framework we can move to @GitHawkApp as a standalone pod
- Much better performance
- Thread safety (there isn't any atm)
- Bounded memory usage
- Unit and snapshot tests
- Crazy simpler string building (like omg its so much better)
- Checkbox editing ✅ 🎉

BasThomas:
So awesome! 😀 Worth a blog post if you have some time maybe? 😉

rnystrom:
@BasThomas great idea. I have a few blog post ideas queued up. Let me spin those off!

> if you have some time

😆

BasThomas:
But first!
[GitHawk Upload by BasThomas](https://i.imgur.com/e6OWGOC.jpg)

PR #1245, opened 2017-12-12
Added wrapping labels view to RepoSummaryCell

### LINKED ISSUE #1220: Replace labels dot view with text labels view in issue/pr cells?
I am a fan of this on GitHub web

GitHawk Upload by joerocca

joerocca:
Closes #1220 

img_0109


Sherlouk:
This looks so nice!

Classes/Views/LabelListView.swift
```diff
@@ -0,0 +1,88 @@
...
+//  Created by Joe Rocca on 12/6/17.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+final class LabelListView: UIView, UICollectionViewDataSource, UICollectionViewDelegate, UICollectionViewDelegateFlowLayout {
+    
+    static func height(width: CGFloat, labels: [RepositoryLabel], cacheKey: String) -> CGFloat {
+        if let cachedHeight = LabelListViewHeightCache.shared.height(forKey: cacheKey) {
```
rnystrom: One crucial issue: the height is only cached based on the given key. What happens when you change orientations? I think internally here we should just merge the **width and key** into a shared key and it will work?

```swift
let key = "\(width)\(cacheKey)"
let cachedHeight = LabelListViewHeightCache.shared.height(forKey: key)
```

Something like that.

Also, can we just define the cache within this class? The cache object itself is very simple.

```swift
private static let cache = [String: CGFloat]()
```

Classes/Views/LabelListView.swift
```diff
@@ -0,0 +1,88 @@
...
+            return cachedHeight
+        }
+        
+        let rowHeight = LabelListCell.size(labels.first?.name ?? "").height
+        var labelTextTotalWidth = CGFloat()
+        for label in labels {
+            labelTextTotalWidth += LabelListCell.size(label.name).width
+        }
```
rnystrom: Same thing, but simpler and w/out leaving a dangling mutable variable

```swift
let labelTextTotalWidth = labels.reduce(0, { $0 + LabelListCell.size($1.name).width })
```

Classes/Views/LabelListView.swift
joerocca:
```swift
let height = LabelListView.height(width: width, labels: labels, cacheKey: cacheKey + "\(width)")
```
I am actually doing this like above, but internally is way better!
Sounds good with the cache!

joerocca:
Good to go!

PR #1278, opened 2017-12-17
Log device environment info

rnystrom:
I'd like to get a better understanding of user demographics to prioritize other languages, accessibility, and RTL. Doesn't mean if there are very few users we give up. Also justification for working on stuff, knowing that we're helping ppl.

However I could understand push back against logging ppl's information. If people strongly disagree, I wont merge this.

### DIFF
--- a/Classes/Systems/AppDelegate.swift
+++ b/Classes/Systems/AppDelegate.swift
@@ -61,6 +61,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // setup 3d touch shortcut handling
         ShortcutHandler.configure(application: application, sessionManager: sessionManager)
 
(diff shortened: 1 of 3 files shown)


BasThomas:
What about a toggle in Settings to turn it off? Would still track it until the user would see that though.

Not opposed though - would love to see these numbers!

rnystrom:
@BasThomas wow what a great idea. I actually think having a toggle to disable ALL logging would be a nice gesture.

Sherlouk:
Fwiw crashlytics has a delegate to stop tracking etc so would explore that!

BasThomas:
Oh yeah, that's what I meant. :)

PR #1291, opened 2017-12-19
View notifications by repository

### LINKED ISSUE #61: Support for filtering/grouping of notifications by repository
Freetime currently displays a list of notifications from a user's public repositories (notifications from private repositories are currently not displayed, see #60 for a discussion). These notifications are ordered via their timestamp in chronological order. The list may contain several hundred read/unread notifications. 

It will be great if we could apply a "filter" to select one or more repositories the user is subscribed for notifications and show notifications from only the selected repositories. In addition, notifications can possibly be grouped by repositories 🤔 . This will probably make things complicated as we would want pagination of notifications. 

image

rnystrom:
- New left/right nav bar items for inbox
- Show alert, select repo
- `NotificationsViewController` now handles unread, all, and repo (show all for repo)
- "Mark Read" for repo clears _only_ that repo's notifications

Pretty happy w/ this quick hack. Been using it all night. This will get us some immediate feedback before investing much in custom UI controls (e.g. the whole nav-bar dropdown thing).

Fixes #61 

img_0554

img_0555

### DIFF
--- a/Classes/Notifications/NotificationClient.swift
+++ b/Classes/Notifications/NotificationClient.swift
@@ -56,7 +56,7 @@ final class NotificationClient {
 
         let path: String
         if let repo = repo {
-            path = "/repos/\(repo.owner)/\(repo.name)/notifications"
+            path = "repos/\(repo.owner)/\(repo.name)/notifications"
         } else {
             path = "notifications"
         }
@@ -157,7 +157,7 @@ final class NotificationClient {
         let path: String
         if let repo = repo {
             // https://developer.github.com/v3/activity/notifications/#mark-notifications-as-read-in-a-repository
-            path = "/repos/\(repo.owner)/\(repo.name)/notifications"
+            path = "repos/\(repo.owner)/\(repo.name)/notifications"
         } else {
             // https://developer.github.com/v3/activity/notifications/#mark-as-read
             path = "notifications"
--- a/Classes/Notifications/NotificationClient.swift
+++ b/Classes/Notifications/NotificationClient.swift
@@ -123,4 +123,28 @@ final class NotificationClient {
         })
     }
 
+    func fetchWatchedRepositories(completion: @escaping (Result<[Repository]>) -> Void) {
+        guard let viewer = githubClient.userSession?.username else {
+            completion(.error(nil))
+            return
+        }
+
+        githubClient.request(GithubClient.Request(
+            path: "users/\(viewer)/subscriptions"
+        ) { (response, _) in
+            // https://developer.github.com/v3/activity/watching/#list-repositories-being-watched
+            if let jsonArr = response.value as? [ [String: Any] ] {
+                var repos = [Repository]()
+                for json in jsonArr {
+                    if let repo = Repository(json: json) {
+                        repos.append(repo)
+                    }
+                }
+                completion(.success(repos))
+            } else {
+                completion(.error(response.error))
+            }
+        })
+    }
+
 }
--- a/Classes/Notifications/NotificationClient.swift
+++ b/Classes/Notifications/NotificationClient.swift
@@ -31,7 +31,8 @@ final class NotificationClient {
     }
 
     // https://developer.github.com/v3/activity/notifications/#list-your-notifications
-    func requestNotifications(
+    func fetchNotifications(
+        repo: (owner: String, name: String)? = nil,
         all: Bool = false,
         participating: Bool = false,
         since: Date? = nil,
@@ -53,10 +54,17 @@ final class NotificationClient {
             parameters["before"] = GithubAPIDateFormatter().string(from: before)
         }
 
+        let path: String
+        if let repo = repo {
+            path = "/repos/\(repo.owner)/\(repo.name)/notifications"
+        } else {
+            path = "notifications"
+        }
+
         let cache = githubClient.cache
 
         githubClient.request(GithubClient.Request(
-            path: "notifications",
+            path: path,
             method: .get,
             parameters: parameters,
             headers: nil
@@ -71,16 +79,12 @@ final class NotificationClient {
         })
     }
 
-    typealias MarkAllCompletion = (Bool) -> Void
-    func markAllNotifications(completion: MarkAllCompletion? = nil) {
-        githubClient.request(GithubClient.Request(
-            path: "notifications",
-            method: .put) { (response, _) in
-                guard let completion = completion else { return }
-                // https://developer.github.com/v3/activity/notifications/#mark-as-read
-                let success = response.response?.statusCode == 205
-                completion(success)
-        })
+    func markAllNotifications(completion: @escaping (Bool) -> Void) {
+        markNotifications(repo: nil, completion: completion)
+    }
+
+    func markRepoNotifications(owner: String, repo: String, completion: @escaping (Bool) -> Void) {
+        markNotifications(repo: (owner, repo), completion: completion)
     }
 
     func notificationOpened(id: String) -> Bool {
@@ -144,6 +148,27 @@ final class NotificationClient {
             } else {
                 completion(.error(response.error))
             }
+        })
+    }
+
+    // MARK: Private API
+
+    func markNotifications(repo: (owner: String, name: String)?, completion: @escaping (Bool) -> Void) {
+        let path: String
+        if let repo = repo {
+            // https://developer.github.com/v3/activity/notifications/#mark-notifications-as-read-in-a-repository
+            path = "/repos/\(repo.owner)/\(repo.name)/notifications"
+        } else {
+            // https://developer.github.com/v3/activity/notifications/#mark-as-read
+            path = "notifications"
+        }
+
+        githubClient.request(GithubClient.Request(
+            path: path,
+            method: .put
+        ) { (response, _) in
+            let success = response.response?.statusCode == 205
+            completion(success)
         })
     }
 
--- a/Classes/Notifications/NotificationClient.swift
+++ b/Classes/Notifications/NotificationClient.swift
@@ -10,6 +10,11 @@ import Foundation
 
 final class NotificationClient {
 
+    struct NotificationRepository {
+        let owner: String
+        let name: String
+    }
+
     private var openedNotificationIDs = Set<String>()
 
     let githubClient: GithubClient
@@ -32,7 +37,7 @@ final class NotificationClient {
 
     // https://developer.github.com/v3/activity/notifications/#list-your-notifications
     func fetchNotifications(
-        repo: (owner: String, name: String)? = nil,
+        repo: NotificationRepository? = nil,
         all: Bool = false,
         participating: Bool = false,
         since: Date? = nil,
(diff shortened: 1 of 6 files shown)


BasThomas:
What is the expected behavior for showing repos? If I have no unreads, does it automatically open archives? (and why has that button been renamed to "View Read"?)

Classes/Notifications/NotificationClient.swift
```diff
@@ -71,16 +79,12 @@ final class NotificationClient {
...
-        githubClient.request(GithubClient.Request(
-            path: "notifications",
-            method: .put) { (response, _) in
-                guard let completion = completion else { return }
-                // https://developer.github.com/v3/activity/notifications/#mark-as-read
-                let success = response.response?.statusCode == 205
-                completion(success)
-        })
+    func markAllNotifications(completion: @escaping (Bool) -> Void) {
+        markNotifications(repo: nil, completion: completion)
```
BasThomas: You can omit the `repo` parameter as it defaults to `nil`. :)

Classes/Notifications/NotificationClient.swift
```diff
@@ -123,4 +127,49 @@ final class NotificationClient {
...
+                completion(.success(repos.sorted { $0.name < $1.name }))
+            } else {
+                completion(.error(response.error))
+            }
+        })
+    }
+
+    // MARK: Private API
+
+    func markNotifications(repo: (owner: String, name: String)?, completion: @escaping (Bool) -> Void) {
```
BasThomas: Looks pretty similar to https://github.com/rnystrom/GitHawk/pull/1291/files#diff-d08adcb8b7b2ae5fc2e0f5abfaa92a42R34, maybe you can share some logic?

Classes/Notifications/NotificationClient.swift
```diff
@@ -123,4 +127,49 @@ final class NotificationClient {
...
+            path = "repos/\(repo.owner)/\(repo.name)/notifications"
+        } else {
+            // https://developer.github.com/v3/activity/notifications/#mark-as-read
+            path = "notifications"
+        }
+
+        githubClient.request(GithubClient.Request(
+            path: path,
+            method: .put
+        ) { (response, _) in
```
BasThomas: Nit, but you can omit the `()` for `(response, _)`.

Classes/Notifications/NotificationSubscriptionsController.swift
```diff
@@ -0,0 +1,74 @@
...
+//
+//  Created by Ryan Nystrom on 12/18/17.
+//  Copyright © 2017 Ryan Nystrom. All rights reserved.
+//
+
+import UIKit
+
+final class NotificationSubscriptionsController {
+
+    enum State {
```
BasThomas: Where did you get this from? I love it 😀

Classes/Notifications/NotificationSubscriptionsController.swift
```diff
@@ -0,0 +1,74 @@
...
+                ToastManager.showGenericError()
+            }
+        }
+    }
+
+    // MARK: Private API
+
+    var refreshAction: UIAlertAction {
+        return UIAlertAction(
+            title: NSLocalizedString("Refresh", comment: ""),
```
BasThomas: Wondering where this is used - why would we "hide" a refresh in a `UIAlertAction` if anywhere else we use a `UIRefreshControl`?

Classes/Notifications/NotificationSubscriptionsController.swift
```diff
@@ -0,0 +1,74 @@
...
+    func repoActions(repos: [Repository]) -> [UIAlertAction] {
+        return repos.map {
+            let repo = $0
+            return UIAlertAction(title: repo.name, style: .default, handler: { [weak self] _ in
+                self?.pushRepoNotifications(owner: repo.owner.login, name: repo.name)
+            })
+        }
+    }
+
+    func pushRepoNotifications(owner: String, name: String) {
```
BasThomas: Maybe `pushRepoNotificationsController` or similar? Was a tiny bit confused with `git push` etc etc.

BasThomas: Also (unrelated to this func), we are using a combination of `owner` and `name` quite a bit (especially in this PR). Maybe wrap it in a quick struct (or see if we can use an existing one; I bet we can)?

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -20,24 +20,41 @@ TabNavRootViewControllerType,
...
+        self.inboxType = inboxType
 
         super.init(
             emptyErrorMessage: NSLocalizedString("Cannot load your inbox.", comment: ""),
             dataSource: self
         )
 
         self.foreground.delegate = self
+
+        switch inboxType {
```
BasThomas: What about doing this logic in an `extension NotificationsViewController.InboxType` to prevent cluttering init?

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -47,53 +64,76 @@ FlatCacheListener {
     override func viewDidLoad() {
         super.viewDidLoad()
 
-        resetLeftBarItem()
-        if !showRead {
-            navigationItem.rightBarButtonItem = UIBarButtonItem(
-                title: NSLocalizedString("Archives", comment: ""),
+        resetRightBarItem()
```
BasThomas: Whyyyyy are you swapping the buttons?

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -47,53 +64,76 @@ FlatCacheListener {
...
+        switch inboxType {
+        case .unread:
+            let item = UIBarButtonItem(
+                image: UIImage(named: "bullets-hollow"),
                 style: .plain,
                 target: self,
-                action: #selector(NotificationsViewController.onViewAll)
+                action: #selector(NotificationsViewController.onMore)
             )
+            item.accessibilityLabel = NSLocalizedString("More options", comment: "")
```
BasThomas: 👊😍

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -47,53 +64,76 @@ FlatCacheListener {
     override func viewDidLoad() {
         super.viewDidLoad()
 
-        resetLeftBarItem()
-        if !showRead {
-            navigationItem.rightBarButtonItem = UIBarButtonItem(
-                title: NSLocalizedString("Archives", comment: ""),
+        resetRightBarItem()
+
+        switch inboxType {
```
BasThomas: `if case .unread = inboxType`? See my beautiful [blog post](https://basthomas.github.io/switches-and-cases).

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -47,53 +64,76 @@ FlatCacheListener {
...
     // MARK: Private API
 
-    @objc func onViewAll() {
-        let controller = NotificationsViewController(client: client, showRead: true)
-        controller.title = NSLocalizedString("Archives", comment: "")
+    @objc func onMore() {
+        let alert = UIAlertController.configured(preferredStyle: .actionSheet)
+
+        alert.add(action: UIAlertAction(
+            title: NSLocalizedString("View Read", comment: ""),
```
BasThomas: Also commented in the PR itself, but why did you rename this?

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -47,53 +64,76 @@ FlatCacheListener {
...
+            style: .default,
+            handler: { [weak self] _ in
+                self?.onViewAll()
+        }))
+        subscriptionController?.actions.forEach({ alert.add(action: $0) })
```
BasThomas: Can omit the `()`.

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -47,53 +64,76 @@ FlatCacheListener {
...
-            )
-        }
+    func resetRightBarItem() {
+        let item = UIBarButtonItem(
+            image: UIImage(named: "check"),
+            style: .plain,
+            target: self,
+            action: #selector(NotificationsViewController.onMarkAll)
+        )
+        item.accessibilityLabel = NSLocalizedString("", comment: "")
```
BasThomas: 🤔

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -108,18 +148,35 @@ FlatCacheListener {
...
     }
 
-    @objc private func onMarkAll(sender: UIBarButtonItem) {
+    @objc private func onMarkAll() {
+        let message: String
+        switch inboxType {
+        case .all, .unread:
+            message = NSLocalizedString("Mark all notifications as read?", comment: "")
+        case .repo(_, let name):
+            let messageFormat = NSLocalizedString("Mark %@ notifications as read?", comment: "")
```
BasThomas: Saw this and my initial thought was "shouldn't this be `%d` to show the number of notifications?" and then I saw that this is for the repo name. What about adding the amount of notifications to this message?

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -159,17 +216,37 @@ FlatCacheListener {
...
     override func fetch(page: NSNumber?) {
+        subscriptionController?.fetchSubscriptions()
+
         let width = view.bounds.width
+
+        let repo: (String, String)?
+        switch inboxType {
+        case .repo(let owner, let name):
+            repo = (owner, name)
+        case .all:
```
BasThomas: `case .all, .unread:`

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -184,9 +261,10 @@ FlatCacheListener {
     func models(listAdapter: ListAdapter) -> [ListDiffable] {
         var models = [NotificationViewModel]()
 
+        let showAll = self.showAll
```
BasThomas: Why shadow?

Classes/Notifications/NotificationSubscriptionsController.swift
rnystrom: Haha, what do you mean?

Classes/Notifications/NotificationSubscriptionsController.swift
rnystrom: So you should almost never see this. I'm making sure that if the fetch request to get your watched repos fails, there's at least a button to refresh it. Maybe it should say something like "Error - Refresh"? Idk, there's such limited space here.

Classes/Notifications/NotificationSubscriptionsController.swift
rnystrom: Ya, a `struct` would be better. Getting a little out of hand here.

Classes/Notifications/NotificationsViewController.swift
rnystrom: Yup. In the repo notifications view, the left button will be "back". So I'm moving the "mark all" to the right so that it's consistent between views.

Classes/Notifications/NotificationsViewController.swift
rnystrom: TIL!

Classes/Notifications/NotificationsViewController.swift
rnystrom: Tbh I'm not sure, lol. I'm going to change it back and be "View Archived"

Classes/Notifications/NotificationsViewController.swift
rnystrom: oops

Classes/Notifications/NotificationsViewController.swift
rnystrom: Don't need to recalc every time in the loop. Grab it once locally and use that instead.

Classes/Notifications/NotificationsViewController.swift
rnystrom: It's paged to 50 mixed-repo events, so number here likely wont be accurate if you have more than 50 unread notifications.

Classes/Notifications/NotificationSubscriptionsController.swift
BasThomas: Using a `State` enum inside of a `Controller` :)

Classes/Notifications/NotificationsViewController.swift
BasThomas: We \*could\* add the number for < 50 then maybe? 🤔

Classes/Notifications/NotificationsViewController.swift
BasThomas: Ah yep, didn't realise it's computed. 👍

PR #1299, opened 2017-12-20
Add Styles.Attributes enum for common NSAttributedString attributes

### LINKED ISSUE #856: Common Attributed Strings Attrs in Styles
We should have functions that take a `UIColor for common attributes that we have all over the place  so instead of calling:

```swift
attributes: [
  .font: Styles.Fonts.body,
  .foregroundColor: Styles.Colors.Gray.dark.color
]
```

We can call:

```swift
attributes: .hawkBody(color: Styles.Colors.Gray.dark.color)
```
<sub>i imagine that'd be implemented as an extension to `Dictionary`</sub

or: 

```swift
attributes: Styles.Attributes.body(color: Styles.Colors.Gray.dark.color)
```

spsjvc:
Fixes #856.

### DIFF
--- a/Classes/Views/Styles.swift
+++ b/Classes/Views/Styles.swift
@@ -166,3 +166,90 @@ extension String {
     }
 
 }
+
+// MARK: Attributes
+
+extension Styles {
+
+    enum Attributes {
+
+        static func body(color: UIColor) -> [NSAttributedStringKey: Any] {
+            return [
+                .font: Styles.Fonts.body,
+                .foregroundColor: color
+            ]
+        }
+
+        static func bodyBold(color: UIColor) -> [NSAttributedStringKey: Any] {
+            return [
+                .font: Styles.Fonts.bodyBold,
+                .foregroundColor: color
+            ]
+        }
+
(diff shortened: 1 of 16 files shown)


rnystrom:
One problem is that I’m about to replace a lot of the fonts with the new TextStyle type from the StyledText framework. The problem with fonts is they store a size, so dynamic content size changes won’t work (size has to change, this font has to change).

Classes/Views/Styles.swift
```diff
@@ -166,3 +166,90 @@ extension String {
...
 
 }
+
+// MARK: Attributes
+
+extension Styles {
+
+    enum Attributes {
+
+        static func body(color: UIColor) -> [NSAttributedStringKey: Any] {
```
BasThomas: Can we limit the `color` parameter to only accept `Styles.Colors`? 🤔

spsjvc:
@rnystrom Ok, so should I abandon this for now, until the fonts are replaced?

rnystrom:
Ya let's hold off on this until some of the foundational text refactors are wrapped up.

Issue #1297, opened 2017-12-20
Link and code should be blue


rnystrom:
![GitHawk Upload by rnystrom](https://i.imgur.com/E7MBVs1.jpg)

<details>
<summary>Bug Report Dump (Auto-generated)</summary>
<pre>
Version 1.17.0 (1513713275)
Device: iPhone X (iOS 11.2)
TestFlight: true
</pre>
</details>

SD10:
Is this resolved by #1302?

PR #1295, opened 2017-12-20
[DO NOT MERGE] Switch markAll logic back to mark all

BasThomas:
Fixes #1293 

Added the Do Not Merge as there is something super strange happening I can't explain - when refreshing, the mark all is enabled for a split second. Seems to happen in [`IGListAdapter`'s `performUpdatesAnimated`](https://github.com/Instagram/IGListKit/blob/master/Source/IGListAdapter.m#L330), called by the feed's `finishLoading`.

@rnystrom, what am I overlooking?

PR #1302, opened 2017-12-21
Fixing link color inside codeSpan

LucianoPAlmeida:
Hey guys, 
First time here :)
This PR is a fix to #1297, what was happening, in that case, was that the attribute for foreground color for code span was being applied after the link style, so the color in that range was being overridden.
There's also a fix for a wrong button being set to disable on notifications :)


rnystrom:
Aaaawesome! I’ll pr badly be merging #1295 for the button fix. Mind reverting that change?

LucianoPAlmeida:
@rnystrom Sure :) Done 👍

BasThomas:
This is cool, thanks @LucianoPAlmeida 🎉😀

Issue #1322, opened 2017-12-31
Scroll away Issue nav bar with sticky status and manage button


rnystrom:
Something like this:

https://github.com/telly/TLYShyNavBar/blob/master/README.md

Keeps the manage button present at all times.

Thoughts?

<details>
<summary>Bug Report Dump (Auto-generated)</summary>
<pre>
Version 1.17.0 (SEE BUILD PHASE)
Device: iPhone X (iOS 11.2)
TestFlight: false
</pre>
</details>

rnystrom:
cc @orta for issue management ideas

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

orta:
I wonder if it could also be fixed by having the same manage button  after the last comment also?

rnystrom:
I dig it! I'll experiment

PR #1327, opened 2018-01-02
Manage button scrolls to bottom to reveal actions

### LINKED ISSUE #1325: Hold and drag "Send" button for management options
Hard to explain in title, but spurred a bit on from the #1324 PR

Basically the "Send Comment" button at the bottom of the issue/PR, I wonder if we could do something when as a user I hold that?

UI wise I was thinking something stylistically like Quick Actions, but triggered by long press (though 3d touch would be slick too!)

However, something like this could also work if done correctly: https://github.com/bradjasper/RadialMenu

Options could be "Comment & Lock/Unlock", "Comment & Close/Open", "Just Comment". Opens door for more contextual management as well I suppose

@rnystrom Yay? Nay? ~~Fifty shades of Gray?~~

rnystrom:
Fixes #1325, #1322

- Tapping manage button now just scrolls to bottom
- Actions always displayed when you can admin
- Adjust layout math to evenly distribute options, no single action hanging

simulator screen shot - iphone x - 2018-01-01 at 23 49 24
simulator screen shot - iphone x - 2018-01-01 at 23 49 30
simulator screen shot - iphone x - 2018-01-01 at 23 49 31

### DIFF
--- a/Classes/Views/HangingChadItemWidth.swift
+++ b/Classes/Views/HangingChadItemWidth.swift
@@ -1,0 +1,36 @@
+//
+//  HangingChadItemWidth.swift
+//  Freetime
+//
+//  Created by Ryan Nystrom on 1/1/18.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import UIKit
+
+func HangingChadItemWidth(
+    index: Int,
+    count: Int,
+    containerWidth: CGFloat,
+    desiredItemWidth: CGFloat
+    ) -> CGFloat {
+    guard count > 0, index < count else { return 0 }
+
+    let maxPerRow = Int(floor(containerWidth / desiredItemWidth))
+    let itemsPerRow = min(count, maxPerRow)
+
+    let fullRows = count / itemsPerRow
+    let fullRowsCount = fullRows * itemsPerRow
+    let remainder = count - fullRowsCount
+
+    let itemsInThisRow: Int
+    if remainder == 1 && count > 2 && itemsPerRow > 1 && index > count - 3 {
+        itemsInThisRow = 2
+    } else if index > fullRowsCount - 1 {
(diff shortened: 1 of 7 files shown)


BasThomas:
Not liking this too much yet. 😐 

rnystrom:
@BasThomas got more feedback? Why not? What do you like more?

BasThomas:
- scrolls up above keyboard after commenting
- always in view (and taking up quite a bit of screen real estate) even though it’s only needed sometimes
- scroll to bottom from the manage button is unexpected / confusing (as afterwards the manage button is “gone”)
- now that it’s not in a clear border or anything (before it was between the button and title) it seems to be missing something / needing some sort of visual indicator

rnystrom:
Agree w/ showing under kb after commenting. I find that _really_ annoying.

I think what might make it better is when we do "scroll to bottom" after notifications or commenting, it should scroll to the bottom of the last timeline event, not the admin tools. That way you have to scroll them up a little in order to see them.

I'll admit its a bit of a different UI, but my gut tells me that w/ a little polish, this could be really delightful and fast after the behavior is learned.

Classes/Views/HangingChadItemWidth.swift
```diff
@@ -0,0 +1,47 @@
...
+//  HangingChadItemWidth.swift
+//  Freetime
+//
+//  Created by Ryan Nystrom on 1/1/18.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import UIKit
+
+func HangingChadItemWidth(
```
BasThomas: What is / was this for? Seems to be unused.

rnystrom: It is now, used to be for spacing the action buttons evenly. We can kill it.

PR #1333, opened 2018-01-03
Only show repos present in notification list

### LINKED ISSUE #1301: Inbox more options showing only own repos
Should show repos you’re contributing to, too. 

Not that useful atm. (At least for me) includes many unused repos and does not include repos I frequently contribute to / repos I recently contributed to.

<details>
<pre>
Version 1.17.0 (1513713275)
Device: iPhone X (iOS 11.2)
TestFlight: true
</pre>
</details>

rnystrom:
Fixes #1301


BasThomas:
This is also a bit tricky / confusing, eg. if there are no unreads, the ... menu will present an actionsheet with just “archived”. 😕 

I mean we can cache, but I think that won’t get rid of the trickyness.

PR #1353, opened 2018-01-05
Fetch issue/PR status and comment count for notifications

### LINKED ISSUE #1316: Fetch notification issue/PR status
Had a 💡 moment w/ @orta this morning. We can request _multiple issue/PR_ status info via the GraphQL API to batch notification status into a **single network request** using GQL aliases. An example request:

```
query {
  repository(owner: "rnystrom", name: "githawk") {
    issue(number:1) {
      title
      state
    }
    issue2: issue(number:2) {
      title
      state
    }
  }
}
```

Response:

```json
{
  "data": {
    "repository": {
      "issue": {
        "title": "Sample issue",
        "state": "CLOSED"
      },
      "issue2": {
        "title": "New MarkdownView",
        "state": "CLOSED"
      }
    }
  }
}
```

The trick to making this work is to:

1. Fetch notifications w/ V3 API
  - Mark all models as status "pending" or something, show as blue
2. Collect all issue/PR numbers
3. Dynamically build a GraphQL request
  - Do **not** use Apollo for this
4. Send request w/ token, manually parse
5. Map status and update models
6. Refresh UI
  - Updates should be done via `FlatCache`, so just reload feed with `.list` type changes

Really, **really** excited about this. Makes consuming notifications so much easier!

rnystrom:
Fixes #1316

simulator screen shot - iphone x - 2018-01-05 at 13 15 23


Classes/Notifications/NotificationClient.swift
```diff
@@ -69,14 +81,61 @@ final class NotificationClient {
...
+                    }
+                }
+
+                processedNotifications = updatedNotifications
+            } else {
+                processedNotifications = notifications
+            }
+            cache.set(values: processedNotifications)
+            completion(.success((processedNotifications, page)))
+        }
```
orta: yeah, this ended up like how I'd expect

Classes/Notifications/NotificationClient.swift
```diff
@@ -69,14 +81,61 @@ final class NotificationClient {
...
+        for notifications: [NotificationViewModel],
+        page: Int?,
+        completion: @escaping (Result<([NotificationViewModel], Int?)>) -> Void
+        ) {
+
+        let content = "state comments{totalCount}"
+        let notificationQueries: String = notifications.flatMap {
+            guard let alias = $0.stateAlias else { return nil }
+            return """
+            \(alias.key): repository(owner: "\($0.owner)", name: "\($0.repo)") { issueOrPullRequest(number: \(alias.number)) { ...on Issue {\(content)} ...on PullRequest {\(content)} } }
```
Sherlouk: Any reason why you chose to put this in a multiline comment, but then put it all on one line anyway?
Would prefer it if this/the content var to be actual multilines so it's easier to read!

Classes/Notifications/NotificationClient.swift
rnystrom: 😂

Classes/Notifications/NotificationClient.swift
rnystrom: Escaped quotes looked gross, and GQL has special requirements for newlines. Plus this was a hack that soon as it was working I shipped 😄

Feel free to PR with better formatting!

PR #1397, opened 2018-01-13
[DO NOT MERGE] Make the comment header cell accessible

### LINKED ISSUE #1243: VoiceOver should select all of comment header except "..."
Don't need to select image view, username, and date label separately. Squish into one AX view (maybe using custom rect?) and make it readable:

"rnystrom commented 3 minutes ago"

BasThomas:
Fixes #1243.

I couldn't find out how to make it so the `...` is also accessible, but wanted to PR anyway just for looks. @Sherlouk, any idea maybe?

### DIFF
--- a/Classes/Issues/Comments/Details/IssueCommentDetailCell.swift
+++ b/Classes/Issues/Comments/Details/IssueCommentDetailCell.swift
@@ -29,6 +29,12 @@ final class IssueCommentDetailCell: IssueCommentBaseCell, ListBindable {
 
     override init(frame: CGRect) {
         super.init(frame: frame)
+        isAccessibilityElement = true
+        accessibilityTraits |= UIAccessibilityTraitButton
+
+        addGestureRecognizer(UITapGestureRecognizer(
+            target: self,
+            action: #selector(onTapCell)))
 
         imageView.configureForAvatar()
         imageView.isUserInteractionEnabled = true
@@ -37,7 +43,7 @@ final class IssueCommentDetailCell: IssueCommentBaseCell, ListBindable {
         }
         imageView.addGestureRecognizer(UITapGestureRecognizer(
             target: self,
-            action: #selector(IssueCommentDetailCell.onTapAvatar))
+            action: #selector(onTapAvatar))
         )
         contentView.addSubview(imageView)
         imageView.snp.makeConstraints { make in
@@ -49,9 +55,10 @@ final class IssueCommentDetailCell: IssueCommentBaseCell, ListBindable {
         loginLabel.font = Styles.Fonts.title
         loginLabel.textColor = Styles.Colors.Gray.dark.color
         loginLabel.isUserInteractionEnabled = true
+        loginLabel.isAccessibilityElement = false
         loginLabel.addGestureRecognizer(UITapGestureRecognizer(
             target: self,
-            action: #selector(IssueCommentDetailCell.onTapLoginLabel))
+            action: #selector(onTapLoginLabel))
         )
         contentView.addSubview(loginLabel)
         loginLabel.snp.makeConstraints { make in
@@ -61,6 +68,7 @@ final class IssueCommentDetailCell: IssueCommentBaseCell, ListBindable {
 
         dateLabel.font = Styles.Fonts.secondary
         dateLabel.textColor = Styles.Colors.Gray.light.color
+        dateLabel.isAccessibilityElement = false
         contentView.addSubview(dateLabel)
         dateLabel.snp.makeConstraints { make in
             make.left.equalTo(loginLabel)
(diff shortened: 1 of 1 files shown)


Sherlouk:
Not sure how the view hierarchy is laid out, if it's nested in the same container view then we'd probably need to take it out of the container but keep constraints as they are - should keep it in the same location, but allow it to be essentially just on top thus allowed to use it

Also need to be careful to not block tapping on the header to open user profile

BasThomas:
> Also need to be careful to not block tapping on the header to open user profile

What do you mean with this one?

And yeah, probably. I thought it would just surface as the `...` is the topmost subview and is a button, but as far as I can tell (haven't tested on device), it's not discoverable.

Sherlouk:
Once a view is marked as an accessibility element, subviews are not discoverable either by tapping or by swiping through elements 😢 

I mean that if you're in VoiceOver, and I double tap on the header (select) then it should open the user's profile

BasThomas:
> I mean that if you're in VoiceOver, and I double tap on the header (select) then it should open the user's profile

https://github.com/rnystrom/GitHawk/pull/1397/files#diff-f852a51ab57af4d90983098a74165159R128 :)

Sherlouk:
ahh misread what that was doing my bad

BasThomas:
Tell me! How can I improve it?

Classes/Issues/Comments/Details/IssueCommentDetailCell.swift
```diff
@@ -127,19 +147,25 @@ final class IssueCommentDetailCell: IssueCommentBaseCell, ListBindable {
...
+            editedLabel.text = "\(Constants.Strings.bullet) \(editedString)"
 
             let detailFormat = NSLocalizedString("%@ edited this issue %@", comment: "")
             editedLabel.detailText = String(format: detailFormat, arguments: [editedLogin, editedDate.agoString])
         } else {
             editedLabel.isHidden = true
+            editedString = ""
         }
+
+        accessibilityLabel = NSLocalizedString("\(viewModel.login) commented \(viewModel.date.agoString). \(editedString)", comment: "")
```
Sherlouk: Not sure using a localized string is best here, there's no way we could translate every possible mutation of this -- Maybe if we use `String(format:` where the format is localizable?

Classes/Issues/Comments/Details/IssueCommentDetailCell.swift
```diff
@@ -114,6 +125,15 @@ final class IssueCommentDetailCell: IssueCommentBaseCell, ListBindable {
         delegate?.didTapProfile(cell: self)
     }
 
+    @objc func onTapCell() {
+        // Only enable the tap of the cell in VoiceOver;
+        // otherwise the avatar and label work as advertised.
+        // It seems like we need to add a recognizer to the cell
+        // to enable an action while using VoiceOver.
+        guard UIAccessibilityIsVoiceOverRunning() else { return }
+        delegate?.didTapProfile(cell: self)
```
Sherlouk: Personal preference for me but when it's such a trivial function like this I think it reads better to use an `if`. I personally feel it reads a bit easier!

Classes/Issues/Comments/Details/IssueCommentDetailCell.swift
```diff
@@ -83,6 +93,7 @@ final class IssueCommentDetailCell: IssueCommentBaseCell, ListBindable {
 
         editedLabel.font = Styles.Fonts.secondary
         editedLabel.textColor = Styles.Colors.Gray.light.color
+        editedLabel.isAccessibilityElement = false
```
Sherlouk: Do we need to set everything to false if the parent is set to true?

Classes/Issues/Comments/Details/IssueCommentDetailCell.swift
```diff
@@ -49,9 +55,10 @@ final class IssueCommentDetailCell: IssueCommentBaseCell, ListBindable {
         loginLabel.font = Styles.Fonts.title
         loginLabel.textColor = Styles.Colors.Gray.dark.color
         loginLabel.isUserInteractionEnabled = true
+        loginLabel.isAccessibilityElement = false
         loginLabel.addGestureRecognizer(UITapGestureRecognizer(
             target: self,
-            action: #selector(IssueCommentDetailCell.onTapLoginLabel))
+            action: #selector(onTapLoginLabel))
```
Sherlouk: 👍

Classes/Issues/Comments/Details/IssueCommentDetailCell.swift
BasThomas: Don't think so no, we can change this.

Classes/Issues/Comments/Details/IssueCommentDetailCell.swift
BasThomas: I don't mind; feel free to change it

Classes/Issues/Comments/Details/IssueCommentDetailCell.swift
BasThomas: Oops, of course.

rnystrom:
Cleaning up

BasThomas:
Yeah I’ll take this on again soon. 

PR #1402, opened 2018-01-13
Balance layout of new nav title view

rnystrom:


Classes/Views/NavigationTitleDropdownView.swift
```diff
@@ -46,7 +46,9 @@ final class NavigationTitleDropdownView: UIControl {
         }
 
         label.snp.makeConstraints { make in
-            make.center.equalTo(self)
+//            make.center.equalTo(self)
```
BasThomas: Can this be removed?

rnystrom: oopsy

PR #1403, opened 2018-01-13
Notifications VC back item empty

rnystrom:
Consistency w/ rest of the app


BasThomas:
Can we put this in a superclass maybe? Or another way to make this more type safe / less forgettable? 😁

rnystrom:
@BasThomas open to ideas. We don't really have a concept of a "base VC" yet (and I personally think that might be a slippery slope). But def open.

BasThomas:
Yeah, not too big of a fan of those. Something else would be preferred.

PR #1424, opened 2018-01-16
Fix bad link for profiles

rnystrom:
From #1411 cc @BasThomas


BasThomas:
Oops, thanks for fixing 👍

Issue #1339, opened 2018-01-04
Comments render white space


BasThomas:
![GitHawk Upload by BasThomas](https://i.imgur.com/A317Rou.jpg)

Happens on both iPad and iPhone. Have a feeling it’s been happening more the last few days? Also happened on previous version though.

<details>
<summary>Bug Report Dump (Auto-generated)</summary>
<pre>
Version 1.17.0 (1514903367)
Device: iPad7,3 (iOS 11.2.1)
TestFlight: true
</pre>
</details>

BasThomas:
Issue in question: https://github.com/Moya/Moya/issues/1523

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rnystrom:
Hmm @BasThomas can you do some debugging here? I can't repro.

SD10:
The spacing is even larger in the latest TestFlight: 1.17.0

![GitHawk Upload by SD10](https://i.imgur.com/X3nLocR.jpg)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BasThomas:
Haven’t seen it lately - can you tap with three fingers to dig a bit with FLEX?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

SD10:
Sorry @BasThomas I don’t know what you mean.

It’s difficult to reproduce but once it happens, any comments have an excessive amount of spacing on any issue throughout the app. Once I kill GitHawk and reopen it’s gone.

It has happened by viewing an issue on a fresh app start, expanding a comment, posting a comment. Can’t single it down to one specific action and it’s inconsistent.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BasThomas:
In Testflight, you can tap with three fingers anywhere on the screen to have additional tools (see image), which enable you to debug the views, network requests and more. 

![GitHawk Upload by BasThomas](https://i.imgur.com/QiUavQX.jpg)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

SD10:
Sorry @BasThomas, noob moment. Thanks for getting me up to speed!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BasThomas:
@SD10 have you been seeing it in the last few days / been able to debug with FLEX?

SD10:
[...]

PR #1443, opened 2018-01-18
Remove KVO and just ask the webview how tall it is

### LINKED ISSUE #1441: Crash from navigating from issue to inbox
App crashes after viewing MessageKit #449 and returning to inbox

Debug notes:
- User defaults for 449 is empty
- If I interrupt the view from scrolling to bottom when viewing the issue, it doesn’t crash on return to inbox
- After interrupting the scrolling, extra white space is rendered on GitHawk comments. The issue no longer crashes while white space is present

I think this is related to scrolling or the content size. I can prevent the crash from scrolling to the top of the issue before exiting. This issue only crashes on return to inbox. Viewing it from the repo page works fine. I noticed that entering from the repo page doesn’t scroll to bottom automatically as well. 

rnystrom:
Fixes #1441, should help w/ #1339

### DIFF
--- a/Classes/Issues/Comments/Html/IssueCommentHtmlCell.swift
+++ b/Classes/Issues/Comments/Html/IssueCommentHtmlCell.swift
@@ -155,6 +155,7 @@ final class IssueCommentHtmlCell: IssueCommentBaseCell, ListBindable, UIWebViewD
     }
 
     func webViewDidFinishLoad(_ webView: UIWebView) {
+        // if the cell is hidden, its been put back in the reuse pool
         guard self.isHidden == false,
             let contentHeight = webView.stringByEvaluatingJavaScript(from: "document.body.offsetHeight") as NSString?
             else { return }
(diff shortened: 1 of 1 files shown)


Classes/Issues/Comments/Html/IssueCommentHtmlCell.swift
```diff
@@ -161,20 +155,12 @@ final class IssueCommentHtmlCell: IssueCommentBaseCell, ListBindable, UIWebViewD
     }
 
     func webViewDidFinishLoad(_ webView: UIWebView) {
-        webView.isHidden = false
-    }
-
-    // MARK: KVO
+        guard self.isHidden == false,
+            let contentHeight = webView.stringByEvaluatingJavaScript(from: "document.body.offsetHeight") as NSString?
```
rnystrom: Took some experimenting, but this is the right way to ask

Sherlouk:
Does this still work with HTML5 details?

<details>
<summary>Oooh  Collapsible Content</summary>
here's a gif


</details>

rnystrom:
@Sherlouk good call out, it probably won’t L.

rnystrom:
Confirmed it doesn’t 

PR #1447, opened 2018-01-19
Update HTML cell height on content height change

### LINKED ISSUE #1298: Add tap handler to details
Same as image tapping. On tap, resize the web view and forward to the delegate. This should actually collapse details when closed.

<details>
<pre>
Version 1.17.0 (1513713275)
Device: iPhone X (iOS 11.2)
TestFlight: true
</pre>
</details>

### LINKED ISSUE #1387: The <details> tag does not render properly
Please refer to the description of the issue https://github.com/fastlane/fastlane/issues/11470 on desktop and in GitHawk app. See the difference.

<details>
<pre>
Version 1.16.0 (1513134612)
Device: iPhone X (iOS 11.2.1)
TestFlight: false
</pre>
</details>

### LINKED ISSUE #1446: HTML5 <details> no longer works
Follow up from https://github.com/rnystrom/GitHawk/pull/1443#issuecomment-358720680

Obviously we don't want to necessarily reintroduce the crash we had in the first place so need to tread carefully on this one!

rnystrom:
Fixes #1446
Fixes #1387
Fixes #1298

### DIFF
--- a/Classes/Issues/Comments/Html/IssueCommentHtmlCell.swift
+++ b/Classes/Issues/Comments/Html/IssueCommentHtmlCell.swift
@@ -32,8 +32,10 @@ private final class IssueCommentHtmlCellWebView: UIWebView {
 final class IssueCommentHtmlCell: IssueCommentBaseCell, ListBindable, UIWebViewDelegate {
 
     private static let ImgScheme = "freetime-img"
+    private static let HeightScheme = "freetime-hgt"
     private static let htmlHead = """
     <!DOCTYPE html><html><head><style>
+    * {margin: 0;padding: 0;}
     body{
     // html whitelist: https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/sanitization_filter.rb#L45-L49
     // lint compiled style with http://csslint.net/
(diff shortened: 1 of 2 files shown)


BasThomas:
You actually need the keyword for every issue #, otherwise they won’t all close. I’ve changed it. 

rnystrom:
@BasThomas TIL, thanks!

Classes/Issues/Comments/Html/IssueCommentHtmlCell.swift
```diff
@@ -77,6 +79,23 @@ final class IssueCommentHtmlCell: IssueCommentBaseCell, ListBindable, UIWebViewD
...
+            (function run() {
+                newHeight = elm.offsetHeight;
+                if(lastHeight != newHeight) {
+                    callback(newHeight);
+                }
+                lastHeight = newHeight;
+                if(elm.onElementHeightChangeTimer) {
+                    clearTimeout(elm.onElementHeightChangeTimer);
+                }
+                elm.onElementHeightChangeTimer = setTimeout(run, 300);
```
rnystrom: This hack is taken from [here](https://stackoverflow.com/a/14901150/940936). There are no DOM events to respond to content size changes. Instead we poll every 300ms and see if the height changed. On change, we try to change the `document.location` (url) to `freetime-hgt://300` where 300 is the new height.

Pretty jank, but this seems to be the right way to handle it.

Another blog post idea!

BasThomas:
Yeah it's a pretty annoying limitation; at least [they documented it](https://help.github.com/articles/closing-issues-using-keywords/#closing-multiple-issues)!

PR #1508, opened 2018-02-10
Update README.md

orta:
Took some notes when setting up

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,19 @@ GitHawk will be released with (at least) a three-phased rollout:
 
 ## Installation
 
-If you want to build GitHawk locally, run `npm install`. It will install [apollo-codegen](https://github.com/apollographql/apollo-codegen) that is required to generate the
-GraphQL models.
+To get started, you will need to install node (`brew install node`), then run these instructions:
+
+```sh
+git clone https://github.com/rnystrom/GitHawk.git
+cd GitHawk
+bundle
+bundle exec pod install
+npm install
+open Freetime.xcworkspace/
+```
+
+Next up you will need to set up a [Client ID and Client Secret](Setup.md) for GitHub.
+
 
 ## Open Source & Copying
 


orta:
Looks like it's installing apollo client globally, you don't need the package.json nor the `npm install`

screen shot 2018-02-10 at 7 59 04 am

Sherlouk:
If I remember correctly, the reason we use `npm install` rather than depending explicitly on the Apollo script is to ensure that everybody is using the same version as minor differences in version means the generated files are completely different (which pads out, unnecessarily, the pull requests)

Definitely not a necessity though

orta:
The script in the build stage doesn't actually use that though, it's tied to the version defined in the Pod itself.

`/Pods/Apollo/scripts/check-and-run-apollo-codegen.sh`:

```sh
# Only major and minor version should be specified here
REQUIRED_APOLLO_CODEGEN_VERSION=0.17


install_apollo_codegen() {
  # Exit immediately if the command fails
  set -e
  npm install -g apollo-codegen@$REQUIRED_APOLLO_CODEGEN_VERSION
  set +e
}

[...]

if ! are_versions_compatible $INSTALLED_APOLLO_CODEGEN_VERSION $REQUIRED_APOLLO_CODEGEN_VERSION; then
  echo "The version of Apollo.framework in your project requires apollo-codegen $REQUIRED_APOLLO_CODEGEN_VERSION, \
but $INSTALLED_APOLLO_CODEGEN_VERSION seems to be installed. Installing..."
  install_apollo_codegen
fi
```

It will just reinstall a global version and ignore your local package completely.

Sherlouk:
My interpretation (albeit maybe wrong) is that `are_versions_compatible` only checks the major/minor version as opposed to the entire version. `npm install` uses `0.17.0-alpha.13` which would match the version in the pod `0.17` and therefore would not install a different version

Sherlouk:
Makes sense, thanks for explaining!

PR #1515, opened 2018-02-13
Display paste menu on URL wrapping

### LINKED ISSUE #1453: Open "Paste" menu control after inserting a Markdown URL
- Select some text in the text view
- Tap the "link" icon
- Surrounds the text with `[selection](|)` where the pipe is where the cursor ends up
- Would be **awesome** to show the paste menu control so you can just paste in the link

SD10:
Resolves #1453 

This PR does the following:
- Automatically displays the paste menu upon using the link markdown action if the `UIPasteboard` contains a `URL`
- Refactors the order of the `switch` statement to check more frequent `Operation`s first

If this is too much of a hack I can try to create an indirect enum that allows both a `.wrap` and `.execute` but why get fancy until it's necessary?

### DIFF
--- a/Classes/Views/TextActionsController.swift
+++ b/Classes/Views/TextActionsController.swift
@@ -29,19 +29,21 @@ final class TextActionsController: NSObject,
     // MARK: IssueTextActionsViewDelegate
 
     func didSelect(actionsView: IssueTextActionsView, operation: IssueTextActionOperation) {
-        switch operation.operation {
-        case .wrap(let left, let right):
-            textView?.replace(left: left, right: right, atLineStart: false)
-            if operation.name == "Wrap text as URL" && UIPasteboard.general.hasURLs {
-                UIMenuController.shared.setMenuVisible(true, animated: trueUnlessReduceMotionEnabled)
+        func handle(_ operation: IssueTextActionOperation.Operation) {
+            switch operation {
+            case .wrap(let left, let right):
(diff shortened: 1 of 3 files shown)


Classes/Views/TextActionsController.swift
```diff
@@ -30,10 +30,17 @@ final class TextActionsController: NSObject,
 
     func didSelect(actionsView: IssueTextActionsView, operation: IssueTextActionOperation) {
         switch operation.operation {
-        case .execute(let block): block()
-        case .wrap(let left, let right): textView?.replace(left: left, right: right, atLineStart: false)
-        case .line(let left): textView?.replace(left: left, right: nil, atLineStart: true)
-        case .uploadImage: displayUploadImage()
+        case .wrap(let left, let right):
+            textView?.replace(left: left, right: right, atLineStart: false)
+            if operation.name == "Wrap text as URL" && UIPasteboard.general.hasURLs {
```
BasThomas: The operation name checking... 😕
This would break if we end up supporting a different language. We could at least put the `NSLocalizedString` in the Constants and use that in both places?

SD10:
Good thoughts @BasThomas 👍 How do you like the approach in 52aed83? I add a `.multi` operation and resolve it in order recursively. No need to lean on any sort of `String`.

rnystrom:
I dig it! Pretty clever. Really excited for this.

PR #1528, opened 2018-02-16
Update StyledText to fix loop bug

rnystrom:


BasThomas:
What bug was this if I may ask?

rnystrom:
Caught it doing something else, but the LRUCache can infinitely loop on compaction bc the tail node wasn’t clipped. Doesn’t seem to be happening in prod tho.

PR #1533, opened 2018-02-18
Add support for pending merges

### LINKED ISSUE #1531: Incorrect merge status
img_1866
img_c94a6af49341-1

Says some checks failed when in fact all are in progress

As seen in #1530

BasThomas:
Fixes #1531

### DIFF
--- a/Classes/Issues/Merge/IssueMergeSectionController.swift
+++ b/Classes/Issues/Merge/IssueMergeSectionController.swift
@@ -68,11 +68,12 @@ MergeButtonDelegate {
             let states = object.contexts.map { $0.state }
             let state: IssueMergeSummaryModel.State
             let stateDescription: String
+            let failureDescription = NSLocalizedString("Some checks failed", comment: "")
             let containsFailure = states.contains(.failure) || states.contains(.error)
             switch states {
             case _ where containsFailure:
                 state = .failure
-                stateDescription = NSLocalizedString("Some checks failed", comment: "")
+                stateDescription = failureDescription
             case let states where states.contains(.pending) && !containsFailure:
                 state = .pending
                 stateDescription = NSLocalizedString("Merge status pending", comment: "")
@@ -80,7 +81,9 @@ MergeButtonDelegate {
                 state = .success
                 stateDescription = NSLocalizedString("All checks passed", comment: "")
             default:
-                fatalError()
+                assert(false, "This should only occur when any of the `states` are of type `.expected`, which we have no clue of when it is used. The documentation (https://developer.github.com/v4/enum/statusstate/) doesn't answer that question either.")
+                state = .failure
+                stateDescription = failureDescription
             }
             
             viewModels.append(IssueMergeSummaryModel(title: stateDescription, state: state))
--- a/Classes/Issues/Merge/IssueMergeSectionController.swift
+++ b/Classes/Issues/Merge/IssueMergeSectionController.swift
@@ -65,13 +65,25 @@ MergeButtonDelegate {
         var viewModels = [ListDiffable]()
 
         if object.contexts.count > 0 {
-            let success = object.contexts.reduce(true, { $0 && $1.state == .success })
-            viewModels.append(IssueMergeSummaryModel(
-                title: success ?
-                    NSLocalizedString("All checks passed", comment: "")
-                    : NSLocalizedString("Some checks failed", comment: ""),
-                state: success ? .success : .failure
-            ))
+            let states = object.contexts.map { $0.state }
+            let state: IssueMergeSummaryModel.State
+            let stateDescription: String
+            let containsFailure = states.contains(.failure) || states.contains(.error)
+            switch states {
+            case _ where containsFailure:
+                state = .failure
+                stateDescription = NSLocalizedString("Some checks failed", comment: "")
+            case let states where states.contains(.pending) && !containsFailure:
+                state = .pending
+                stateDescription = NSLocalizedString("Merge status pending", comment: "")
+            case let states where states.reduce(true, { $0 && $1 == .success }):
+                state = .success
+                stateDescription = NSLocalizedString("All checks passed", comment: "")
+            default:
+                fatalError()
+            }
+            
+            viewModels.append(IssueMergeSummaryModel(title: stateDescription, state: state))
         }
 
         viewModels += object.contexts as [ListDiffable]
--- a/Classes/Issues/Merge/IssueMergeSectionController.swift
+++ b/Classes/Issues/Merge/IssueMergeSectionController.swift
@@ -66,24 +66,7 @@ MergeButtonDelegate {
 
         if object.contexts.count > 0 {
             let states = object.contexts.map { $0.state }
-            let state: IssueMergeSummaryModel.State
-            let stateDescription: String
-            let failureDescription = NSLocalizedString("Some checks failed", comment: "")
-            switch states {
-            case let states where states.contains(.failure) || states.contains(.error):
-                state = .failure
-                stateDescription = failureDescription
-            case let states where states.contains(.pending):
-                state = .pending
-                stateDescription = NSLocalizedString("Merge status pending", comment: "")
-            case let states where states.reduce(true, { $0 && $1 == .success }):
-                state = .success
-                stateDescription = NSLocalizedString("All checks passed", comment: "")
-            default:
-                assert(false, "This should only occur when any of the `states` are of type `.expected`, which we have no clue of when it is used. The documentation (https://developer.github.com/v4/enum/statusstate/) doesn't answer that question either.")
-                state = .failure
-                stateDescription = failureDescription
-            }
+            let (state, stateDescription) = combinedMergeStatus(for: states)
             
             viewModels.append(IssueMergeSummaryModel(title: stateDescription, state: state))
         }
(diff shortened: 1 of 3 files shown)


Classes/Issues/Merge/IssueMergeSectionController.swift
```diff
@@ -65,13 +65,25 @@ MergeButtonDelegate {
...
+                state = .failure
+                stateDescription = NSLocalizedString("Some checks failed", comment: "")
+            case let states where states.contains(.pending) && !containsFailure:
+                state = .pending
+                stateDescription = NSLocalizedString("Merge status pending", comment: "")
+            case let states where states.reduce(true, { $0 && $1 == .success }):
+                state = .success
+                stateDescription = NSLocalizedString("All checks passed", comment: "")
+            default:
+                fatalError()
```
BasThomas: ~Not too sure if we want to do this; maybe at least add a message?~ Writing this and realizing we can get here if that [expected state](https://developer.github.com/v4/enum/statusstate/) is a thing. But even then, how do we recover?

Classes/Issues/Merge/IssueMergeSectionController.swift
```diff
@@ -65,13 +65,25 @@ MergeButtonDelegate {
...
+            let state: IssueMergeSummaryModel.State
+            let stateDescription: String
+            let containsFailure = states.contains(.failure) || states.contains(.error)
+            switch states {
+            case _ where containsFailure:
+                state = .failure
+                stateDescription = NSLocalizedString("Some checks failed", comment: "")
+            case let states where states.contains(.pending) && !containsFailure:
+                state = .pending
+                stateDescription = NSLocalizedString("Merge status pending", comment: "")
```
BasThomas: GitHub uses `[waiting to hear about {commit}](https://user-images.githubusercontent.com/4190298/36353428-a37fb9cc-14c6-11e8-9a04-91fdc9f95f10.jpeg)`, but couldn't find any last-commit info so changed it to this.

Classes/Issues/Merge/IssueMergeSectionController.swift
```diff
@@ -65,13 +65,25 @@ MergeButtonDelegate {
...
 
         if object.contexts.count > 0 {
-            let success = object.contexts.reduce(true, { $0 && $1.state == .success })
-            viewModels.append(IssueMergeSummaryModel(
-                title: success ?
-                    NSLocalizedString("All checks passed", comment: "")
-                    : NSLocalizedString("Some checks failed", comment: ""),
-                state: success ? .success : .failure
-            ))
+            let states = object.contexts.map { $0.state }
```
BasThomas: As this function is a bit bulky at the moment, I wasn't sure if we would want to pull this out. Thoughts?

Classes/Issues/Merge/IssueMergeSectionController.swift
BasThomas: Maybe take the `.failure` path and assert for debug builds?

Classes/Issues/Merge/IssueMergeSectionController.swift
```diff
@@ -65,13 +65,25 @@ MergeButtonDelegate {
...
+            let stateDescription: String
+            let containsFailure = states.contains(.failure) || states.contains(.error)
+            switch states {
+            case _ where containsFailure:
+                state = .failure
+                stateDescription = NSLocalizedString("Some checks failed", comment: "")
+            case let states where states.contains(.pending) && !containsFailure:
+                state = .pending
+                stateDescription = NSLocalizedString("Merge status pending", comment: "")
+            case let states where states.reduce(true, { $0 && $1 == .success }):
```
Sherlouk: This line blew my mind for just a few seconds, took a bit to work out what was happening, could it be made simpler at all do you think? (Obviously just a personal thing)

BasThomas: It hasn't changed from [what it was before](https://github.com/rnystrom/GitHawk/pull/1533/files#diff-7c180300a62c299e2c10544bd45d9223L68), but I do agree its readability could be improved. We can add a `allSucceeded` bool at the top, just like [`containsFailure`](https://github.com/rnystrom/GitHawk/pull/1533/files#diff-7c180300a62c299e2c10544bd45d9223R71) (at a super tiny performance penalty for always executing that `reduce`)?

BasThomas: And / or we (re)pitch [this proposal](https://github.com/apple/swift-evolution/pull/664) that adds `all` to `Sequence`? 😀

Sherlouk: Fair point about it not changing, not saying you have to change it though!

I wonder if something like `states.contains(where: { $0 != .success })` or similar (I think that's a thing...)

Or yea as you say, move a var to the top!

Also that all syntax looks really interesting 🤔 Would be super helpful actually

BasThomas:
> I wonder if something like `states.contains(where: { $0 != .success })` or similar (I think that's a thing...)

Yeah but then you would need [`!states.contains(where: { $0 != .success })`](https://github.com/apple/swift-evolution/pull/664/files#diff-b5c7d543f27e9ba8db174753956310eeR21) (notice the `!` in front), which is even less readable (unless you don't know `reduce`, I suppose).

Classes/Issues/Merge/IssueMergeSectionController.swift
BasThomas: > I wonder if something like `states.contains(where: { $0 != .success })` or similar (I think that's a thing...)

Yeah but then you would need [`!states.contains(where: { $0 != .success })`](https://github.com/apple/swift-evolution/pull/664/files#diff-b5c7d543f27e9ba8db174753956310eeR21) (notice the `!` in front), which is even less readable (unless you don't know `reduce`, I suppose).

Sherlouk: Super fair comment, I'll leave it to you to pick the solution with the least devils - if that means keeping as is I'm fine with that!

Classes/Issues/Merge/IssueMergeSectionController.swift
```diff
@@ -65,13 +65,28 @@ MergeButtonDelegate {
...
-                    NSLocalizedString("All checks passed", comment: "")
-                    : NSLocalizedString("Some checks failed", comment: ""),
-                state: success ? .success : .failure
-            ))
+            let states = object.contexts.map { $0.state }
+            let state: IssueMergeSummaryModel.State
+            let stateDescription: String
+            let failureDescription = NSLocalizedString("Some checks failed", comment: "")
+            let containsFailure = states.contains(.failure) || states.contains(.error)
+            switch states {
```
SD10: Can this entire thing become a `switch true`? Looks like we're not using `states`?

Classes/Issues/Merge/IssueMergeSectionController.swift
```diff
@@ -65,13 +65,28 @@ MergeButtonDelegate {
...
+            case _ where containsFailure:
+                state = .failure
+                stateDescription = failureDescription
+            case let states where states.contains(.pending) && !containsFailure:
```
SD10: `!containsFailure` will get hit by the previous case

Classes/Issues/Merge/IssueMergeSectionController.swift
BasThomas: Don't think so; the `case let states where ...` uses it.

Classes/Issues/Merge/IssueMergeSectionController.swift
BasThomas: That's the inverse (`containsFailure`) right? Or am I missing something?

BasThomas: Good catch!

Classes/Issues/Merge/IssueMergeSectionController.swift
SD10: Ah, what I meant is we're not really using `states` here. It's available out of the scope of the `switch` statement. Every case has a `where` clause. So this means what we really care about is a `Bool` and not `states`.
```Swift
switch true {
case containsFailure: //
case states.contains(.pending): //
case states.reduce(true, { $0 && $1 == .success }): //
default: //
}
```
Sorry to be "that guy" 😓

BasThomas: Oh wow, I didn't know about that! That is... interesting, to say the least. That would kind of be more readable, huh? 😮

SD10: Yeah, I do think it is more clear in its intent. I just love `switch true` 🤓

Classes/Issues/Merge/IssueMergeSectionController.swift
rnystrom: Fine w/ this. Less text is better since we have less screen to work w/, and the `UILabel` is limited to one line w/ truncation.

Classes/Issues/Merge/IssueMergeSectionController.swift
rnystrom: Can someone explain why using `switch true {...}` would be better than just `if` statements here?

Classes/Issues/Merge/IssueMergeSectionController.swift
rnystrom: It's only used here, so not the end of the world. But maybe at the very least we pull it all out into a function? Bonus points if we can find a way to make this stateless inputs/outputs and unit test it!

Classes/Issues/Merge/IssueMergeSectionController.swift
BasThomas: Other than aesthetics, I can’t.

rnystrom:
FYI this is the last PR I think we need before sending off 1.18. Once we merge I’ll give it a few days to soak then we’re off! I’ll clean up Milestones.

Classes/Issues/Merge/IssueMergeSectionController.swift
SD10: No reason other than I prefer how `switch` looks. Ironically, did some digging last night into `switch` vs `if-else` and there is some talk that the compiler can better optimize for a switch. However, even if true, this would be a micro-optimization that has _no real value_ in production. 

Other than that, the switch is capable of using `fallthrough`, which we don't need here. So it is all a matter of preference.

Classes/Issues/Merge/IssueMergeSectionController.swift
BasThomas: b1cb3c2

rnystrom:
@Sherlouk + @SD10 + @BasThomas feeling good on this?

BasThomas:
I am!

PR #1536, opened 2018-02-19
Check if default value is for client ID and secret is set

### LINKED ISSUE #1535: App crashes when attempting to login
## Description
When tapping the login button the app crashes.

Reason: Client ID is read from the environment and it's force unwrapped: 
```swift
private let loginURL = URL(string: "http://github.com/login/oauth/authorize?client_id=\(Secrets.GitHub.clientId)&scope=user+repo+notifications")!
```

jamuc:
Fixes #1535 

## What this does
Removes the pre-configured GitHub client ID and secret. This has the advantage that if the user doesn't have them set in their environment an explicit error message will be displayed.

### DIFF
--- a/Freetime.xcodeproj/xcshareddata/xcschemes/Freetime-AppCenter.xcscheme
+++ b/Freetime.xcodeproj/xcshareddata/xcschemes/Freetime-AppCenter.xcscheme
@@ -81,23 +81,8 @@
             isEnabled = "NO">
          </EnvironmentVariable>
          <EnvironmentVariable
-            key = "GITHUB_CLIENT_SECRET"
-            value = "SET ON CI"
-            isEnabled = "YES">
-         </EnvironmentVariable>
-         <EnvironmentVariable
             key = "NETWORK_RECORD_PATH"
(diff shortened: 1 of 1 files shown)


Sherlouk:
I feel like we just shouldn't have the variables set in the shared scheme, if CI genuinely does define them then it's just not important for them to exist with dummy values causing this issue!

BasThomas:
Let's wait for @rnystrom to see why he set it up the way he did :)

Sherlouk:
Agree @BasThomas, at bare minimum if we **must** keep them for what ever quirky reason they should probably just be a space or something to make the empty comparison easier

rnystrom:
Ya lets just remove them from shared. I just duplicated my local one and needed to strip the values.

jamuc:
Thanks for the feedback. Done.

rnystrom:
@jamuc mind removing the Imgur key too?

jamuc:
@rnystrom Not at all. Was thinking about doing that anyway. Updated the PR.

PR #1538, opened 2018-02-19
Remove optimizations from debug build

rnystrom:
Noticed during debugging that I had fast optimizations enabled for debug builds. Following [this blog post](https://jobs.zalando.com/tech/blog/improving-swift-compilation-times-from-12-to-2-minutes/?gh_src=4n3gxh1) and going back to `-0None` but still enabling whole-module optimizations.

cc @Sherlouk who's been doing some build optimizations work recently


Sherlouk:
You mean, someone reads my tweets 😱 

That blog and these changes back to work I've been doing, though my work has been primarily focusses on making builds themselves build projects faster rather than reconfiguring projects 😉 (Surprising difference!)

PR #1546, opened 2018-02-20
Release notifications

rnystrom:
Only thing I can't quite figure out is how to link to it... We get an API path that we can extract the id from, like this:

https://api.github.com/repos/facebook/Docusaurus/releases/9721470

Where `9721470` is the ID. But I'm not sure how to link it to [the actual release](https://github.com/facebook/Docusaurus/releases/tag/v1.0.7).

Any ideas?

Shot of it working:

simulator screen shot - iphone x - 2018-02-19 at 22 32 26

### DIFF
--- a/Classes/Bookmark/BookmarkViewModel.swift
+++ b/Classes/Bookmark/BookmarkViewModel.swift
@@ -36,6 +36,9 @@ final class BookmarkViewModel: ListDiffable {
                 ]
             ))
             attributedText = mAttributedText
+        case .release:
+            // TODO does this even exist?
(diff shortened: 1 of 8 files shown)


rizwankce:
if we can get the tag name, then we can do something like below

`https://github.com/facebook/Docusaurus/releases/tag/v1.0.7` to link to it

Sherlouk:
Totally misread this as you wanted to do push notifications for releases 🤦‍♂️

rnystrom:
@rizwankce nice! Tag name exists, we can use that.

rnystrom:
I’m just gonna cheat and do an extra network request to get the tag name.

PR #1549, opened 2018-02-20
Move API stringiness to client

BasThomas:
This moves some String-dependant URLs to the GithubClient, making it easier to support Enterprise at some point.

### DIFF
--- a/Classes/Issues/Comments/IssueCommentSectionController.swift
+++ b/Classes/Issues/Comments/IssueCommentSectionController.swift
@@ -70,7 +70,7 @@ IssueCommentDoubleTapDelegate {
     func shareAction(sender: UIView) -> UIAlertAction? {
         let attribute = object?.asReviewComment == true ? "#discussion_r" : "#issuecomment-"
         guard let number = object?.number,
-            let url = URL(string: "https://github.com/\(model.owner)/\(model.repo)/issues/\(model.number)\(attribute)\(number)")
+            let url = GithubClient.url(path: "\(model.owner)/\(model.repo)/issues/\(model.number)\(attribute)\(number)")
         else { return nil }
         weak var weakSelf = self
 
--- a/Classes/Issues/Files/GithubClient+PullRequestFiles.swift
+++ b/Classes/Issues/Files/GithubClient+PullRequestFiles.swift
@@ -17,7 +17,7 @@ extension GithubClient {
         page: Int,
         completion: @escaping (Result<([File], Int?)>) -> Void
         ) {
-        request(Request(
+        request(Request.api(
             path: "repos/\(owner)/\(repo)/pulls/\(number)/files",
             parameters: [
                 "per_page": 50,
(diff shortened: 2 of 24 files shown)


Classes/Issues/Files/GithubClient+PullRequestFiles.swift
```diff
@@ -37,40 +37,4 @@ extension GithubClient {
                 }
         }))
     }
-
```
BasThomas: Removed as this was apparently unused?

Classes/Systems/GithubClient.swift
```diff
@@ -28,16 +28,34 @@ struct GithubClient {
         let logoutOnAuthFailure: Bool
         let completion: (DataResponse<Any>, Page?) -> Void
 
-        init(
+        static func api(
```
BasThomas: Not too sure about this approach with `static func`s. Would it be more readable in another form, like eg. `APIRequest.init`? Or something else entirely? Would love some feedback!

Classes/Systems/GithubClient.swift
```diff
@@ -46,7 +64,24 @@ struct GithubClient {
             )
         }
 
-        init(
+        static func status(
+            method: HTTPMethod = .get,
```
BasThomas: Don't know if we need all the default arguments here as parameters, but just let them be as they don't harm anything really (unless someone misinterprets this and changes them, of course).

Classes/Systems/GithubClient.swift
```diff
@@ -164,4 +199,8 @@ struct GithubClient {
         })
     }
 
+    static func url(baseURL: String = "https://github.com/", path: String) -> URL? {
```
BasThomas: So most client-independent URLs have `https://github.com/` as their baseURL, but some [have others](https://github.com/rnystrom/GitHawk/pull/1549/files#diff-9f0bae268086ea5dfc986d1a77f28abaR114) (`https://status.github.com/` in this case). Would it be overkill to put those variations in an enum?

Classes/Systems/GithubClient.swift
```diff
@@ -164,4 +210,43 @@ struct GithubClient {
...
+        let baseURL: String
+        if let enterpriseURL = userSession?.enterpriseURL {
+            switch request.type {
+            case .api:
+                baseURL = "\(enterpriseURL)api/v3"
+            case .site:
+                baseURL = enterpriseURL
+            case .status:
+                assert(false, "Status checking is unsupported in GitHub Enterprise")
+                baseURL = "invalid-url"
```
BasThomas: Not sure how we would want to make this easier to bubble up? Maybe we can handle with a Toast somewhere?

BasThomas:
@Sherlouk @rnystrom would love your opinion on this!

Classes/Systems/GithubClient.swift
jamuc: Let's have a chat about this in the office :)

Classes/Issues/Comments/IssueCommentSectionController.swift
```diff
@@ -70,7 +70,7 @@ IssueCommentDoubleTapDelegate {
     func shareAction(sender: UIView) -> UIAlertAction? {
         let attribute = object?.asReviewComment == true ? "#discussion_r" : "#issuecomment-"
         guard let number = object?.number,
-            let url = URL(string: "https://github.com/\(model.owner)/\(model.repo)/issues/\(model.number)\(attribute)\(number)")
```
jamuc: One small remark: You've setup the base url to have a trailing slash. I don't think you can expect users to enter a host name with a traling slash. I would advise to go without the trailing slash syntax.

BasThomas: We should definitely handle that somewhere and not bother the user with it. Either we go with or without, and then check the entered URL and have it conform to what we choose.

Classes/Systems/GithubClient.swift
BasThomas: Yep, we have to get rid of this when / if we pull out `Request`. Then we need another way of injecting the baseURL, but good catch regardless.

Classes/Systems/GithubClient.swift
BasThomas: Yeah - currently we are using [`https://status.github.com/`](https://github.com/rnystrom/GitHawk/pull/1549/files#diff-9f0bae268086ea5dfc986d1a77f28abaR114) and [`https://api.github.com/`](https://github.com/rnystrom/GitHawk/pull/1549/files#diff-2af355000c118286c09e081675a7b959R36).

rnystrom:
I busted a lot w/ the networking refactor 😨

BasThomas:
I'll take a look 👍

PR #1550, opened 2018-02-20
Remove image upload description

### LINKED ISSUE #1527: Title and description on image upload should be added with the image tag
When you upload an image there's space for an (optional) title and description. Where does this content go? imgur?

I expected it to be added as part of the image markdown?

GitHawk Upload by gingerbeardman

<details>
<pre>
Version 1.18.0 (1518216007)
Device: iPhone 6s (iOS 11.2.5)
TestFlight: false
</pre>
</details>

jamuc:
Fixes #1527

## What this does
* Remove the description text field
* When a title is set, use it as alt text
* When return on keyboard is tapped, close keyboard
* Increase size of image view
screen shot 2018-02-20 at 14 17 11

### DIFF
--- a/Classes/Image Upload/ImageUpload.storyboard
+++ b/Classes/Image Upload/ImageUpload.storyboard
@@ -4,7 +4,6 @@
         <adaptation id="fullscreen"/>
     </device>
     <dependencies>
-        <deployment identifier="iOS"/>
         <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="13772"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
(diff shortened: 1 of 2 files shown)


jamuc:
Side note, could also remove the white spaces above and below the image

BasThomas:
> Side note, could also remove the white spaces above and below the image

That would be nice, I think. And then just leave the white spaces in case of a vertical image? Or maybe use the background color?

rnystrom:
I think the white exists because it’s a grouped table view. Otherwise you see the separators awkwardly.

Honestly could probably just remove the table view entirely, it’s a very simple view.

I’m almost more in favor of killing this view entirely.

- Check if we can upload images
- Show the picker
- Upload after you select an image

Something to think about

jamuc:
@rnystrom You mean you wouldn't bother even allowing the user to set a title?

rnystrom:
@jamuc nope, I'd rather edit the image alt text myself in the markdown

SD10:
There is a ton of wasted space here. I think a picker would look much better. Maybe we could even do something like display a collection view/picker under the text input bar (like slack)

rnystrom:
@SD10 that’d be ideal, tho we don’t need to block this PR on it. I’d love to bake that into `MessageViewController` if possible!

rnystrom:
@SD10 that’d be ideal, tho we don’t need to block this PR on it. I’d love to bake that into `MessageViewController` if possible!

rnystrom:
That might have actually been some sort of bug. I got an error in app and resent it. Didn’t realize it sent twice! 😱

PR #1577, opened 2018-02-24
Fixing overlap content after react menu on comment.

### LINKED ISSUE #1391: Overlapping content
Not sure how this happened. I was trying to like the comment and it didn't work, so I then tapped the content to expand it and the emoji thing stayed where it was. 

Imgur

<details>
<pre>
Version 1.17.0 (1515261323)
Device: iPhone 6s (iOS 11.2.2)
TestFlight: true
</pre>
</details>

LucianoPAlmeida:
Hey guys 
This fixes #1391 :))


LucianoPAlmeida:
@rnystrom Should Bitrise CI be running? Remember you saying in a comment on other PR something about it :))

SD10:
Yeah, that's odd. According to [this comment](https://discuss.bitrise.io/t/building-forked-projects/3415/2) this should work

rnystrom:
Hmm this will totally reload everything in the section. I wonder if there’s a more elegant way to do this? What if we just invalidate the layout?

LucianoPAlmeida:
@rnystrom I've tried this 
```swift 
        collectionContext?.invalidateLayout(for: self, completion: nil)
```
But doesn't work, but I have no idea why my knowledge of IGListKit now is very very basic 😅

rnystrom:
@LucianoPAlmeida do you have a good grasp on what exactly is wrong and how this fixes it? I'm a little confused.

LucianoPAlmeida:
@rnystrom 
My stream of thought debugging this was 
The error only appears when the UIMenuController appears and disappears -> So it should be something UIMenuController does that mess up with the collection view graphical context/layout -> First I thought that it seems like a bug on UIMenuController with CollectionViews, but since I don't know the underlyings of how it is presented in the view hierarchy I could not assume that -> So just thought in after menu disappears relayout the cells -> Since the invalidateLayout doesn't work, I thought in just reload the cells and it works :)

LucianoPAlmeida:
@rnystrom What you think? It's possible that I'm missing something on the code? I'm still getting familiar with the codebase :)

LucianoPAlmeida:
Hey @rnystrom Took me a little time, but found that `collectionContext?.invalidateLayout(for: self, completion: nil)` work's I was just not calling it in the right place :))  
But now I think it is fixed :)) 
Sorry for the delay 😅

PR #1578, opened 2018-02-24
Removing #available conditionals

LucianoPAlmeida:
Hey guys :))
Since now the target deployment is iOS 11
Could these conditionals be removed?


rnystrom:
Yes! All of em

PR #1583, opened 2018-02-25
SFSafariViewController pops up GitHub logout page

### LINKED ISSUE #1555: Sign in after sign out issue
First of all, thank you for with amazing project. @rnystrom 
I have an issue with `GitHawk` that I cannot sign in with other credentials after a sign out.
When I press the sign in button, the app shows the same profile as before without asking me to put the credentials.

I haven't checked the code thoroughly, but probably you are storing the `token` in the Keychain and not removing it after a sign out. This may cause the issue.

heshamsalman:
Closes #1555 

not a big fan of https://github.com/rnystrom/GitHawk/pull/1583/files#diff-9f0bae268086ea5dfc986d1a77f28abaR254 but it seemed the quickest way of dealing with the race condition of the dismiss animation


heshamsalman:
not a big fan of https://github.com/rnystrom/GitHawk/pull/1583/files#diff-9f0bae268086ea5dfc986d1a77f28abaR254 but it seemed the quickest way of dealing with the race condition of the dismiss animation

heshamsalman:
@rnystrom just calling `sessionManager.logout()` retains the credentials in the `SFSafariViewController` -- if user clicks the login button, they'll just automatically sign back into whatever user they had authenticated last time.

rnystrom:
Well butts. So how in the world are people supposed to use Oauth in iOS 11 apps and log out of their sessions w/out this ugly UX?

heshamsalman:
Could use `WKWebView` for login and get full access to the cookies, but that comes with its own set of baggage (setting up callbacks, etc)

PR #1589, opened 2018-02-26
[CHOR] Delete old bookmark model and cleanup folder structure

rizwankce:
While browsing the code via our app, found that we have some old model classes. So cleaning up with this PR.


heshamsalman:
:+1: 

I noticed that in #1583 as well, but might be a better bet to merge this because it's a different block of work than what's at the root of #1583

PR #1591, opened 2018-02-26
Refactoring shortcut handler to use enum based Routes

### LINKED ISSUE #708: Refactor shortcut VC indices
This is way too brittle and will be super easy for someone to break.

https://github.com/rnystrom/GitHawk/blob/980fba6cc03cbfef39fb16fb8956e41b5d58807c/Classes/Systems/ShortcutHandler.swift#L13

Need another way to map VCs to tabs that will be caught by the compiler.

rizwankce:
Hopefully fixes #708 and give us a way of handling deep linking possibilities.

- Adds `Route` enum with possible routes - extensible 
- Adds `TabBarController` with `Tab` enum
- Refactores `ShortCutHandler` to use `Route`

Now adding deep linking with `Route` is as easy as like below

```swift
init?(url: URL) {
        switch url.absoluteString {
        case "\(scheme)://tab/inbox": self = .tab(.inbox)
        case "\(scheme)://tab/search": self = .tab(.search)
        case "\(scheme)://tab/bookmars": self = .tab(.bookmarks)
        case "\(scheme)://tab/settings": self = .tab(.settings)
        default: return nil
        }
    }
```

Not sure this is the right/scalable way. Feel free to comment/suggest/close this PR.

### DIFF
--- a/Classes/View Controllers/TabBarController.swift
+++ b/Classes/View Controllers/TabBarController.swift
@@ -1,0 +1,22 @@
+//
+//  TabBarController.swift
+//  Freetime
+//
+//  Created by Rizwan on 26/02/18.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+final class TabBarController: UITabBarController {
+    enum Tab: Int {
+        case inbox = 0
+        case search = 1
+        case bookmarks = 2
+        case settings = 3
+    }
+
+    func showTab(_ tab: Tab) {
+        selectedIndex = tab.rawValue
+    }
+}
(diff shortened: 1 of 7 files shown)


Classes/View Controllers/TabBarController.swift
```diff
@@ -0,0 +1,22 @@
...
+//
+//  Created by Rizwan on 26/02/18.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+final class TabBarController: UITabBarController {
+    enum Tab: Int {
+        case inbox = 0
```
SD10: An enum rawValue of `Int` will default to `0` and increment by `1` for every case, unless specified

PR #1612, opened 2018-03-05
Clean up login client helpers

rnystrom:


Classes/Login/Client+AccessToken.swift
```diff
@@ -0,0 +1,44 @@
...
+        completion: @escaping (Result<AccessTokenUser>) -> Void
+        ) {
+        send(GitHubAccessTokenRequest(
+            code: code,
+            clientId: Secrets.GitHub.clientId,
+            clientSecret: Secrets.GitHub.clientSecret)
+        ) { [weak self] result in
+            switch result {
+            case .success(let response):
+                self?.send(V3VerifyPersonalAccessTokenRequest(token: response.data.accessToken)) { result in
```
BasThomas: Nit: Maybe put this in its own function?

Classes/Login/LoginSplashViewController.swift
```diff
@@ -144,8 +143,13 @@ final class LoginSplashViewController: UIViewController, GithubSessionListener {
         safariController?.dismiss(animated: trueUnlessReduceMotionEnabled)
         state = .fetchingToken
 
-        client.requestAccessToken(code: code) { result in
-            self.handle(result: result, authMethod: .oauth)
+        client.requestAccessToken(code: code) { [weak self] result in
+            switch result {
+            case .error:
```
BasThomas: 🤔, why is this `.error` and not `.failure`?

BasThomas: Also, why do you handle the `.error` here first, when in other places you handle `.success` first?

rnystrom: Two different `Result`s and the naming diverged

PR #1625, opened 2018-03-08
Copyedit on Immutable blog post

BasThomas:


BasThomas:
Great post by the way! 😀 Simple and to the point.

Issue #1544, opened 2018-02-20
Refactor networking into its own library


rnystrom:
I want to put everything:

- `Alamofire`
- `GithubClient`
- All session management stuff
- All Apollo stuff (lib and gen)

Into a single library, call it "GithubAPI" or w/e. I want to prepare to allow things like a watch app or extension, but all of this needs to get moved out.

While we're at it, I really like the networking API design in [this article](https://www.raywenderlich.com/158803/sourcery-tutorial-generating-swift-code-ios). We can make requests + responses all typed and use enums to reduce all of the request options into a single API on `GithubClient`. That should reduce a lot of complexity and greatly decouple requests, responses, and parsing.

Also I want to remove the `FlatCache` from the network client. That's cheating.

Phase 2 will be to migrate and use shared app storage for sessions, bookmarks, etc. But we can do that later.

BasThomas:
Any idea as to when you want to tackle this? Might be a bit hairy when working on enterprise support at the same time.

rnystrom:
I’m still not sold that enterprise support will ever truly work given the GraphQL version issues.

Very least after 1.18 ships. Not a huge rush, but I’d like to clean this stuff up plus pave the way for someone to hack freely on a watch app or extension (I’ll basically ship whatever anyone wants to build).

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BasThomas:
Yep, fair! Will update on my findings regarding Enterprise. 👍

BasThomas:
After #1549 the Request type can probably be pulled out of the GithubClient by the way. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #1636, opened 2018-03-11
Move session management to local pod and add app group

rnystrom:
Moves session to a shared container so that auth'd session info can be shared with extensions. Finishes #1544.


rnystrom:
Gonna land to get Testflight usage and open this open for others to hack at. Extensions are a possibility now!

PR #1646, opened 2018-03-14
Update 3D touch items when switching account.

### LINKED ISSUE #1592: Switch account 3D Touch not updated after switching account
<details>
<pre>
Version 1.18.0 (1519488669)
Device: iPhone X (iOS 11.2.5)
TestFlight: false
</pre>
</details>

jdisho:
This PR tends to fix #1592

### DIFF
--- a/Classes/Settings/SettingsAccountsViewController.swift
+++ b/Classes/Settings/SettingsAccountsViewController.swift
@@ -113,6 +113,7 @@ final class SettingsAccountsViewController: UITableViewController, GitHubSession
 
     func didFocus(manager: GitHubSessionManager, userSession: GitHubUserSession, dismiss: Bool) {
         updateUserSessions()
+        ShortcutHandler.configure(application: UIApplication.shared, sessionManager: sessionManager)
         tableView.reloadData()
     }
 
--- a/Classes/Settings/SettingsAccountsViewController.swift
+++ b/Classes/Settings/SettingsAccountsViewController.swift
@@ -118,4 +118,5 @@ final class SettingsAccountsViewController: UITableViewController, GitHubSession
 
     func didReceiveRedirect(manager: GitHubSessionManager, code: String) {}
     func didLogout(manager: GitHubSessionManager) {}
+
 }
--- a/Classes/Settings/SettingsAccountsViewController.swift
+++ b/Classes/Settings/SettingsAccountsViewController.swift
@@ -113,11 +113,9 @@ final class SettingsAccountsViewController: UITableViewController, GitHubSession
 
     func didFocus(manager: GitHubSessionManager, userSession: GitHubUserSession, dismiss: Bool) {
         updateUserSessions()
-        ShortcutHandler.configure(application: UIApplication.shared, sessionManager: sessionManager)
         tableView.reloadData()
     }
 
     func didReceiveRedirect(manager: GitHubSessionManager, code: String) {}
     func didLogout(manager: GitHubSessionManager) {}
-
 }
--- a/Classes/Systems/AppDelegate.swift
+++ b/Classes/Systems/AppDelegate.swift
@@ -76,8 +76,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         completionHandler(ShortcutHandler.handle(route: route,
                                                  sessionManager: sessionManager,
                                                  navigationManager: rootNavigationManager))
-
-        ShortcutHandler.configure(application: application, sessionManager: sessionManager)
     }
 
     func applicationDidBecomeActive(_ application: UIApplication) {
--- a/Classes/Systems/AppDelegate.swift
+++ b/Classes/Systems/AppDelegate.swift
@@ -30,6 +30,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }()
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
+
+        sessionManager.addListener(listener: self)
+
         // initialize a webview at the start so webview startup later on isn't so slow
         _ = UIWebView()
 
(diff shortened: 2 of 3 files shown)


Classes/Systems/ShortcutHandler.swift
```diff
@@ -27,10 +27,10 @@ struct ShortcutHandler {
         navigationManager: RootNavigationManager
         ) -> Bool {
         switch route {
-        case .tab(let tab):
+        case let .tab(tab):
```
rnystrom: Why change this format? We follow the previous style throughout the rest of the app

Classes/Systems/AppDelegate.swift
```diff
@@ -76,6 +76,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         completionHandler(ShortcutHandler.handle(route: route,
                                                  sessionManager: sessionManager,
                                                  navigationManager: rootNavigationManager))
+
+        ShortcutHandler.configure(application: application, sessionManager: sessionManager)
```
rnystrom: Hmm this will only update when you switch via the 3d touch menu, right? It wont update when you're switching users via settings, right?

I actually think we should add the `AppDelegate` as a listener to the session manager, and on `didFocus(...)` call this line to update the shortcuts.

What do you think?

Classes/Systems/ShortcutHandler.swift
```diff
@@ -57,18 +57,23 @@ struct ShortcutHandler {
...
+                for userSession in sessionManager.userSessions where userSession.username != focusedUsername {
+                    let userSessionIndex = sessionManager.userSessions.index(of: userSession)
+                    let userIcon = UIApplicationShortcutIcon(templateImageName: "organization")
+                    let userItem = UIApplicationShortcutItem(
+                        type: Items.switchAccount.rawValue,
+                        localizedTitle: NSLocalizedString("Switch to \(userSession.username ?? "")", comment: ""),
+                        localizedSubtitle: "From \(focusedUsername)",
+                        icon: userIcon,
+                        userInfo: ["sessionIndex": userSessionIndex ?? 1])
+                    items.append(userItem)
```
rnystrom: Does this add an item for _every_ account you're logged into? Is that something we want to do? Not sure I've seen that discussed anywhere before. I kind of like having the one item to switch to the most-recently-used account.

BasThomas: Yeah, we should limit this, especially since there is a cap of four items anyway, it's just going to be ignored.

Classes/Systems/ShortcutHandler.swift
jdisho: Got it 👍

Classes/Systems/ShortcutHandler.swift
jdisho: So yes, it adds an item for every account that you're logged in. I don't know if you guys have discussed it before or not, but in my opinion is an incomplete feature to switch only to one account when you may have more than 2. Also for me was a bit unclear from which account I was switching on that's why I am showing the `focusedUsername` and where I can switch from there.

In my opinion, it's useful to have this kind of possibility 😊

What do you think? @rnystrom @BasThomas

Classes/Systems/AppDelegate.swift
jdisho: Actually it updates when you're switching users via settings.. 
I am okay adding the `AppDelegate` as a listener to the session manager and configuring `ShortcutHandler ` on `didFocus`, but I don't understand what is the problem the actual implementation...

rnystrom: It doesn’t, the only way we update the menu item now is in the app delegate which requires a cold start. This PR takes it a step closer and updates when you switch accounts via the shortcut, but if you go to Settings and switch accounts there, the previous account won’t be updated.

Classes/Systems/ShortcutHandler.swift
rnystrom: I think we should leave this how it was personally. If you want to switch to other accounts you can do so via settings. I can imagine another UI in app once we have a profile tab too.

jdisho:
Ping @rnystrom  😄

Classes/Settings/SettingsAccountsViewController.swift
```diff
@@ -113,6 +113,7 @@ final class SettingsAccountsViewController: UITableViewController, GitHubSession
 
     func didFocus(manager: GitHubSessionManager, userSession: GitHubUserSession, dismiss: Bool) {
         updateUserSessions()
+        ShortcutHandler.configure(application: UIApplication.shared, sessionManager: sessionManager)
```
rnystrom: Wont the accounts not update if you switch using the 3d touch shortcut? Why not just add the app delegate as a listener and update in one place?

jdisho: I am not understanding right your suggestion 😕  @rnystrom .. 
I would do it like this:  

In AppDelegate
```swift
func application(_ application: UIApplication, performActionFor shortcutItem: UIApplicationShortcutItem, completionHandler: @escaping (Bool) -> Void) {
       ...
         ShortcutHandler.configure(
            application: UIApplication.shared,
            sessionManager: sessionManager)
    }
```

and 

In Settings

```swift
func application(_ application: UIApplication, performActionFor shortcutItem: UIApplicationShortcutItem, completionHandler: @escaping (Bool) -> Void) {
       ...
         ShortcutHandler.configure(
            application: UIApplication.shared,
            sessionManager: sessionManager)
    }
```

rnystrom: In [didFinishLaunching](https://github.com/GitHawkApp/GitHawk/blob/master/Classes/Systems/AppDelegate.swift#L32) add

```swift
sessionManager.addListener(listener: self)
```

Conform to the required protocol, then in the `didFocus(...)` function configure the shortcuts.

PR #1653, opened 2018-03-15
Displaying Target branch for pull requests

### LINKED ISSUE #1389: Display target branch for pull requests
If we could see which branch a pull request was targeted at on GitHawk that would be super useful!

YuryBogdanov:
This PR closes #1389  

Not a big fan of requesting full PR info just for branch name, but didn't find any suitable requests in code. GitHub API doesn't have any lightweight methods either ☹️ any suggestions are welcome though

How this looks in app:
screen shot 2018-04-08 at 15 21 55


YuryBogdanov:
It turns out we don't have to create separate query for branches, since all info we need can be queried by `IssueOrPullRequest`. I just added `baseRefName` and made some changes so it could be visible in interface 🙂

I noticed that there are some project building issues after I merged `master` into my branch to resolve merging conflicts. Something with `__unknown` case. Will have a look later.

rnystrom:
@YuryBogdanov you might need to update your `apollo-codegen` to `0.18.8`

```
$ [sudo] npm install -g apollo-codegen@0.18.8
```

YuryBogdanov:
@rnystrom ya, that was the issue, thank you. Updated Apollo and build succeeded.

Classes/Issues/Branches/IssueTargetBranchCell.swift
```diff
@@ -0,0 +1,17 @@
...
+//  Freetime
+//
+//  Created by Yury Bogdanov on 13/03/2018.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import UIKit
+import SnapKit
+
+final class IssueTargetBranchCell: AttributedStringCell {
```
rnystrom: Would love to land #1674 first and then merge this over to use the new text system!

YuryBogdanov: Ok then 🙂 Waiting for #1674 merge

Classes/Issues/Branches/IssueTargetBranchModel.swift
```diff
@@ -0,0 +1,44 @@
...
+    let targetBranchText: NSAttributedStringSizing
+
+    
+    init(branch: String, width: CGFloat) {
+        
+        let branchTitleAttributes: [NSAttributedStringKey: Any] = [
+            .font: Styles.Text.secondaryCode.preferredFont,
+        ]
+        let titleAttributes: [NSAttributedStringKey: Any] = [
+            .font: Styles.Text.secondaryBold.preferredFont
```
rnystrom: Could we change this to just `secondary`? I don't think the title should be bold.

YuryBogdanov: Sure! Changed it `secondary`, looks a little bit better :)

PR #1673, opened 2018-03-22
Finish refactoring styled text cells to use text lib

rnystrom:


rnystrom:
Just going to merge it all in #1674

PR #1689, opened 2018-03-27
Readd repo notifications

### LINKED ISSUE #1682: Viewing archived repo notifications shows all

BasThomas:
Fixes #1682

### DIFF
--- a/Classes/Notifications/NotificationClient.swift
+++ b/Classes/Notifications/NotificationClient.swift
@@ -59,13 +59,25 @@ final class NotificationClient {
         width: CGFloat,
         completion: @escaping (Result<([NotificationViewModel], Int?)>) -> Void
         ) {
-        githubClient.client.send(V3NotificationRequest(all: all, page: page)) { result in
-            switch result {
-            case .success(let response):
-                let viewModels = CreateViewModels(containerWidth: width, v3notifications: response.data)
-                self.fetchStates(for: viewModels, page: response.next, completion: completion)
-            case .failure(let error):
-                completion(.error(error))
+        if let repo = repo {
+            githubClient.client.send(V3RepositoryNotificationRequest(all: all, owner: repo.owner, repo: repo.name)) { result in
+                switch result {
+                case .success(let response):
+                    let viewModels = CreateViewModels(containerWidth: width, v3notifications: response.data)
+                    self.fetchStates(for: viewModels, page: response.next, completion: completion)
+                case .failure(let error):
+                    completion(.error(error))
+                }
+            }
+        } else {
+            githubClient.client.send(V3NotificationRequest(all: all, page: page)) { result in
+                switch result {
+                case .success(let response):
+                    let viewModels = CreateViewModels(containerWidth: width, v3notifications: response.data)
+                    self.fetchStates(for: viewModels, page: response.next, completion: completion)
+                case .failure(let error):
+                    completion(.error(error))
+                }
             }
         }
     }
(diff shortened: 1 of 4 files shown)


Classes/Notifications/NotificationClient.swift
```diff
@@ -59,13 +59,25 @@ final class NotificationClient {
...
-        githubClient.client.send(V3NotificationRequest(all: all, page: page)) { result in
-            switch result {
-            case .success(let response):
-                let viewModels = CreateViewModels(containerWidth: width, v3notifications: response.data)
-                self.fetchStates(for: viewModels, page: response.next, completion: completion)
-            case .failure(let error):
-                completion(.error(error))
+        if let repo = repo {
+            githubClient.client.send(V3RepositoryNotificationRequest(all: all, owner: repo.owner, repo: repo.name)) { result in
+                switch result {
```
BasThomas: I want to not repeat myself with the completionHandler, but as it relies on the generic type of the Request, I couldn't find a way. Any ideas?

jdisho: I would wrap this into a function: 
```
switch result {
      case .success(let response):
           let viewModels = CreateViewModels(containerWidth: width, v3notifications: response.data)
           self.fetchStates(for: viewModels, page: response.next, completion: completion)
      case .failure(let error):
           completion(.error(error))
}
```

What do you think?

Classes/Notifications/NotificationClient.swift
```diff
@@ -59,13 +59,25 @@ final class NotificationClient {
...
+                case .success(let response):
+                    let viewModels = CreateViewModels(containerWidth: width, v3notifications: response.data)
+                    self.fetchStates(for: viewModels, page: response.next, completion: completion)
+                case .failure(let error):
+                    completion(.error(error))
+                }
+            }
+        } else {
+            githubClient.client.send(V3NotificationRequest(all: all, page: page)) { result in
```
jdisho: This is not an issue related to this PR, but would't be better to have something like this: `github.client` and not `githubClient.client`?

Local Pods/GitHubAPI/GitHubAPI/V3RepositoryNotificationRequest.swift
```diff
@@ -13,7 +13,21 @@ public struct V3RepositoryNotificationRequest: V3Request {
     public var pathComponents: [String] {
         return ["repos", owner, repo, "notifications"]
     }
+    public var parameters: [String : Any]? {
+        return [
+            "all": all ? "true" : "false",
```
jdisho: Why not like this: ` "all": all.description` ?

Classes/Notifications/NotificationClient.swift
BasThomas: That's exactly what I wanted to do, but as the result is generic over `T` (from the `send()` function), the compiler won't allow it.

Classes/Notifications/NotificationClient.swift
BasThomas: cc @rnystrom

Local Pods/GitHubAPI/GitHubAPI/V3RepositoryNotificationRequest.swift
BasThomas: Because I copied it over and didn't know about this returning the exact strings. Nice, changed!

Classes/Notifications/NotificationClient.swift
rnystrom: Probably, leftover from refactors. I want to revisit the whole `GithubClient` class, maybe change it to like `AppSession` or something. It's really just a grab-bag object for injection.

Classes/Notifications/NotificationClient.swift
rnystrom: Probably fine for now, not too offended

PR #1712, opened 2018-04-10
Improve number readability

BasThomas:


rnystrom:
TIL

PR #1716, opened 2018-04-10
Link to the right repo

BasThomas:
I missed this in #1652.

Finally fixes the bug reporter in-app. 🙏


rnystrom:
omg thank you

BasThomas:
I think I ran into it at least 10 times too so yeah 😛

PR #1724, opened 2018-04-12
Updated the IssuesViewController to handle a target branch model

sgl0v:
The application crashes with `fatalError` when the user opens a pull request screen. The reason is that the `IssueTargetBranchModel` object is created by `GithubClient`, but not handled in the `IssuesViewController`. Introduced required updates.


jdisho:
Already merged into master #1720 😊

PR #1740, opened 2018-04-21
IGListKit update

rnystrom:
Prereq for #1739


rnystrom:
Screwed this up but its working so fine

PR #1761, opened 2018-04-24
Update Tabman to latest

msaps:
Noticed that GitHawk was using a fairly old version of Tabman (& associates) - so decided to update to the latest version. 😀 

Fixes quite a few edge cases that were lingering around and also gets rid of the dependency on PureLayout which is nice. Was annoying me being able to see my own bugs in an app that I use every day! 🤪


rnystrom:
@msaps this is based off of a _really old_ commit. Think you can fork a new version and update the pod?

msaps:
@rnystrom yeah sure, my bad! 😀

PR #1781, opened 2018-04-29
Add a watchOS app

rnystrom:
Huge PR b/c I just hacked on & built the whole thing. Some notable changes:

- Podfile restructured to add deps to specific targets
- Split date formatting into its own local lib
- Split string helpers into their own local lib

App features:

- See grouped list of notifications
- Force tap on inbox to show option & mark all read
- Open repo notifications to view actual notifications
- Tap "Read All" button to mark repo notifications read

Most challenging bit was syncing the user session with WatchConnectivity. Not impossible, but a little tricky.

edit: nevermind I'll just do the icon now


Local Pods/GitHubAPI/GitHubAPI/V3NotificationSubject.swift
```diff
@@ -24,3 +24,32 @@ public struct V3NotificationSubject: Codable {
     public let url: URL?
 
 }
+
+public extension V3NotificationSubject {
```
BasThomas: Nice, I was looking at the `String` extension we use now and it had some limitations when there wouldn't be a `URL` present (invitations). This is a better place. 👍

PR #1791, opened 2018-05-05
update bookmark nav item state between viewDidLoad and Issue Load

### LINKED ISSUE #1700: Make bookmark button disabled instead of not appearing
On an issue load, the bookmark button is not shown until it has fully loaded. 

Although small, I propose to allways show said button but make it disabled during loading. Feels a bit more intuitive.

BrianLitwin:
fixes #1700 
Sets up a disabled placeholder bookmark button in navBar during viewDidLoad, then configures the button after the Issue loads, instead of waiting until the Issue loads to put the bookmark button in the navBar. 
Feels like a bit much in terms of refactoring for such a small issue. No worries if that's the case.

### DIFF
--- a/Classes/Bookmark/BookmarkNavigationController.swift
+++ b/Classes/Bookmark/BookmarkNavigationController.swift
@@ -22,12 +22,19 @@ final class BookmarkNavigationController {
     }
 
     // MARK: Public API
-
+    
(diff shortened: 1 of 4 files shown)


BasThomas:
Nice tests!

PR #1798, opened 2018-05-07
⚠️WIP: reply, edit, delete PR comments

jdisho:


jdisho:
I am stuck at something...

Is it possible for the IssueCommentSectionController to know the root comment ID of the thread? 🤔

PR #1805, opened 2018-05-13
Fill in missing switch in watch

rnystrom:


BasThomas:
🤔 could we merge master in before testing / merging maybe to prevent this in the future?

BasThomas:
Ah wait, the merge commit was red. I guess that's enough of an indicator.

rnystrom:
Ya there might be a bug in GitHawk then because it was green and let me merge. Will have to keep an eye out.

BasThomas:
That's right as the branch was not in sync with master. I think you can enable in Settings that a branch has to be in sync with master before merging. The merge commit itself did fail, so everything was as expected.

PR #1832, opened 2018-05-25
updates add reviewers url path

BrianLitwin:
Changed the path component for adding reviewers. Hopefully resolves #1829.  I'm basing the change off the API docs and the unit tests.. I'm not sure how to reproduce this issue manually in the app. 

Also, the unit tests use GithubClient to send the requests as opposed to a URLSession.. the downside being that a failed test doesn't give you much info other than "not a ClientError.unauthorized i.e. 401 response". If it would be preferable to use a URLSession that gives a regular status code to test against, let me know and I'll swap them out.

PR #1838, opened 2018-05-29
Disable flaky tests

rnystrom:
From #1832. These tests are doing real networking which should be faked instead.


BrianLitwin:
Would it be preferable to simplify these and test the request's url against a string, e.g. assert(request.url = "api.github/repos/...") ?

rnystrom:
@BrianLitwin ya that’d probably be more reliable 

PR #1863, opened 2018-06-06
Unread badge count not to show on read on open set

TheCodeTalker:
Hiding Count on  if `read on open` is turned on https://github.com/GitHawkApp/GitHawk/issues/1534

### DIFF
--- a/Classes/Notifications/NotificationsViewController.swift
+++ b/Classes/Notifications/NotificationsViewController.swift
@@ -58,25 +58,11 @@ FlatCacheListener {
     }
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
-        if NotificationClient.readOnOpen(){
+        if NotificationClient.readOnOpen() {
             self.readOnOpen()
         }
-        
     }
-    
-   private func readOnOpen(){
-        var count = 0
-        let showAll = self.showAll
-        for id in notificationIDs {
-            let readOrNot = client.notificationOpened(id: id)
-            if (showAll || !readOrNot) {
-                // count the unread notifications
-                count = count+1
-            }
-        }
-        // readOnOpen is set then update  update bar  badges
-    updateUnreadState(count: count)
-    }
+
     
 
     override func viewDidLoad() {
@@ -102,6 +88,17 @@ FlatCacheListener {
     }
 
     // MARK: Private API
+    // count the unread notifications if read on open is true
+    private func readOnOpen() {
+        var count = 0
+        let showAll = self.showAll
+        for id in notificationIDs where showAll || client.notificationOpened(id: id) == false {
+            // count the unread notifications
+            count += 1
+        }
+        // readOnOpen is set then update  update bar  badges
+        updateUnreadState(count: count)
+    }
 
     @objc func onMore(sender: UIBarButtonItem) {
         let alert = UIAlertController.configured(preferredStyle: .actionSheet)
--- a/Classes/Notifications/NotificationsViewController.swift
+++ b/Classes/Notifications/NotificationsViewController.swift
@@ -56,6 +56,28 @@ FlatCacheListener {
     required init?(coder aDecoder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        if NotificationClient.readOnOpen(){
+            self.readOnOpen()
+        }
+        
+    }
+    
+   private func readOnOpen(){
+        var count = 0
+        let showAll = self.showAll
+        for id in notificationIDs {
+            let readOrNot = client.notificationOpened(id: id)
+            if (showAll || !readOrNot) {
+                // count the unread notifications
+                count = count+1
+            }
+        }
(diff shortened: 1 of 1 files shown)


rnystrom:
Doesn’t this disable all notifications if you have the setting enabled? We want to still display the correct number, but just filter out things that have been locally read.

TheCodeTalker:
got you let me see

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -56,6 +56,28 @@ FlatCacheListener {
...
+        let showAll = self.showAll
+        for id in notificationIDs {
+            let readOrNot = client.notificationOpened(id: id)
+            if (showAll || !readOrNot) {
+                // count the unread notifications
+                count = count+1
+            }
+        }
+        // readOnOpen is set then update  update bar  badges
+    updateUnreadState(count: count)
```
BasThomas: The indentation here is incorrect.

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -56,6 +56,28 @@ FlatCacheListener {
...
     }
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        if NotificationClient.readOnOpen(){
+            self.readOnOpen()
+        }
+        
+    }
+    
+   private func readOnOpen(){
```
BasThomas: Space before the `{`

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -56,6 +56,28 @@ FlatCacheListener {
...
+        var count = 0
+        let showAll = self.showAll
+        for id in notificationIDs {
+            let readOrNot = client.notificationOpened(id: id)
+            if (showAll || !readOrNot) {
+                // count the unread notifications
+                count = count+1
```
BasThomas: spaces around `+`

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -56,6 +56,28 @@ FlatCacheListener {
     required init?(coder aDecoder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        if NotificationClient.readOnOpen(){
+            self.readOnOpen()
+        }
+
```
BasThomas: Can omit this line.

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -56,6 +56,28 @@ FlatCacheListener {
     required init?(coder aDecoder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        if NotificationClient.readOnOpen(){
```
BasThomas: Space before the `{`

Classes/Notifications/NotificationsViewController.swift
BasThomas: I have a hard time parsing what's happening in this function. Can we simplify this?

BasThomas: Some header docs explaining what's going on might help too.

Classes/Notifications/NotificationsViewController.swift
jdisho: I would change the name of `readOrNot` to `isRead`. 
Additionally, you can remove the parenthesis in the `if` clause.

jdisho: `count = count + 1` can be done also like this `count += 1`

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -56,6 +56,28 @@ FlatCacheListener {
...
+            self.readOnOpen()
+        }
+        
+    }
+    
+   private func readOnOpen(){
+        var count = 0
+        let showAll = self.showAll
```
jdisho: Is this necessary?

Classes/Notifications/NotificationsViewController.swift
```diff
@@ -56,6 +56,28 @@ FlatCacheListener {
...
+        for id in notificationIDs {
```
jdisho: I would do this loop a bit shorter: 

```swift
    for id in notificationIDs where showAll || !client.notificationOpened(id: id) {
        count += 1
    }

```

What do you think?

BasThomas: Or even `client.notificationOpened(id: id) == false`, to increase readability.

BasThomas:
A unit test would also be 💯

TheCodeTalker:
ya I will update the pr with the unit test👍

Classes/Notifications/NotificationsViewController.swift
TheCodeTalker: i don't have an idea .

rnystrom:
Looks like this is fixed now

https://github.com/GitHawkApp/GitHawk/blob/master/Classes/Notifications/NotificationsViewController.swift#L111-L131

PR #1865, opened 2018-06-06
Refactor Part 1

jdisho:
Refactor some code in `Bookmark` group.

### DIFF
--- a/Classes/Bookmark/Bookmark.swift
+++ b/Classes/Bookmark/Bookmark.swift
@@ -39,12 +39,10 @@ extension Bookmark: Filterable {
     func match(query: String) -> Bool {
         let lowerQuery = query.lowercased()
 
-        if type.rawValue.contains(lowerQuery) { return true }
-        if String(number).contains(lowerQuery) { return true }
-        if title.lowercased().contains(lowerQuery) { return true }
-        if owner.lowercased().contains(lowerQuery) { return true }
-        if name.lowercased().contains(lowerQuery) { return true }
-
-        return false
+        return type.rawValue.contains(lowerQuery) ||
+            String(number).contains(lowerQuery) ||
+            title.lowercased().contains(lowerQuery) ||
+            owner.lowercased().contains(lowerQuery) ||
+            name.lowercased().contains(lowerQuery)
     }
 }
--- a/Classes/Bookmark/BookmarkCell.swift
+++ b/Classes/Bookmark/BookmarkCell.swift
@@ -30,6 +30,17 @@ final class BookmarkCell: SwipeSelectableCell {
 
         contentView.clipsToBounds = true
 
+        configureImageView()
+        contentView.addSubviews([textView, detailLabel])
+
+        addBorder(.bottom, left: Styles.Sizes.gutter)
+    }
+
+    required init?(coder aDecoder: NSCoder) {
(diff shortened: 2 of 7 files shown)


Classes/Bookmark/Bookmark.swift
```diff
@@ -39,12 +39,10 @@ extension Bookmark: Filterable {
...
         let lowerQuery = query.lowercased()
 
-        if type.rawValue.contains(lowerQuery) { return true }
-        if String(number).contains(lowerQuery) { return true }
-        if title.lowercased().contains(lowerQuery) { return true }
-        if owner.lowercased().contains(lowerQuery) { return true }
-        if name.lowercased().contains(lowerQuery) { return true }
-
-        return false
+        return type.rawValue.contains(lowerQuery) ||
```
BasThomas: 🤔, not sure if this is more readable.

Classes/Bookmark/BookmarkCell.swift
```diff
@@ -14,8 +14,8 @@ final class BookmarkCell: SwipeSelectableCell {
 
     static let titleInset = UIEdgeInsets(
         top: Styles.Sizes.rowSpacing,
-        left: Styles.Sizes.icon.width + 2*Styles.Sizes.columnSpacing,
-        bottom: Styles.Text.secondary.preferredFont.lineHeight + 2*Styles.Sizes.rowSpacing,
+        left: Styles.Sizes.icon.width + 2 * Styles.Sizes.columnSpacing,
+        bottom: Styles.Text.secondary.preferredFont.lineHeight + 2 * Styles.Sizes.rowSpacing,
```
BasThomas: Neat!

Classes/Views/UIView+Extensions.swift
```diff
@@ -0,0 +1,16 @@
...
+//
+//  Created by Joan Disho on 06.06.18.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+extension UIView {
+
+    func addSubviews(_ subviews: [UIView]) {
```
BasThomas: I feel like this is overkill.

BasThomas: And if we do want to keep it, some tests / header documentation would be nice.

jdisho: Why overkill? 🤔

I agree for the tests 👍

SD10: Nit: but I don't like how the local variable name `subviews` shadows `self.subviews`

Classes/Bookmark/Bookmark.swift
```diff
@@ -39,12 +39,10 @@ extension Bookmark: Filterable {
...
+            String(number).contains(lowerQuery) ||
+            title.lowercased().contains(lowerQuery) ||
+            owner.lowercased().contains(lowerQuery) ||
+            name.lowercased().contains(lowerQuery)
```
rnystrom: nit: other places in the codebase put the `||` as the first part of the newline

Classes/Views/UIView+Extensions.swift
rnystrom: Can we remove this? Unless this is fixing a real problem I don't understand.

Issue #1895, opened 2018-07-09
Watch/unwatch should produce haptic feedback


rnystrom:
See `NotificationSectionController`

PR #1898, opened 2018-07-11
Add haptic feedback on watch/unwatch notifications

maniramezan:
Fixes [#1895](https://github.com/GitHawkApp/GitHawk/issues/1895)

### DIFF
--- a/Classes/Notifications/NotificationModelController.swift
+++ b/Classes/Notifications/NotificationModelController.swift
@@ -183,8 +183,10 @@ final class NotificationModelController {
 
         githubClient.client.send(V3SubscribeThreadRequest(id: model.v3id, ignore: model.watching)) { result in
             switch result {
-            case .success: break
+            case .success:
+                Haptic.triggerSelection()
             case .failure:
+                Haptic.triggerNotification(.error)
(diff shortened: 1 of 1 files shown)


Classes/Notifications/NotificationModelController.swift
```diff
@@ -183,8 +183,10 @@ final class NotificationModelController {
 
         githubClient.client.send(V3SubscribeThreadRequest(id: model.v3id, ignore: model.watching)) { result in
             switch result {
-            case .success: break
+            case .success:
+                Haptic.triggerSelection()
```
BasThomas: Shouldn't this be `Haptic.triggerNotification(.success)`?

maniramezan: That's what I wasn't sure about. I was more thinking about making it a bit transparent for user that there's an API call and the chance of failure exists. And only for scenarios that it fails, trigger the failure haptic. But open to the feedback, if you'd feel it'd better to be `notification`, I'm down to update the code

rnystrom: Let's give it a try as-is

Issue #1889, opened 2018-06-26
Do not dim actionable buttons when a notification is marked as read


BasThomas:
From https://github.com/GitHawkApp/GitHawk/pull/1870#issuecomment-398201006 / https://github.com/GitHawkApp/GitHawk/pull/1870#issuecomment-400076295

BrianLitwin:
i don't want to guess about which items are actionable.. I'm thinking Watch and More buttons?

PR #1908, opened 2018-07-15
dims only non-actionable subviews in notification cell

BrianLitwin:
addresses #1889 
this looks better after re-ordering the watch and read buttons


rnystrom:
Brilliant! Nice and simple 

Issue #1920, opened 2018-07-16
Selection bug on Inbox


rnystrom:
After I refreshed, it’s selecting the first cell and I can’t deselect it. So annoying! Not sure what it’s from.

<details>
<summary>Bug Report Dump (Auto-generated)</summary>
<pre>
Version 1.22.0 (1531698972)
Device: iPhone X (iOS 11.4)
TestFlight: true
</pre>
</details>

BasThomas:
Yeah, really weird. As far as I can tell it is pretty random. Sometimes selects everything.

For what it's worth, holding and then dragging deselects the cell.

rnystrom:
Has to be new, maybe something with the content fading?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BasThomas:
Definitely something from the latest beta. Would think it has to be something in the `NotificationCell` right?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rnystrom:
Ah found it. It is from #1908. It's [iterating over all `contentView.subviews`](https://github.com/GitHawkApp/GitHawk/pull/1908/files#diff-bf0607ce020fdba56586bad3f41e227eR186) and changing their `alpha`. The dimmable overlay view is a subview. This is an easy fix.

PR #1925, opened 2018-07-17
explicitly deselect notification cell

BrianLitwin:
For consideration as a fix for #1920. More notes in Issue thread


rnystrom:
See my comment on #1920

Issue #1923, opened 2018-07-16
Make X on close menu item red


rnystrom:
Requires adding color to items struct

<details>
<summary>Bug Report Dump (Auto-generated)</summary>
<pre>
Version 1.22.0 (1531698972)
Device: iPhone X (iOS 11.4)
TestFlight: true
</pre>
</details>

jdisho:
Sorry, I was just testing. 😅 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Huddie:
@rnystrom Is this not already red?

PR #1931, opened 2018-07-18
Setup page rewrite + Adding color to ContrastContextMenu

Huddie:
[Setup] Rewrote the getting started page.
The current setup page helps with getting an OAuth key but there are a few required steps that need to be taken in the project itself in order to get the app running on your iPhone (for contributors). 
Those steps are outlined in the new document.

[Issue #1923] Updated the ContrastContextMenuItem struct to contain a color element and updated the color element to recolor red


Huddie:
@rnystrom I'll seporate the PR's. Not sure why the podfiles. Storyboard should have 1 change (Adding the Review App cell). I'll look into fixing those things up

Issue #1927, opened 2018-07-18
Add a "Rate on App Store" cell to settings


rnystrom:
Should be really simple. Tap on cell and open rating dialog in App Store. Note this should _not_ show the in-app rating, instead want to open App Store. Trying to drive more reviews.

Leaving this up for grabs for someone else to contribute! Great starter task.

rizwankce:
Closed by #1939

PR #1939, opened 2018-07-19
[Settings]  [Issue #1927] Added Rate on App Store cell 

Huddie:
Storyboard changes are due to the new cell addition. Y pos of each cell in 2nd section were moved down as the new cell was placed on the very top.

### DIFF
--- a/Classes/Settings/SettingsViewController.swift
+++ b/Classes/Settings/SettingsViewController.swift
@@ -24,7 +24,7 @@ NewIssueTableViewControllerDelegate {
     @IBOutlet weak var versionLabel: UILabel!
     @IBOutlet weak var reviewAccessCell: StyledTableCell!
     @IBOutlet weak var githubStatusCell: StyledTableCell!
-    @IBOutlet weak var reivewOnAppStoreCell: StyledTableCell!
+    @IBOutlet weak var reviewOnAppStoreCell: StyledTableCell!
     @IBOutlet weak var reportBugCell: StyledTableCell!
     @IBOutlet weak var viewSourceCell: StyledTableCell!
     @IBOutlet weak var signOutCell: StyledTableCell!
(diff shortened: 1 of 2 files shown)


Classes/Settings/SettingsViewController.swift
```diff
@@ -101,6 +102,8 @@ NewIssueTableViewControllerDelegate {
             onAccounts()
         } else if cell === githubStatusCell {
             onGitHubStatus()
+        }else if cell === reivewOnAppStoreCell {
```
BasThomas: Nit: space before `else`.

Classes/Settings/SettingsViewController.swift
```diff
@@ -134,6 +137,14 @@ NewIssueTableViewControllerDelegate {
             else { fatalError("Should always create GitHub Status URL") }
         presentSafari(url: url)
     }
+  
+    func onReivewOnAppStore() {
```
BasThomas: `onReviewOnAppStore*`

Classes/Settings/SettingsViewController.swift
```diff
@@ -134,6 +137,14 @@ NewIssueTableViewControllerDelegate {
...
+      guard let url = URL(string: "itms-apps://itunes.apple.com/us/app/githawk-for-github/id1252320249")
+        else { fatalError("Should always be valid app store URL") }
+      if UIApplication.shared.canOpenURL(url) {
+        UIApplication.shared.open(url, options: [:], completionHandler: nil)
```
BasThomas: Nit: you can omit the `options` and `completionHandler` parameter as they default to `[:]` and `nil` respectively.

Classes/Settings/SettingsViewController.swift
Huddie: Whoops. My bad.

Classes/Settings/SettingsViewController.swift
Huddie: Makes sense.

Classes/Settings/SettingsViewController.swift
Huddie: Got it. @BasThomas Whats Nit stand for?

BasThomas: It’s defined [here](https://www.merriam-webster.com/dictionary/nit) as a “minor shortcoming”. In this case that it doesn’t really need fixing, but it would be nice.

Issue #1957, opened 2018-07-20
Issue didn’t render as link


rnystrom:
![GitHawk Upload by rnystrom](https://i.imgur.com/P4k6dfV.jpg)

<details>
<summary>Bug Report Dump (Auto-generated)</summary>
<pre>
Version 1.22.0 (1532006598)
Device: iPhone X (iOS 11.4)
TestFlight: true
</pre>
</details>

rnystrom:
Can’t tap on it either

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BrianLitwin:
Looks like [this piece](https://github.com/GitHawkApp/GitHawk/blob/eb19255d97e47846c3e9da17f4fc9cbbf2f0b2a2/Classes/Issues/Comments/Markdown/String%2BDetectShortlink.swift#L19-L22]) isn't recognizing the link.. specifically, I think it'll miss anything without a space before the "#", ie ( #1958) will work but not (#1958)

rnystrom:
@BrianLitwin all yours!!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #1963, opened 2018-07-22
patches shortlink regex to include links inside parentheses

BrianLitwin:
bare minimum patch to consider for #1957
added an alternative expression that matches an opening "("


BasThomas:
Are there tests around this already to which we can add this case? Would be nice!

BrianLitwin:
@BasThomas heard that! I'll update this pr with tests

PR #1967, opened 2018-07-22
updates/tests shortlink string matching

BrianLitwin:
re-did PR #1963 
- added unit tests 
- added a more specific RegEx clause to match issue numbers enclosed in parentheses 
- added a RegEx clause to catch invalid trailing characters. #1900W is considered a match in the current RegEx pattern. 

The RegEx is more long-winded than before but easy to test.


BasThomas:
Awesome!

BrianLitwin:
actually this is farting up downstream.. the issue links aren't working when you tap on them, needs more work.

Issue #1905, opened 2018-07-15
Flip checkbox and watch icons on Inbox


rnystrom:
Just occurred to me, I think that’d feel like a more natural UX hierarchy.

<details>
<summary>Bug Report Dump (Auto-generated)</summary>
<pre>
Version 1.22.0 (1531607209)
Device: iPhone X (iOS 11.4)
TestFlight: true
</pre>
</details>

Issue #25, opened 2017-06-28
Username highlighting and tap


rnystrom:
V1 open profile in web view

PR #1971, opened 2018-07-23
updates/tests detectShortlink method 

BrianLitwin:
addressing #1957 

I think this regex pattern matches more closely Github's pattern. It seems Github allows Issue numbers to be embedded inside anything besides alphabet characters, eg !#1905! is fine but not t#1905 or #1905t.  

The unit tests cover 4 things:
     1) positive text match
     2) that the correct text is blue/linked (eg the parentheses aren't blue also)
     3) that the MarkdownIssue is linked correctly with an issue number
     4) that the string displayed after reformatting is equal to the original string

Should be easy to test anything going forward. 

*It might be helpful to look at this comment on laptop and phone to see that it the links in the app align with the way they appear online

For summary/testing: 

Text that should be linked:
(#25)
(#25
#25) 

Text that shouldn't be linked: 
f#25
#25f

### DIFF
--- a/FreetimeTests/DetectShortlinkTests.swift
+++ b/FreetimeTests/DetectShortlinkTests.swift
@@ -104,14 +104,7 @@ class DetectShortlinkTests: XCTestCase {
         XCTAssertEqual(containsLink.linkText, "#432")
         XCTAssertEqual(containsLink.issueNumber, 432)
         XCTAssertEqual(builder.allText, testString)
-        
-        testString = "#4!"
-        builder = setupBuilder(with: testString)
-        containsLink = checkForIssueLink(builder.styledTexts)!
-        XCTAssertEqual(containsLink.linkText, "#4")
-        XCTAssertEqual(containsLink.issueNumber, 4)
-        XCTAssertEqual(builder.allText, testString)
-        
+
         testString = "!#4 yada yada"
         builder = setupBuilder(with: testString)
         containsLink = checkForIssueLink(builder.styledTexts)!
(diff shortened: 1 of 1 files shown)


FreetimeTests/DetectShortlinkTests.swift
```diff
@@ -0,0 +1,144 @@
...
+        XCTAssertEqual(builder.allText, testString)
+        
+        testString = "Fixes(#432)"
+        builder = setupBuilder(with: testString)
+        containsLink = checkForIssueLink(builder.styledTexts)!
+        XCTAssertEqual(containsLink.linkText, "#432")
+        XCTAssertEqual(containsLink.issueNumber, 432)
+        XCTAssertEqual(builder.allText, testString)
+        
+        testString = "#4!"
```
BrianLitwin: this is a duplicate test.. tests same thing as "Fixes #12)" above

FreetimeTests/DetectShortlinkTests.swift
```diff
@@ -0,0 +1,144 @@
...
+        
+        builder = setupBuilder(with: "#123F")
+        containsLink = checkForIssueLink(builder.styledTexts)
+        XCTAssertNil(containsLink)
+        
+        builder = setupBuilder(with: "f#123")
+        containsLink = checkForIssueLink(builder.styledTexts)
+        XCTAssertNil(containsLink)
+        
+        builder = setupBuilder(with: "Fixes#1234")
```
BrianLitwin: this is a duplicate of the test above it

Classes/Issues/Comments/Markdown/String+DetectShortlink.swift
```diff
@@ -59,6 +59,14 @@ extension String {
...
                     ])
                 .restore()
+            
+            //if match's trailing character is not a number, include that
+            if let right = substring(with: NSRange(location: match.range.right - 1, length: 1)) {
+                let nums: Set<Character> = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9"]
+                if Set(right).isSubset(of: nums) == false {
+                    builder.add(text: right)
+                }
+            }
```
BrianLitwin: not sure why I wrote this elaborate number checking scheme instead of just seeing if the string could be cast to an Int

Issue #1109, opened 2017-11-28
GitHub Discussions


BasThomas:
GitHub recently introduced [Discussions](https://github.com/blog/2471-introducing-team-discussions). Although there is not API (yet), it wouldn't hurt to think if / how this would fit into the app.

heshamsalman:
I like this idea, and think we should be using it instead of issues for some stuff 👍

Sherlouk:
So in my head I'm seeing User Profiles (#317) as the root to this, I think once we have user profiles we can then get organisation/team profiles fairly easily.

Obviously a user can have multiple repositories, multiple teams, multiple organisations etc etc

A team can have multiple repositories, multiple users, and now discussions but will likely share most of the UI (but with a bit of planning, we can design the UI to use a TabMan controller and therefore switch between repos and discussions lists?)

Also we should create a GitHawk team so we can try this out 😅

BasThomas:
Hey look they added an API: https://developer.github.com/v4/previews/#team-discussions

Sherlouk:
We really need to get a MVP user profiles out... opens up so many doors and cool shi!

SD10:
Wow, look they even have hovercard info. That was surprisingly fast 😅

PR #1987, opened 2018-07-28
Initial support for the discussions endpoint

BasThomas:
References #1109 

I don't want to do more than this as there is [also a V4 API](https://developer.github.com/v4/previews/#team-discussions), but I'd need some help on how they work 😇

### DIFF
--- a/Local Pods/GitHubAPI/GitHubAPI/V3DiscissionsRequest.swift
+++ b/Local Pods/GitHubAPI/GitHubAPI/V3DiscissionsRequest.swift
@@ -1,0 +1,41 @@
+//
+//  V3DiscissionsRequest.swift
+//  GitHubAPI
+//
+//  Created by Bas Broek on 28/07/2018.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+public struct V3DiscussionsRequest: V3Request {
+    public typealias ResponseType = V3DataResponse<[V3Discussion]>
+    public var pathComponents: [String] {
+        return ["teams", teamID, "discussion"]
+    }
+
+    public var parameters: [String : Any]? {
+        return ["per_page": 50, "page": page, "direction": direction.rawValue]
+    }
+
+    public var headers: [String : String]? {
+        return ["Accept": "application/vnd.github.echo-preview+json"]
+    }
+
+    public enum Direction: String {
+        case ascending = "asc"
+        case descending = "desc"
+    }
+
+    public let teamID: String
+    public let page: Int
+    public let direction: Direction
+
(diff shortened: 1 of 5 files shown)


Local Pods/GitHubAPI/GitHubAPITests/Discussions/discussions.json
```diff
@@ -0,0 +1,50 @@
...
+        "html_url": "https://github.com/orgs/github/teams/justice-league/discussions/1",
+        "node_id": "MDE0OlRlYW1EaXNjdXNzaW9uMQ==",
+        "number": 1,
+        "pinned": false,
+        "private": false,
+        "team_url": "https://api.github.com/teams/2343027",
+        "title": "Our first team post",
+        "updated_at": "2018-01-25T18:56:31Z",
+        "url": "https://api.github.com/teams/2343027/discussions/1",
+        "reactions": {
```
BasThomas: This is in the [example json](https://developer.github.com/v3/teams/discussions/#response) but we shouldn't try to decode it yet; it needs an [Accept header](https://developer.github.com/v3/teams/discussions/#list-discussions) (`application/vnd.github.squirrel-girl-preview`)... but we already [have one])https://github.com/GitHawkApp/GitHawk/pull/1987/files#diff-c452e40c33cddad76f5cc2ba892c06d8R22).

Should we refactor that to accept a `[(String, String)]` instead of a `Dictionary`?

Local Pods/GitHubAPI/GitHubAPI/V3DiscissionsRequest.swift
```diff
@@ -0,0 +1,41 @@
...
+
+    public var parameters: [String : Any]? {
+        return ["per_page": 50, "page": page, "direction": direction.rawValue]
+    }
+
+    public var headers: [String : String]? {
+        return ["Accept": "application/vnd.github.echo-preview+json"]
+    }
+
+    public enum Direction: String {
```
BasThomas: And we might want to abstract this to somewhere else so it can be reused — I think it is supported by some other requests as well.

rnystrom:
What's the end goal for this? How are we gonna add discussion support?

BasThomas:
I guess there's no endgoal yet, I just wanted to chip away at some of the work that'd be required to add support for this, and make the GitHubAPI a bit more complete.

rnystrom:
I need to read up on how to even use discussions, never have.

BasThomas:
OK so organizations can have teams (GitHawkApp has 0 atm) and teams can have discussions. It's just a place for a team within an organization to have a forum-like thread to discuss stuff.

rnystrom:
Closing as its stale

PR #1990, opened 2018-07-28
Make inbox buttons more tappable

### LINKED ISSUE #1968: Inbox cell hit targets too small
I think we can add more padding (insets) to the action buttons to give them better hit targets

rnystrom:
Fixes #1968


rnystrom:
@BasThomas 😝

PR #1998, opened 2018-07-29
Improved review prompt, open to write action

rnystrom:
simulator screen shot - iphone x - 2018-07-29 at 09 12 12


BasThomas:
I don’t really like that the App Store gives zero context in the modal; I felt unsure about what was going to happen in this view. 😐

Even though I just tapped “leave a review”. 

What do you think?

GitHawk Upload by BasThomas

rnystrom:
@BasThomas not bothered by it. Is there a better option?

BasThomas:
Don't think so. It's fine :)

PR #2022, opened 2018-07-31
Choose double tap

Huddie:
Allowing users to choose the reaction that occurs when double tapping on a comment

PR #2023, opened 2018-07-31
None reaction

Huddie:
This contains the same changes as the choose-double-tap PR with the additinal disabled option


rnystrom:
Why a separate PR? Can you instead just make a new branch off of #2022 so we can merge that first, then this?

PR #2029, opened 2018-08-04
Fixed preferredContentSizeCategory called on background thread issue

### LINKED ISSUE #1907: preferredContentSizeCategory called on background thread
Repeat of #1225.. as stated in original thread, not an urgent issue, just throwing a "Main Thread Checker: UI API called on a background thread" warning in XCode. 

The preferredFont property that's causing the warning is accessed [here](
https://github.com/GitHawkApp/GitHawk/blob/3028081c3c0d5b052ad6dc16f916208ff9d6517b/Classes/Notifications/NotificationCell.swift#L28) and then [here](https://github.com/GitHawkApp/GitHawk/blob/3028081c3c0d5b052ad6dc16f916208ff9d6517b/Classes/Notifications/NotificationCell.swift#L22) in NotificationCell, from [this function](https://github.com/GitHawkApp/GitHawk/blob/3028081c3c0d5b052ad6dc16f916208ff9d6517b/Classes/Notifications/CreateNotificationTitle.swift#L22).

ismetanin:
**Changes:**
- Added `static` property `preferred` to `UIContentSizeCategory` - now we can use this property instead of `UIApplication.shared.preferredContentSizeCategory`. These changes fix #1907 issue
- Added `UIContentSizeCategoryChangeHandler` to handle changing of preferredContentSizeCategory
- Changed access level from internal to private for methods in `ForegroundHandler`

**Steps to reproduce #1907**:
1. Open the app
2. Select the "Settings" tab
3. Select the "View Source Code" sell
4. ✅ You can see the "UI API called from background thread" in your Runtime errors

**Notes:**
I also found a bug on my iPhone 6S with iOS 11.4: when the application is running and I increase `Larger Text` in Settings/Accessibility, layout in the application goes wrong. I attached screenshots below.
We can fix it easily by removing `UIContentSizeCategoryChangeHandler`, but then a user will need to restart the application to see changed fonts. 

group

### DIFF
--- a/Classes/Systems/UIContentSizeCategoryChangeHandler.swift
+++ b/Classes/Systems/UIContentSizeCategoryChangeHandler.swift
@@ -1,0 +1,34 @@
+//
+//  UIContentSizeCategoryChangeHandler.swift
+//  Freetime
+//
+//  Created by Ivan Smetanin on 08/04/2018.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import UIKit
+
+final class UIContentSizeCategoryChangeHandler {
+
+    public static let shared = UIContentSizeCategoryChangeHandler()
+
+    private init() {}
+
+    func setup() {
+        let center = NotificationCenter.default
+        center.addObserver(
+            self,
+            selector: #selector(contentSizeCategoryDidChange),
+            name: .UIContentSizeCategoryDidChange,
+            object: nil
+        )
+    }
+
+    // MARK: Private API
+
+    @objc private func contentSizeCategoryDidChange() {
+        print("change")
+        UIContentSizeCategory.preferred = UIContentSizeCategory.preferred
+    }
+
+}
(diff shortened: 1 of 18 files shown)


Classes/Systems/UIContentSizeCategoryChangeHandler.swift
```diff
@@ -0,0 +1,34 @@
...
+            selector: #selector(contentSizeCategoryDidChange),
+            name: .UIContentSizeCategoryDidChange,
+            object: nil
+        )
+    }
+
+    // MARK: Private API
+
+    @objc private func contentSizeCategoryDidChange() {
+        print("change")
```
rnystrom: remove?

ismetanin: @rnystrom Yeap, I forgot to delete :(

ismetanin:
@rnystrom Yeas, of course. I’ll split these changes into two PRs. Can I add the first ones branch as a target for the second one? Or how should I provide these changes?

ismetanin:
Close PR. See you in next tow PRs.

rnystrom:
@ismetanin ya if the second depends on the first that’s fine 

ismetanin:
@rnystrom But I don't have an access to create pull requests like this. Because I'm creating them from my fork. Anyway, I've created #2037 yet, so you can check this and after that, I'll push the second one

PR #2043, opened 2018-08-06
Added preferred content size category change handling

ismetanin:
According to #2029

**Changes:**
- Added `UIContentSizeCategoryChangeHandler` to handle changing of `preferredContentSizeCategory`
- Changed access level from internal to private for methods in `ForegroundHandler`


rnystrom:
Returning to prev convo: I'm not sure about this b/c the whole app doesn't support changing the text size. So w/ this enabled some views will update to the new content size while others wont. I'd almost rather `exit()` the app to force a restart than have a mix-matched UI.

What do you think?

ismetanin:
@rnystrom I think this is better than find and fix all views in the app
I'll update my PR

BasThomas:
I wonder if this is worth doing though if we need to force quit the app. 😕

ismetanin:
@BasThomas You can see what's going with the layout in #2029. I think that we need to restart the app or we will need to layout some views but before layout, we should find these views in the app.

BasThomas:
Ah right, it at least doesn't seem OK if something like that screenshot can happen 😛

rnystrom:
Worth a try. Holding off til we ship the latest prod version tho.

Classes/Systems/ForegroundHandler.swift
```diff
@@ -0,0 +1,58 @@
...
+        if CACurrentMediaTime() - backgrounded > threshold {
+            // there seems to be a bug where refreshing while still becoming active messes up the iOS 11 header and
+            // refresh control. put an artificial delay to let the system cool down?
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.3, execute: {
+                self.delegate?.didForeground(handler: self)
+            })
+        }
+    }
+
+}
```
rnystrom: Can we delete this if its unused?

ismetanin: Yeap, though that I've deleted this when merged master

Issue #609, opened 2017-10-18
Enable copying


rnystrom:
Not sure the best approach here. Options:

- Individual cell copying. Have to support on every type of cell, but can highlight individual hunks of text
- Copy all original md. Noisier, but very easy to implement (`UIAlertController` etc)

Thoughts?

Inspired by #600

Sherlouk:
I quite like the idea of a "See Original" action, which just shows you the entire markdown in it's own view (and that's copyable)

Would be super helpful for testing where the markdown content is the cause of some render issues!

gingerbeardman:
My preferences FWIW:

1. highlight individual chunks of text the iOS standard way so i get to use the share sheet
2. less good: view markdown so that I can do as above, but less good as I will have to content with markdown special characters
3. worst option: copy all text in cell, because that will require editing after the fact

Sherlouk:
There's still nothing preventing us doing both, as I do see two different use cases for this especially with the functionality of being able to quote a previous message which depends on this functionality but would be easier with paragraphs rather than entire posts!

rnystrom:
FYI we wont be able to do normal text selection b/c we're not using `UITextView` (and wont for perf reasons).

I'd be open to a "view raw markdown" alert action on a comment's "..." menu item, open a modal `UITextView`, and let you select + highlight from that?

gingerbeardman:
Works for me.

PR #2073, opened 2018-08-10
Long press details

Huddie:
Allows for "quick" access to the more section (Action sheet) 

Pretty common. Just thought I'd make a PR and whatever happens with it happens.

Quick note: The didTapMore method took 2 params. cell & sender. Not sure why cell was needed so I cleaned it out, it wasn't used anywhere. (The more button still works)


rnystrom:
I’d actually like to reserve the long press to highlight and “copy” text on cells. What do you think?

rnystrom:
Like text views. Long press, highlights all of the text (requires updates to StyledTextKit probably), then show a menu control that says "Copy". Tap that and put the text in UIPasteboard. Maybe have a "Copy" and "Copy Markdown" option.

See #609

Gonna close this b/c having the menu in one place and enabling copy is a simpler UX IMO. I want to avoid cluttering w/ too many actions/controls.

Issue #2074, opened 2018-08-10
Resign keyboard when select search result


ismetanin:
Now when I select search result on the Bookmarks tab it looks buggy - keyboard hides on the repository screen

![2fmfqp](https://user-images.githubusercontent.com/11653316/43949146-5a9dee52-9c95-11e8-891b-0a617cd794d8.gif)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rnystrom:
Ew yes!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rnystrom:
@Huddie wanna pick this off too? Probably exactly the same fix as #2071

Huddie:
@rnystrom Sure. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BasThomas:
Fixed in #2077 :)

PR #2077, opened 2018-08-10
Resign keyboard search

Huddie:
Addresses #2074 

Search already employed resign properly so this is only for bookmarks


Huddie:
Another keyboard bits the dust!

PR #2082, opened 2018-08-10
lines up contextMenu labels

BrianLitwin:
quick patch for the imageViews in contextMenu having variable widths (between 18 and 25 pt). 

before: 
screenshot 2018-08-10 13 58 47
 
after: 
screenshot 2018-08-10 13 59 23

I'm reluctant to change the imageView's frames as I think the system does a good job of laying them out as is, but I'm all ears if you all have suggestions for a better way to do this.


BasThomas:
Can’t unsee

rnystrom:
omg

rnystrom:

PR #2084, opened 2018-08-11
Clean up reactions settings

rnystrom:
From #2022 cc @Huddie 

There were enough changes that it would've taken a while to update. Some things that I want to call out:

- Explicit "disabled" value vs hacking `"Disabled"` into the unknown value
- Localization for disabled
- Code formatting! 4-space tabs, `{` on sameline (TODO SwiftLint....)
- Fixed up some naming
- Files weren't added to .xcodeproj, build was failing


Huddie:
@rnystrom Yicks. My bad. Thanks for fixing it up. 

Issue #2067, opened 2018-08-09
Add a "Is something wrong?" button on inbox zero


rnystrom:
I get the question a lot that there are notifications missing when they exist on the web. Maybe we can add a "fine print" button at the bottom of the inbox-zero display that opens the "Review Github Access" page. That usually seems to solve the problem.

BrianLitwin:
Is this the right track? 

![simulator screen shot - iphone x - 2018-08-09 at 23 32 21](https://user-images.githubusercontent.com/26695477/43937588-9f869bae-9c2c-11e8-9814-9ffcb56f4119.png)

rnystrom:
Ya! Few notes: can the text be even smaller? Also no border, we don’t want to over emphasize.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BrianLitwin:
V2.. bumped the font from 13 to 12. 
<img width="300" alt="screenshot 2018-08-10 19 16 10" src="https://user-images.githubusercontent.com/26695477/43984975-ee87f23e-9cd1-11e8-9c62-afe7e1cee841.png">

What you think @Huddie ?

rnystrom:
Ok last nit, maybe it just says “Review GitHub Access” in all grey. Single line, simple and clear, but not a dominant color that takes away from the animation.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Huddie:
Looks great. I agree removing the ... and upper casing the to Access.
Removing the blue makes me feel like it’s not clickable.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #2085, opened 2018-08-11
add GithubAccess btn to noNewNotifications cell

BrianLitwin:
Final: 

screenshot 2018-08-11 12 16 07

Thanks for feedback! 

quick edit: capitalized the A in "Review GitHub access" in 2nd commit

### DIFF
--- a/Classes/Notifications/NoNewNotificationsCell.swift
+++ b/Classes/Notifications/NoNewNotificationsCell.swift
@@ -9,11 +9,17 @@
 import UIKit
 import SnapKit
 
+protocol ReviewGithubAccessDelegate: class {
+    func reviewGithubAccessButtonTapped()
+}
+
 final class NoNewNotificationsCell: UICollectionViewCell {
 
     let emojiLabel = UILabel()
     let messageLabel = UILabel()
+    let reviewGithubAccessButton = UIButton()
     let shadow = CAShapeLayer()
+    weak var reviewGithubAccessDelegate: ReviewGithubAccessDelegate?
 
     override init(frame: CGRect) {
         super.init(frame: frame)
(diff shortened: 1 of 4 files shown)


Classes/Notifications/NotificationsViewController.swift
```diff
@@ -304,4 +307,13 @@ BaseListViewController2EmptyDataSource {
...
         didSingleTapTab()
     }
+    
+    // MARK: ReviewGithubAccessDelegate
+    func reviewGithubAccessButtonTapped() {
+        //copied/pasted from SettingsViewController .. could consolidate 
+        guard let url = URL(string: "https://github.com/settings/connections/applications/\(Secrets.GitHub.clientId)")
+            else { fatalError("Should always create GitHub issue URL") }
+        // iOS 11 login uses SFAuthenticationSession which shares credentials with Safari.app
+        UIApplication.shared.open(url, options: [:])
```
BasThomas: Nit: you can omit the `options` parameter as it defaults to an empty Dictionary.

Classes/Notifications/NoNewNotificationsCell.swift
```diff
@@ -42,6 +48,23 @@ final class NoNewNotificationsCell: UICollectionViewCell {
             make.top.equalTo(emojiLabel.snp.bottom).offset(Styles.Sizes.tableSectionSpacing)
         }
 
+        reviewGithubAccessButton.setTitle("Review GitHub Access...", for: .normal)
```
BasThomas: Should at least be a localized string.

rnystrom: Yes! Needs this then g2g I think

rnystrom: Also can we omit the “...”?

BrianLitwin:
updated! 👍

BrianLitwin:
If this looks tacky, we could replace with a question mark (or "need help" icon that triggers an alert or ContextMenu that shows Review GitHub Access + help options? We could also put this in the left hand nav bar button's alert controller under "View all", or possibly give it it's own space on the navigation bar.

Huddie:
#2067

Huddie:
If this is a button I feel like blue makes more sense and fits better stylistically with the app (Like load more). Gray has a disabled or side note feel, blue gives off a button feel. IMO

rnystrom:
@BrianLitwin mind resolving conflicts really quick?

Issue #317, opened 2017-09-17
User Profiles


Sherlouk:
- Adds new tab to bar "My Profile"
- Where a user can tap on a name/avatar to view profile, it should open natively

Page should have:
https://developer.github.com/v4/reference/object/user/

- Name, Avatar and Handle
- Description/Bio
- Details
  - Location/Website/Email/etc
- Organisations
  - Grid of icons tappable to open
- Pinned Repositories
- All Repositories (Limit 10? then tap on a "See all" button to see entire list)
- "More" button in corner
  - Block
  - Follow

For the most part the API is very similar to the [Organisation](https://developer.github.com/v4/reference/object/organization/) one - So we should be able to support both of them with a common interface

Sherlouk:
@rnystrom Are you picking this up or is it free game?

rnystrom:
I'm gonna knock this fella out soon

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

weyert:
Is this something you can use StackedSC's for? Still not grasping when you want to use that

Sherlouk:
@rnystrom You still working on this bad boy?

rnystrom:
On hold atm, @ocrickard might be having at it

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ocrickard:
I'm probably going to have to bail on this.

Sherlouk:
https://github.com/tipsy/github-profile-summary

If we wanted to do something a bit more unique with our profiles (and not to 100% match what desktop has) then I quite like the profile summary stuff above

rnystrom:
Would be so cool to have some visuals! Some ideas:

Maybe there should be a “stats” tab? So the root tab is overview, then there could be:

- Stars
- repos
- stats

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Sherlouk:
So GitHub desktop has:
- Overview
  - Pinned Repositories
  - Contribution/Commit Graph
  - Contribution Activity
- Repos
- Stars
- Followers
- Following
all that as well as the sidebar which has all the actual user information

I think for the followers/following, we could do (on the Overview? page) a horizontal view of the most recent 10 or something with a link to open a full list vaguely similar to assignees on issues?

----

I drew a little example of what I'm thinking, as a starting point anyway

[...]

PR #2113, opened 2018-08-13
User profile WIP

BrianLitwin:
I didn't want to go too far on this w/o letting you all stop me if I'm way off on something. I hope that the GraphQL models and client are something other people can use; i'm not attached to any of the views or layout. 

Referring to @Sherlouk 's design in #317, are we going to have a horizontal, scrollable list of followers and users we're following? I've played with that idea and I find it cool to be able to scroll people's profile pics, but I'm not sure what others will think. 

I'm most worried about that top section with the picture + profile detail items.. that's going to take some feel to lay out correctly. 

Nav bar items that need implementing: 
- the follow/unfollow item in top right 
- ~~a title/subtitle navBar title with name + login~~ (added this in 3rd commit)


rnystrom:
@BrianLitwin regarding the follow lists, I think we should start w/ _very simple_ profiles and expand it from there. Maybe we have buttons w/ follower/following counts, but tapping on that opens the web view. _Then_ we create native versions of those.

Thinking of ways to keep the very first PR more manageable. Also if we change course at any time, I don't want to throw away any work!

BrianLitwin:
okay lemme walk this back and re-submit after we get some clarity on what the simplest MVP looks like.

PR #2114, opened 2018-08-13
Changing accounts resets to main tab fix

### LINKED ISSUE #2096: Changing accounts now resets to main tab

ismetanin:
Fixes #2096

**Note:**
I'm not sure that my solution is good because that's not a good idea to use `dismiss` property from `didFocus` method but I didn't find any better. If anyone knows some solution better - you are welcome

### DIFF
--- a/Classes/Systems/RootNavigationManager.swift
+++ b/Classes/Systems/RootNavigationManager.swift
@@ -54,10 +54,13 @@ final class RootNavigationManager: GitHubSessionListener {
         } else {
             block()
         }
+
+        self.tabBarController?.selectedIndex = 0
     }
 
-    public func resetRootViewController(userSession: GitHubUserSession?, isNeedRenderInbox: Bool = true) {
+    public func resetRootViewController(userSession: GitHubUserSession?) {
         guard let userSession = userSession else { return }
+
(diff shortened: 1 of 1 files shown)


Huddie:
@ismetanin  What about just placing the `selectedIndex` code `public func showLogin(animated: Bool = false) {`

```
public func showLogin(animated: Bool = false) {
        guard let root = rootViewController else { return }

        let login = newLoginViewController()
        login.modalPresentationStyle = .formSheet

        let block: () -> Void = { root.present(login, animated: animated) }

        if let presented = root.presentedViewController {
            presented.dismiss(animated: animated, completion: block)
        } else {
            block()
        }
        
        self.tabBarController?.selectedIndex = 0
    }
```

I've only tested it with. 
- Adding new account
- Switching accounts
- Logout

Haven't though about any edge cases so not sure if this would cause a problem.

rnystrom:
I lean towards @Huddie's suggestion. It feels like we're making things even more complicated here.

ismetanin:
@Huddie That's great, thanks!

Huddie:
@ismetanin Happy to lend a hand!

Issue #2106, opened 2018-08-13
Wireframe / design profile


rnystrom:

Huddie:
I am not a graphic designer but maybe these can get the discussion sparked here.


![profile stage 1](https://user-images.githubusercontent.com/10662471/44243868-3708dd80-a19f-11e8-86ae-414339791c1d.jpg)

![profile stage 3](https://user-images.githubusercontent.com/10662471/44243869-3a03ce00-a19f-11e8-8a8c-8a1d22259774.jpg)

![profile stage 4](https://user-images.githubusercontent.com/10662471/44243873-3c662800-a19f-11e8-8a71-5c569df49437.jpg)

Huddie:
I'll also point out. I don't really like that bookmarks are the center icon on the tabbar. For some reason I feel it should be profile or inbox. But I wasn't really focused on that

BrianLitwin:
Including the following images here for quick reference against what Github already does. It would be cool to find some fresh sources of inspiration beyond GitHub. 

---------------------------------------------------------------------------------------------------
<img width="356" alt="screenshot 2018-08-15 16 02 06" src="https://user-images.githubusercontent.com/26695477/44244164-8f8caa80-a1a0-11e8-9e84-132b2655e35c.png">
---------------------------------------------------------------------------------------------------
<img width="355" alt="screenshot 2018-08-15 16 02 53" src="https://user-images.githubusercontent.com/26695477/44244169-99161280-a1a0-11e8-9ae0-396c5fa29f21.png">
---------------------------------------------------------------------------------------------------
<img width="353" alt="screenshot 2018-08-15 16 01 49" src="https://user-images.githubusercontent.com/26695477/44244215-cebafb80-a1a0-11e8-900f-014e5154c4a0.png">
---------------------------------------------------------------------------------------------------
<img width="398" alt="screenshot 2018-08-16 22 13 34" src="https://user-images.githubusercontent.com/26695477/44244383-b8616f80-a1a1-11e8-887e-82b2f5c8b0be.png">
---------------------------------------------------------------------------------------------------
<img width="363" alt="screenshot 2018-08-16 22 15 02" src="https://user-images.githubusercontent.com/26695477/44244455-14c48f00-a1a2-11e8-8049-b1439f911410.png">

Huddie:
[...]

PR #2125, opened 2018-08-16
[WIP] prototype for userProfile tab

BrianLitwin:
Trying to just get something on the board that y'all can look at, play with, etc on the Profile Tab. 

Two pressing concerns about this particular commit - 
- how big should the users profile image be, and should it be variable with screenwidth? 
- how do we determine the size of the content in a section controller after all the labels have been set up and communicate the collection view to resize the cell? There will be lots of variation in the content of people's 'basic info' (their bios, etc) and the size of that content... I'm not sure what the pattern is to handle those variable content sizes on the fly w/ IGListKit. 

screenshot 2018-08-16 17 02 07

screenshot 2018-08-16 17 13 11

screenshot 2018-08-16 17 16 43

screenshot 2018-08-16 17 18 12

### DIFF
--- a/Classes/Profile/LocationLabel.swift
+++ b/Classes/Profile/LocationLabel.swift
@@ -1,0 +1,46 @@
+//
+//  LocationLabel.swift
+//  Freetime
+//
+//  Created by B_Litwin on 8/16/18.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import UIKit
+
+final class LocationLabel: UILabel {
+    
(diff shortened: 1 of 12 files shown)


Huddie:
Hey it’s me! Cool! Lol. 

I think Ryan created an issue dedicated to wire framing / designing the profile page #2106 
Looks like a good start but I’d personally wait for the wireframe/design to start taking shape before really going at it.
That being said. Wireframe/design hasn’t seen much movement. I’ll try mocking something up. 

Issue #2105, opened 2018-08-13
Evaluate routing libraries and blog posts


rnystrom:
### Blog Posts

- [ ] [A Flexible Routing Approach in an iOS App](https://medium.com/rosberryapps/the-flexible-routing-approach-in-an-ios-app-eb4b05aa7f52)
- [ ] [URL Routing in iOS apps: Compass Beginner Guide](https://medium.com/flawless-app-stories/url-routing-with-compass-d59c0061e7e2)
- [ ] [Universal links in iOS](https://medium.com/@abhimuralidharan/universal-links-in-ios-79c4ee038272)
- [x] [Coordinator essentials pt 1](https://medium.com/blacklane-engineering/coordinators-essential-tutorial-part-i-376c836e9ba7)
- [x] [Coordinator essentials pt 2](https://medium.com/blacklane-engineering/coordinators-essential-tutorial-part-ii-b5ab3eb4a74)

### Criteria

- **Deep Links:** Does the lib have external "open url" support?
- **Data Params:** Can you provide arbitrary data when navigating?
- **Deep Nav:** Does navigation support arbitrary nesting (not just push a single VC)
- **Tab Selection:** Does the lib allow selection of `UITabBarController`?
- **Split VC:** Does the lib automatically support `UISplitViewController`?
- **Custom Transitions:** Can you do custom VC transitions?

### Legend

- ✅ Supported
- ❌ Not supported
- ⚠️ Supported with manual implementation

### Libraries

Name | Deep Links | Data Params | Deep Nav | Tab Selection | Split VC | Custom Transitions
---|:-:|:-:|:-:|:-:|:-:|:-:|
[Compass](https://github.com/hyperoslo/Compass) |✅|✅|❌|❌|❌|⚠️|
[URLNavigator](https://github.com/devxoul/URLNavigator) |✅|✅|❌|❌|❌|⚠️|
[DeepLinkKit](https://github.com/button/DeepLinkKit) |✅|✅|❌|❌|❌|⚠️|
[routable-ios](https://github.com/clayallsopp/routable-ios) |✅|✅|❌|❌|❌|❌|
[Router](https://github.com/freshOS/Router) |✅|✅|❌|❌|❌|⚠️|
[WAAppRouting](https://github.com/Wasappli/WAAppRouting) |✅|✅|❌|✅|❌|⚠️|
[HHRouter](https://github.com/Huohua/HHRouter) |✅|✅|❌|❌|❌|❌|
[...]

PR #2134, opened 2018-08-19
Adds Action extension PoC

flovilmart:
demo |
---|
gh

Let open github.com url's from Safari into GitHawk.

- Adds simple routing (replacing scheme from https to freetime)
- Inject through JS in web pages a redirection to GitHawk
- Presents an IssuesController based on URL


Huddie:
Well, this is only AWESOME. Super cool!

flovilmart:
Thanks @Huddie 😅

FreetimeAction Extension/ActionRequestHandler.swift
```diff
@@ -0,0 +1,91 @@
...
+        
+        // In this very simple example, the JavaScript will have passed us the
+        // current background color style, if there is one. We will construct a
+        // dictionary to send back with a desired new background color style.
+        guard let location = javaScriptPreprocessingResults["location"] as? String,
+            let url = URL(string: location),
+            let host = url.host,
+            host.contains("github.com") else {
+
+            self.doneWithResults(["error": "Unable to open on GitHawk"])
```
rnystrom: localize? Also "Unable to open **in** GitHawk"

FreetimeAction Extension/ActionRequestHandler.swift
```diff
@@ -0,0 +1,91 @@
...
+
+    var extensionContext: NSExtensionContext?
+    
+    func beginRequest(with context: NSExtensionContext) {
+        // Do not call super in an Action extension with no user interface
+        self.extensionContext = context
+        
+        var found = false
+        // Find the item containing the results from the JavaScript preprocessing.
+        outer:
```
rnystrom: TIL

FreetimeAction Extension/ActionRequestHandler.swift
```diff
@@ -0,0 +1,91 @@
...
+import UIKit
+import MobileCoreServices
+
+class ActionRequestHandler: NSObject, NSExtensionRequestHandling {
+
+    var extensionContext: NSExtensionContext?
+    
+    func beginRequest(with context: NSExtensionContext) {
+        // Do not call super in an Action extension with no user interface
+        self.extensionContext = context
```
rnystrom: nit: `self` necessary?

FreetimeAction Extension/ActionRequestHandler.swift
```diff
@@ -0,0 +1,91 @@
...
+//  FreetimeAction Extension
+//
+//  Created by Florent Vilmart on 2018-08-19.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import UIKit
+import MobileCoreServices
+
+class ActionRequestHandler: NSObject, NSExtensionRequestHandling {
```
rnystrom: `final`

FreetimeAction Extension/ActionRequestHandler.swift
```diff
@@ -0,0 +1,91 @@
...
+        self.extensionContext = context
+        
+        var found = false
+        // Find the item containing the results from the JavaScript preprocessing.
+        outer:
+        for item in context.inputItems as! [NSExtensionItem] {
+            if let attachments = item.attachments {
+                for itemProvider in attachments as! [NSItemProvider] {
+                    if itemProvider.hasItemConformingToTypeIdentifier(String(kUTTypePropertyList)) {
+                        itemProvider.loadItem(forTypeIdentifier: String(kUTTypePropertyList), options: nil, completionHandler: { (item, error) in
```
rnystrom: can store the `String` instead of creating it twice
```swift
let identifier = String(kUTTypePropertyList)
```

Classes/View Controllers/UIViewController+MenuDone.swift
```diff
@@ -10,13 +10,22 @@ import UIKit
...
+                action: #selector(onMenuDone)
+            )
+            navigationItem.leftBarButtonItem?.tintColor = Styles.Colors.Gray.light.color
+        } else {
+            navigationItem.rightBarButtonItem = UIBarButtonItem(
+                barButtonSystemItem: .done,
+                target: self,
+                action: #selector(onMenuDone)
+            )
+            navigationItem.rightBarButtonItem?.tintColor = Styles.Colors.Gray.light.color
```
rnystrom: Lots of duped code here. Can we just capture which bar button item we're operating on and set everything once?

Classes/Systems/AppDelegate.swift
```diff
@@ -93,6 +93,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
 
     func application(_ app: UIApplication, open url: URL, options: [UIApplicationOpenURLOptionsKey : Any] = [:]) -> Bool {
+
+        if let issue = IssueDetailsModel(url: url) {
+            rootNavigationManager.show(issue: issue)
+            return true
+        }
```
rnystrom: Makes me want to get #2105 wrapped up!

Classes/Systems/RootNavigationManager.swift
```diff
@@ -86,6 +86,18 @@ final class RootNavigationManager: GitHubSessionListener {
...
+    public func show(issue: IssueDetailsModel) {
+        guard let client = lastClient else { return }
+
+        if tabBarController?.presentedViewController != nil {
+            tabBarController?.dismiss(animated: false, completion: nil)
+        }
+        let issues = IssuesViewController(client: client, model: issue, scrollToBottom: true)
+        issues.addMenuDoneButton(left: true)
+        let nav = UINavigationController(rootViewController: issues)
+        tabBarController?.present(nav, animated: true, completion: nil)
```
rnystrom: for animation param use `trueUnlessReduceMotionEnabled`

flovilmart:
Thanks for the review :) the ActionRequestHandler is the boilerplate from Xcode, I figured it would need a rewrite if the feature was a go ;) i’ll Do the changes.

As for the routing lib, yes, that would have tremendously helped :)

rnystrom:
Ha! Awesome. I’ve never even seen an action extension before.

rnystrom:
@flovilmart what's the issue w/ the "Open in GitHawk" alert? It's just kind of clunky? Does it happen every time?

I'm not too offended if this is a one-time thing.

flovilmart:
It's a Safari thing, which will occur all the times (unfortunately) as we hack into the window.location API.

The main benefit with the http redirection alternative is that it would remove this issue.

rnystrom:
@flovilmart so what's required to make that work?

- Add some logic to githawk.com to handle redirects
- Add githawk.com deep linking to the app

That's probably it?

flovilmart:
> Add some logic to githawk.com to handle redirects

IIRC, with [universal links](https://developer.apple.com/library/archive/documentation/General/Conceptual/AppSearch/UniversalLinks.html), we don't even need a redirection. If the app is installed, the app should open if deep linking is properly configured server side through the app site association.

So technically we could tell the app site association config to hit on the '/redirect' path, and it would open the app with the proper bundle Id (tough to test though with the current setup as the 'real' bundle ID is owned by you).

I haven't configured those in years, and from memory it's always a hit or miss when developing (lots of stars have to align) once aligned it works usually well.

rnystrom:
Closing as its stale

PR #2142, opened 2018-08-26
Replace "Inbox Zero" functionality and remove Firebase

rnystrom:
- [File](http://githawk.com/holidays.json) can have more dates that let us plan holidays in the year
- Remove Firebase


Sherlouk:
Wow this is pretty cool! One "concern" is that this is highly Americanised, like I can get over it and I don't know what your actual distribution of users is like but GitHub is very world wide.

Something as small as Mother's Day being on different days depending on US vs UK. Maybe a future improvement to specify a country code or something too?

rnystrom:
Ya that’s a great idea! We can make separate files for each locale and append it to the url being fetched.

Side note that this change doesn’t regress anything, localized holidays has always been an issue.

Issue #2097, opened 2018-08-12
Load more consistency


Huddie:
![GitHawk Upload by Huddie](https://i.imgur.com/rGGahQk.jpg)

![GitHawk Upload by Huddie](https://i.imgur.com/yf0BTNr.jpg)

<details>
<summary>Bug Report Dump (Auto-generated)</summary>
<pre>
Version 1.22.0 (1534015300)
Device: iPhone 6s (iOS 11.4.1)
TestFlight: true
</pre>
</details>

Huddie:
I can get this one

BrianLitwin:
looking ahead at #317, there'll be a need for two more "load more" cells in user repositories and starred repositories.. would be really nice to have a standardized cell to drop in.

rnystrom:
@BrianLitwin definitely. That was my original intention. We should be using the same cell for everything. We can do a sweep and delete the ones we don’t want.

Should we stick with the blue or grey? Also we need to use the one with the spinner.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Huddie:
Gray looked disabled IMO

Huddie:
There’s already a loadMore cell problem is that there is an IssueNeckLoadCell or something like that also. I’m gonna just fix up LoadMore to support the needed functions and clear out the IssueLoadMore. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #2159, opened 2018-09-13
Load Consistency

Huddie:


Huddie:
Addresses #2097

Classes/Section Controllers/LoadMore/LoadMoreCell.swift
```diff
@@ -42,14 +49,14 @@ final class LoadMoreCell: UICollectionViewCell {
...
-        didSet {
-            if isSelected {
-                label.alpha = 0.5
-            } else {
-                label.alpha = 1
-            }
+    
+    func configure(loading: Bool) {
+        label.isHidden = loading
+        if loading {
```
jdisho: you can make this shorter

```swift 
loading ? activity.startAnimating() : activity.stopAnimating()
```

Huddie: I actually just merged this from the IssueNeckLoadMore. Personality I don’t feel ternary vs full if-else makes a difference but I can switch it

jdisho: Functionaly is the same, but is shorter.

Huddie:
If someone who has a better overall knowledge of the codebase knows other places where there are load more let me know. 

rnystrom:
Should be used in the base list VC, base list 2 VC, and issues for sure (sorry for short names, on mobile). I _think_ that would cover everything.

Huddie:
Okay thanks! I’ll go ahead and see if those are using their own or using LoadMoreCell 

Huddie:
@rnystrom I think I nabbed them all now. There was one in BaseViewController, nice catch!

PR #2175, opened 2018-09-23
Fetch current detail controller and compare

### LINKED ISSUE #244: Selecting the currently selected issue reloads the detail view.

rob-nash:
This pull request prevents the detail view from being reloaded unnecessarily. Which closes the following.

Closes #244

### DIFF
--- a/Classes/Issues/IssueDetailsModel.swift
+++ b/Classes/Issues/IssueDetailsModel.swift
@@ -15,3 +15,5 @@ struct IssueDetailsModel {
     let number: Int
 
 }
+
+extension IssueDetailsModel: Equatable {}
--- a/Classes/Issues/IssueDetailsModel.swift
+++ b/Classes/Issues/IssueDetailsModel.swift
@@ -8,12 +8,10 @@
 
 import Foundation
 
-struct IssueDetailsModel {
+struct IssueDetailsModel: Equatable {
 
     let owner: String
     let repo: String
     let number: Int
 
 }
-
-extension IssueDetailsModel: Equatable {}
--- a/Classes/Issues/IssuesViewController.swift
+++ b/Classes/Issues/IssuesViewController.swift
@@ -36,7 +36,7 @@ final class IssuesViewController:
     MessageTextViewListener {
 
     private let client: GithubClient
-    private let model: IssueDetailsModel
+    let model: IssueDetailsModel
     private let addCommentClient: AddCommentClient
     private let textActionsController = TextActionsController()
     private var bookmarkNavController: BookmarkNavigationController? = nil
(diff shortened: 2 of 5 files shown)


Classes/Notifications/NotificationSectionController.swift
```diff
@@ -97,6 +101,10 @@ final class NotificationSectionController: ListSwiftSectionController<Notificati
             showRelease(release, model: model)
         }
     }
+    
+    private var currentDetailController: IssuesViewController? {
+        return (UIApplication.shared.delegate as? AppDelegate)?.rootNavigationManager.detailNavigationController?.topViewController as? IssuesViewController
```
jdisho: Would be better to split this up

Classes/Issues/IssueDetailsModel.swift
```diff
@@ -15,3 +15,5 @@ struct IssueDetailsModel {
     let number: Int
 
 }
+
+extension IssueDetailsModel: Equatable {}
```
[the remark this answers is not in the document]
rob-nash: Sure

Classes/Issues/IssuesViewController.swift
```diff
@@ -36,7 +36,7 @@ final class IssuesViewController:
     MessageTextViewListener {
 
     private let client: GithubClient
-    private let model: IssueDetailsModel
+    let model: IssueDetailsModel
```
jdisho: Why did you change the access control? If there is a reason, move it before the `private` properties 😊

Classes/Notifications/NotificationSectionController.swift
```diff
@@ -86,9 +86,13 @@ final class NotificationSectionController: ListSwiftSectionController<Notificati
         case .hash(let hash):
             viewController?.presentCommit(owner: model.owner, repo: model.repo, hash: hash)
         case .number(let number):
+            let model = IssueDetailsModel(owner: model.owner, repo: model.repo, number: number)
+            if let controller = currentDetailController, controller.model == model {
+                return
+            }
             let controller = IssuesViewController(
```
jdisho: Would be better to change the name to `issuesViewController`, because atm there are two controllers with the same name.

Classes/Notifications/NotificationSectionController.swift
rob-nash: Ok

Classes/Issues/IssuesViewController.swift
rob-nash: I did this so that I could reach the property from the `NotificationSectionController`

rob-nash: Have moved

Classes/Notifications/NotificationSectionController.swift
rob-nash: Done

PR #2189, opened 2018-09-26
Merge button status

Huddie:
Shows merge status without button when you cannot merge.
simulator screen shot - iphone xr - 2018-09-26 at 16 14 07

### DIFF
--- a/Classes/Issues/Merge/IssueMergeSectionController.swift
+++ b/Classes/Issues/Merge/IssueMergeSectionController.swift
@@ -17,7 +17,7 @@ ListBindingSectionControllerSelectionDelegate {
     private let model: IssueDetailsModel
     private let client: GithubClient
     private let resultID: String?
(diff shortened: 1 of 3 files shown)


Classes/Issues/Merge/IssueMergeSectionController.swift
```diff
@@ -17,11 +17,18 @@ ListBindingSectionControllerSelectionDelegate {
     private let model: IssueDetailsModel
     private let client: GithubClient
     private let resultID: String?
+    private var mergeCapable = false
```
jdisho: `isMergeCapable` would be better/clearer

Huddie: I was actually thinking about this was gonna ask. isMergeCapable is def more swift style.

Huddie:
@jdisho looks better?

PR #2192, opened 2018-09-27
fix double tap tab invalid

certainly:
this bug is caused by the time interval between two taps which is too small.

### DIFF
--- a/Classes/View Controllers/TabBarControllerDelegate.swift
+++ b/Classes/View Controllers/TabBarControllerDelegate.swift
@@ -28,7 +28,7 @@ final class TabBarControllerDelegate: NSObject, UITabBarControllerDelegate {
             nav.transitionCoordinator?.viewController(forKey: UITransitionContextViewControllerKey.from) == nil,
             let root = nav.viewControllers.first as? TabNavRootViewControllerType {
             tapCount += 1
-            DispatchQueue.main.asyncAfter(deadline: .now() + 0.17, execute: {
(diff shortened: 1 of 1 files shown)


Huddie:
I don't really see this as a bug, works fine for me. Nevertheless Apple's default timeout for double tap in voiceover is 0.25 so maybe 0.25 is a good fit.

certainly:
thanks for the reply.
I tested it on simulator and iPhoneX, and this has nothing to do with voiceover. Just double tap on the search or bookmark tap, see if the inputbar will become the first responder. In my case, 0.17 will reproduce this bug every time，0.25 is ok.

certainly:
Test

Huddie:
@certainly sorry my response was a bit vague. 
I’m using voiceover just as a reference for what the “default” double tap speed should be. 
I don’t think this is a bug, it’s more a matter of preference/ how fast one personally double tap. I was able to active the doubleTap by double tapping normally that’s why I said it worked for me.

certainly:
Wow, that's interesting. Because no matter how hard I try, I just couldn't double tap within 0.17 on my iPhoneX😅

Huddie:
@certainly haha well then +1 for this PR. To be honest I never even knew this was an option until this PR. Anyways sorry for the mixup def would like it to work for everyone!

certainly:
>@certainly haha well then +1 for this PR. To be honest I never even knew this was an option until this PR. Anyways sorry for the mixup def would like it to work for everyone! ...

@Huddie now the only  thing I what to know is if I'm too slow or you're too quick.😃

PR #2193, opened 2018-09-27
Settings cell textColor

Huddie:
Set to custom to match the rest of settings 😀

### DIFF
--- a/Classes/Settings/Settings.storyboard
+++ b/Classes/Settings/Settings.storyboard
@@ -179,7 +179,7 @@
                                                     <rect key="frame" x="298.5" y="13" width="41.5" height="19.5"/>
                                                     <autoresizingMask key="autoresizingMask"/>
                                                     <fontDescription key="fontDescription" type="system" pointSize="16"/>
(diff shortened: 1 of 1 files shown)


viktorgardart:
While you're at it, could you also fix the textColor of the "Double Tap Reaction" value label? 
It's currently black when it says "Off" 👍

Huddie:
@viktorgardart that’s a good point

Huddie:
Any ideas why "Enabled" under Double Tap Reaction is a different font? Everything seems the same to me. 
Cell - StyledTextCell
TextColor - Custom
Font - System 16

rnystrom:
@Huddie nope, can we standardize that one too?

Huddie:
@rnystrom Ya! But I was a bit confused because it seems to be exactly like the others. Same font, cell, color. Not sure why it’s different but I’ll check it out more

Huddie:
I didn't realize they inherited a class called SettingsLabel 😊

Huddie:
@rnystrom Should be good to go

Issue #2177, opened 2018-09-23
Filter icon instead of 3 dots


Huddie:
On the inbox page, the 3 dots (top left) allow for switching to a specific repo or view all. Maybe a filter icon is more appropriate

rnystrom:
I dig it!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Huddie:
So I couldn't find a filter octicons so I went at making it myself.
Anyone have a better filter icon or suggestions to improve this one?

![simulator screen shot - iphone xr - 2018-09-25 at 22 27 56](https://user-images.githubusercontent.com/10662471/46053814-4f5f0700-c112-11e8-8a02-984d5752c93e.png)

Huddie:
My bad filling in the top. We seem to use outlined icons for navBar items

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Huddie:
Take 2
![simulator screen shot - iphone xr - 2018-09-26 at 21 57 56](https://user-images.githubusercontent.com/10662471/46119054-497f2980-c1d7-11e8-9741-58e52cb6c773.png)

Huddie:
Thoughts on improvements, or anyone have an icon ?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rnystrom:
Looks a little sharp on the edges. How about this one?

https://thenounproject.com/search/?q=Filter&i=1956576

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BasThomas:
There’s no icon at all in the latest beta here?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rnystrom:
@BasThomas 😱 your right me neither! Where’d it go? Ha

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BasThomas:
No idea 😛

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Huddie:
@rnystrom I like the icon but I don’t have the ability to download it

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Huddie:
@rnystrom would you be able to add the icon to the project ?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Huddie:
[...]

PR #2203, opened 2018-09-30
3 dot icon

Huddie:
I think the filter icon @rnystrom showed from noun project is good so this is unnecessary if we go with that one. But until then this fixes a mistake where the image name was changed in one of my merges.


Huddie:
#2177 

Issue #2199, opened 2018-09-29
Bottom border of PR status box missing


j-f1:
When I don’t have push access, the bottom border doesn’t show up.

![GitHawk Upload by j-f1](https://i.imgur.com/Z6gufmW.jpg)

<details>
<summary>Bug Report Dump (Auto-generated)</summary>
<pre>
Version 1.23.0 (1538176803)
Device: iPhone 8 (iOS 12.0)
TestFlight: true
</pre>
</details>

rnystrom:
@Huddie from the merge change

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Huddie:
I’m on it

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Huddie:
I think this should be able to be closed, but totally get waiting for the next beta to confirm

PR #2204, opened 2018-09-30
Fixed cell border on merge button

Huddie:
The issue before was caused by the fact that we never had less than 2 cells for the mergeSectionController to render. (Summary + Merge Button) 
Our border enum had 3 options, head, tail, neck which works well for 2+ cells (Head + Tail) but with only 1 we had issues (Just summary cell since the button is not added).

I added .full to the enum and checked if the head = the tail.

### DIFF
--- a/Classes/Views/CardCollectionViewCell.swift
+++ b/Classes/Views/CardCollectionViewCell.swift
@@ -9,54 +9,55 @@
 import UIKit
 
 class CardCollectionViewCell: UICollectionViewCell, UIGestureRecognizerDelegate {
-
+    
     enum BorderType {
         case head
         case neck
         case tail
+        case full
     }
-
+    
     var border: BorderType = .neck {
         didSet { setNeedsLayout() }
     }
-
+    
     private let borderLayer = CAShapeLayer()
     private let backgroundLayer = CAShapeLayer()
-
+    
     override init(frame: CGRect) {
         super.init(frame: frame)
-
+        
         contentView.clipsToBounds = true
-
+        
         // insert above contentView layer
         borderLayer.strokeColor = UIColor.red.cgColor
         borderLayer.strokeColor = Styles.Colors.Gray.border.color.cgColor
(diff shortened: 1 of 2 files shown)


Huddie:
#2199

Classes/Views/CardCollectionViewCell.swift
```diff
@@ -9,54 +9,55 @@
 import UIKit
 
 class CardCollectionViewCell: UICollectionViewCell, UIGestureRecognizerDelegate {
-
+
```
rnystrom: @Huddie can you update your Xcode so these don't keep popping up?

screen shot 2018-09-30 at 12 02 45 am

Classes/Views/CardCollectionViewCell.swift
```diff
@@ -94,23 +95,65 @@ class CardCollectionViewCell: UICollectionViewCell, UIGestureRecognizerDelegate
...
             )
             borderPath.addLine(to: CGPoint(x: pixelSnapBounds.maxX, y: bounds.minY))
-
+            
+            fillPath = borderPath.copy() as! UIBezierPath
+            fillPath.close()
+        case .full:
+            
+            borderPath.move(to: CGPoint(x: pixelSnapBounds.minX,
+                                        y: bounds.maxY - cornerRadius))
```
rnystrom: nit: can you follow the indent and newline style of the previous `case`s?

Classes/Views/CardCollectionViewCell.swift
Huddie: So that's how you get rid of that!
@rnystrom Thanks!

Classes/Views/CardCollectionViewCell.swift
Huddie: Ya makes sense

Huddie:
Should be better. If anyone can think of situations where there would still be an issue with the border let me know.

Issue #2140, opened 2018-08-23
Move the “lock” button to below the divider line


j-f1:
**Is your feature request related to a problem? Please describe.**
When clicking on the hamburger button to modify the issue, the buttons are grouped into two groups: the close option and everything else:

<img src="https://user-images.githubusercontent.com/25517624/44527566-37303e00-a6b5-11e8-8a2f-28c867b1b571.jpeg" width="200">

**Describe the solution you'd like**
I suggest this divider be moved up one spot, above the lock button. This would divide the options into ones that prompt for more information (above the line) and ones that take action immediately (lock/close).

**Describe alternatives you've considered**
We could leave it as-is, but there’s currently no indication that locking the issue will lock it right away instead of prompting for confirmation, like all similarly-styled options do.

Huddie:
I agree

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

kokkimphal:
@````
~~~~
````

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

kokkimphal:
<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rnystrom:
I’m down with this. Lift gray would be good for the icon. Easy issue for someone to pick up.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Huddie:
This issue should be able to be closed

Huddie:
@j-f1 really?! That’s cool. Thank you

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BasThomas:
@Huddie https://help.github.com/articles/closing-issues-using-keywords/

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

j-f1:
I think the requirement is that either you or the person merging the PR has the ability to close the issue (although I’m not sure about the former).

Huddie:
@j-f1 Well, thanks for the tip. I never knew that.

PR #2206, opened 2018-09-30
Move lock

Huddie:
Moves lock under divider.

Lock always has the divider above it assuming your a collaborator.
If you aren't a collaborator (Lock does not show), close has the divider above it.

Lock's color was switched to light gray.


Huddie:
I think it looks a bit weird that lock and close are colored and unlocks and reopen are blue like the rest. Not sure what colors unlock and reopen should be but maybe. Light gray for unlock (just like lock) and green for reopening.

Huddie:
#2140

Issue #2214, opened 2018-09-30
Try again when loading fails


Huddie:
This is what can happen if you lack a network (see imagine) when trying to view an issue (similar view would appear with many other sections) 
I happen to know to pull down the view to cause the pull to refresh to occur but someone else just looking at this view may not think they can pull to refresh.

Maybe we can add a “Try again” or “Refresh” button beneath the “Issue cannot be found” wording. That way people can click it to start a refresh. 
 ![GitHawk Upload by Huddie](https://i.imgur.com/PcRQWN4.jpg)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rnystrom:
This would be great. Much better UX.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

cameronmoreau:
How does something like this look? I'll follow up with a PR soon. 

![img_1927](https://user-images.githubusercontent.com/732722/46390661-bdb14580-c68d-11e8-84c3-aeb6cc255eb0.PNG)

Huddie:
Looks nice! I think the current GitHawk style would be more of just "Try Again" in blue, no background (See "Load More" cells and "Review Github Access", which don't have backgrounds to them).

kngu9:
@cameronmoreau I think "Cannot load README" defines a system specific behavior, like there might or might not contain a README, but unable to parse it.

Maybe specify that there is not a README for the repository?

rnystrom:
Exactly what @Huddie said 😄

BasThomas:
This is there now 👍

PR #2226, opened 2018-10-04
Add "Try Again" button for EmptyView (#2214)

cameronmoreau:
This pr adds the proposed feature from #2214 

When loading fails a "Try Again" button apears right under, tapping this button does the equivelent of "Pulling to Refresh"

**Changes**
- Add "Try Again" button to `EmptyView`
- Add `EmptyViewDelegate` for retry button taps
- Add delegate and show retry button for every class using `EmptyView`

Here's a screenshot
simulator screen shot - iphone 8 - 2018-10-03 at 18 34 30

### DIFF
--- a/Classes/Issues/IssuesViewController.swift
+++ b/Classes/Issues/IssuesViewController.swift
@@ -34,6 +34,7 @@ final class IssuesViewController:
     FlatCacheListener,
     IssueCommentSectionControllerDelegate,
     IssueTextActionsViewSendDelegate,
+    EmptyViewDelegate,
     MessageTextViewListener {
 
     private let client: GithubClient
@@ -501,6 +502,8 @@ final class IssuesViewController:
         case .idle:
             let emptyView = EmptyView()
             emptyView.label.text = NSLocalizedString("Issue cannot be found", comment: "")
+            emptyView.delegate = self
+            emptyView.button.isHidden = false
             return emptyView
         case .loading, .loadingNext:
             return nil
@@ -613,6 +616,12 @@ final class IssuesViewController:
                 body: text
             )
         }
+    }
+
+    // MARK: EmptyViewDelegate
+
+    func didTapRetry() {
+        self.feed.refreshHead()
     }
 
     // MARK: MessageTextViewListener
--- a/Classes/PullRequestReviews/PullRequestReviewCommentsViewController.swift
+++ b/Classes/PullRequestReviews/PullRequestReviewCommentsViewController.swift
@@ -16,6 +16,7 @@ final class PullRequestReviewCommentsViewController: MessageViewController,
     ListAdapterDataSource,
     FeedDelegate,
     PullRequestReviewReplySectionControllerDelegate,
+    EmptyViewDelegate,
     IssueTextActionsViewSendDelegate {
 
     private let model: IssueDetailsModel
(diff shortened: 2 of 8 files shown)


Huddie:
Could we add a bit of space between the message (Cannot load issues) and button (Try Again)?

Huddie:
Also I can see a scenario where no issues exist but it loaded properly (Like a new repo!). So "Try Again" wouldn't make much sense, but users should still have an intutive way to refresh the list. Can the text be changed based on the state of the list 
- Error -> "Try Again"
- Empty ->" Refresh"

Maybe something like that?

cameronmoreau:
For empty scenarios, it looks like currently `RepositoryEmptyResultsType` and `InitialEmptyView` display a message and icon such as "There aren't any issues." or "Add a bookmark" with an emoji. Do you suggest adding a similar button("Refresh") for these as well? I believe only "failed to load" scenarios are using an `EmptyView`.

Sure! This changes the margin from 15 to 22, how is this looking? More? Less?
simulator screen shot - iphone 8 - 2018-10-03 at 19 16 35

Huddie:
@cameronmoreau Looks good with spacing, can't tell if the font size should be smaller or not.
My personal belief is every view with pull to refresh should have a button to activate it when empty. Users need to know that you can refresh the view and people don't really think about pulling down empty views

cameronmoreau:
Hmm, yeah totally fine adding that functionality, it looks like screens that display `RepositoryEmptyResultsType` have a pull to refresh, unlike the `InitialEmptyView`, so I could definitely add the button there.

Personally, I've always felt like it really comes down to a UX vs UI and what kind of user you have. For instance "Hamburger Menu" vs "Tab Menu" layout. Tabs are generally much more intuitive but offer limited space, I feel similar arguments apply to "pull to refresh" vs a button vs both. [Neat article on the subject](https://uxplanet.org/pull-to-refresh-ui-pattern-42a85f671cdf), so I guess finding a nice balance or sticking to a pattern? Lots of things to consider but let me know, happy to the "Refresh" button in a follow-up commit!

Button using `small` instead of `preferred` font size. Agreed that smaller looks a bit better, it's pretty subtle but nice change 😎
simulator screen shot - iphone 8 - 2018-10-03 at 19 52 02

Huddie:
Ill check out the article. My personal feeling is that adding a “refresh” or “try again” does not harm UI at all (personally I actually like the UI better because it has a friendlier UX) and it improves UX.
If a user doesn’t know how to refresh an empty list, we aren’t doing a good enough job in UX. A user could create his/her first issue online and wouldn’t know how to refresh the GitHawk app without going out of the view and coming back in

cameronmoreau:
Ahh, completely true!

One more thought is to add a more subtle "Pull to Refresh" or "Pull to Retry" text hint, rather than a button these screens. Thoughts?

Huddie:
@cameronmoreau could work. Personally something feels weird to me about pulling to refresh a blank view. Not sure why, just my brain needs there to be cells for pull to refresh to visually make sense. Otherwise I think the ptr indicator is appears from nowhere. Anyways I’d wait to hear from @rnystrom on the button or “pull to refresh” choice. 

Huddie:
Everything  looks really cool though !

rnystrom:
Quick question, what's the actual UX after tapping the button?

Issue #2161, opened 2018-09-13
“Copy name” and more actions when browsing code


rnystrom:
- ... button in right nav when browsing dir 
- Replace right bar nav with ... button when viewing code
- Add following options:
  - “Copy path”
  - “Copy name”
  - “Share URL”
  - “Share content” (field only)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ashleyng:
I can start work on this. and assuming you meant `files only` instead of `field only`?

Issue #1958, opened 2018-07-20
Bookmark icon should be right-aligned


rnystrom:
Too much space between the icons

![GitHawk Upload by rnystrom](https://i.imgur.com/duCrcT2.jpg)

<details>
<summary>Bug Report Dump (Auto-generated)</summary>
<pre>
Version 1.22.0 (1532006598)
Device: iPhone X (iOS 11.4)
TestFlight: true
</pre>
</details>

ashleyng:
I was taking a stab at this, and should this only affect the `IssueViewController` or should it also include the icon on the `RepositoryViewController`?

rnystrom:
@ashleyng both please! Mind knocking out the repo VC too?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ashleyng:
yep! I can do that!

BasThomas:
Fixed in #2255 right?

PR #2237, opened 2018-10-06
Adds more share actions when browsing a repository (#2161)

ashleyng:
Adds in the feature for more share options when browsing a repository (#2161)
- Adds the `...` when viewing a directory
- Changes the share button to `...` when viewing a file
- Adds "Copy Path", "Copy Name", "Share URL", and "Share Content" options
- each option will open the share sheet giving you the option to share with a different app or just copy it to the clipboard.

For a file `Classes/Issues/Files/IssueFileCell.swift`
- Copy path: `Classes/Issues/Files/IssueFileCell.swift`
- Copy name: `IssueFileCell.swift`
- Share URL: `https://github.com/GitHawkApp/GitHawk/blob/master/Classes/Issues/Files/IssueFileCell.swift`

For a directory `Classes/Issues/Files`
- Copy path: `Classes/Issues/Files`
- Copy Name: `Files`
- Share URL: `https://github.com/GitHawkApp/GitHawk/tree/master/Classes/Issues/Files`

simulator screen shot - iphone 8 - 2018-10-06 at 12 00 54 simulator screen shot - iphone 8 - 2018-10-06 at 12 00 58

### DIFF
--- a/Classes/Utility/AlertAction.swift
+++ b/Classes/Utility/AlertAction.swift
@@ -21,6 +21,24 @@ struct AlertAction {
     let title: String?
     let style: UIAlertActionStyle
 
+    enum AlertShareType {
+        case shareUrl
+        case shareContent
+        case shareFilePath
+        case shareFileName
+        case `default`
+
+        var localizedString: String {
+            switch self {
+            case .shareUrl: return NSLocalizedString("Share URL", comment: "")
+            case .shareContent: return NSLocalizedString("Share Content", comment: "")
+            case .shareFilePath: return NSLocalizedString("Copy Path", comment: "")
+            case .shareFileName: return NSLocalizedString("Copy Name", comment: "")
+            default: return NSLocalizedString("Share", comment: "")
+            }
+        }
+    }
+
     // MARK: Init
 
     init(_ builder: AlertActionBuilder) {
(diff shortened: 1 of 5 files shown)


rnystrom:
This is so good! Thank you!!

Classes/Utility/AlertAction.swift
```diff
@@ -21,6 +21,24 @@ struct AlertAction {
...
+        case shareFileName
+        case `default`
+
+        var localizedString: String {
+            switch self {
+            case .shareUrl: return NSLocalizedString("Share URL", comment: "")
+            case .shareContent: return NSLocalizedString("Share Content", comment: "")
+            case .shareFilePath: return NSLocalizedString("Copy Path", comment: "")
+            case .shareFileName: return NSLocalizedString("Copy Name", comment: "")
+            default: return NSLocalizedString("Share", comment: "")
```
BasThomas: should this be `case .default` instead?

BasThomas:
Awesome stuff @ashleyng!

Classes/Utility/AlertAction.swift
ashleyng: I figured it would be a good fallback if there were any other new sharing options to be added.

ashleyng: That makes sense. I’ll be making another PR soon for #1958 and can add the change for this in there

PR #2241, opened 2018-10-08
New routing library and refactor Shortcuts

rnystrom:
Really excited on where this landed!

This routing architecture consists of multiple parts:

- `Routable` protocol conforming objects describe a route
  - Path, init from params, and encode params
  - Can be encoded and decoded to/from `URL` for eventual push notification and deep links!
- `RoutePerformable` conforming objects simply handle a route
- `AppController` has a map of `String` (path) to `Routable & RoutePerformable`
  - Give the controller a `URL` or path+params to handle a route

> _But why separate `Routable` and `RoutePerformable`?_

Once this is proven, I'm going to move all of the `Routable` stuff into its own Pod. Why? Eventually other apps will be able to install the pod and open GitHawk deep links with strongly typed routes! 🎉 

Separating them lets GitHawk.app do all of the handling internally.

### Shortcuts

This PR also contains a refactor for shortcut handling. It's a bit cleaner and pretty basic. The nice part is that it couples the router to the shortcut item, removing the possibility of tab index changes breaking or crashing the app.

Notice I also added a `UITabBarController` helper to select tabs based on the content view controller type.


BasThomas:
Looks pretty lightweight, nice!

Issue #2244, opened 2018-10-08
When assigning people, put self as first option


rnystrom:
Similar to bumping the default branch

<details>
<summary>Bug Report Dump (Auto-generated)</summary>
<pre>
Version 1.23.0 (1538970418)
Device: iPhone X (iOS 12.0)
TestFlight: true
</pre>
</details>

PR #2246, opened 2018-10-08
PeopleVC: Sort users with self first

BrianLitwin:
#2244
- moves sorting logic from fetch(page:) into type method
- unit test

### DIFF
--- a/FreetimeTests/SortUsers.swift
+++ b/FreetimeTests/SortUsers.swift
@@ -1,0 +1,52 @@
+//
+//  SortUsers.swift
+//  FreetimeTests
+//
+//  Created by B_Litwin on 10/8/18.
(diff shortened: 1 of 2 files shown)


rnystrom:
@BrianLitwin you're on 🔥

FreetimeTests/SortUsers.swift
```diff
@@ -0,0 +1,52 @@
...
+            siteAdmin: false,
+            type: .user
+        )
+    }
+    
+    func test_SortWithCurrentUserFirst() {
+        
+        var users = [
+            "A_rnystrom",
+            "B_basthomas",
```
BasThomas: 😁

BasThomas:
Good stuff!

PR #2254, opened 2018-10-09
Store last search term to avoid repetitive queries

cameronmoreau:
While using the app, I noticed that the search page will make an API call to search for a term, even if the term has already been queried. No bug here, just an opportunity for some UX improvements.

**Example:**
1. User searches "GitHawk"
1. API called, results load and appear
1. User hits "search" on the keyboard (Possibly just to get the keyboard out of the way)
1. API called, the same results load and appear

This PR only avoids this "repetitive search" for the last queried term. The last term will be stored and if the user searches the same term, the function will just escape before the API call.

This is a pretty naive implementation so feel free to review! Didn't want to dig too deep if this wasn't a wanted feature.

Other popular apps such as Instagram and Twitter seem to cache all search terms and their results. This functionality could definitely be implemented if there's a desire for it.

**Advanced Example:**
1. User searches "Git"
1. API called, results load and appear
1. User types "Hawk" (Full-term now "GitHawk")
1. API called, results load and appear
1. User deletes "Hawk"
1. App will load results from the cache instead of querying the API

### DIFF
--- a/Classes/Search/SearchViewController.swift
+++ b/Classes/Search/SearchViewController.swift
@@ -26,6 +26,7 @@ SearchResultSectionControllerDelegate {
     private var recentStore = SearchRecentStore()
     private let debouncer = Debouncer()
     private var keyboardAdjuster: ScrollViewKeyboardAdjuster?
+    private var lastSearchTerm: String?
 
     enum State {
         case idle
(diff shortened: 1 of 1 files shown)


Classes/Search/SearchViewController.swift
```diff
@@ -114,25 +115,29 @@ SearchResultSectionControllerDelegate {
...
         case .success(_, let results):
+            self.lastSearchTerm = term
             self.state = .results(results)
         }
         self.update(animated: animated)
     }
 
     func search(term: String) {
+        if term == self.lastSearchTerm { return }
+        self.lastSearchTerm = nil
```
rnystrom: Can we drop the `self.` in here?

rnystrom:
Closing as its stale

PR #2257, opened 2018-10-10
Add steps in the Setup.md file.

Maxador:
When following the steps for the setup of the project. The user is given steps to follow that could result in a leak of environment variables in the commit history. While those variables are not critical, I think it would show good development practices to people (especially people with not much experience with `git`).


Huddie:
@Maxador Just a heads up. Your commit contains a bunch of files that I’m sure you didn’t mean to commit. Some of which change the bundle ID. You should try removing all files but setup.md 

Maxador:
Well that is kinda ironic, I will fix this in a new pull request.

PR #2278, opened 2018-10-13
Add new iPhone models to UIDevice+Model.swift

MichaelLutaaya:
This PR updates `UIDevice+Model.swift` to include iPhone XS, iPhone XS Max, and iPhone XR.

The identifiers were retrieved from <https://ipsw.me/otas>.


rnystrom:
Thank you!!

Issue #1942, opened 2018-07-19
Add “clear” button on all menus


rnystrom:
Labels, milestone, people

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Maxador:
I would like to tackle this issue. Is there more indication on what the “clear” button should look like and how can I access those views with some relevant data?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BrianLitwin:
@Maxador you do need permissions to get the [ContextMenu](https://github.com/GitHawkApp/GitHawk/blob/master/Classes/Issues/IssueManagingContextController.swift) to show up on the screen. For a quick hack if you just want to mess with it/ see what it looks like, return `[ .labels, .milestone, .assignees ]` [in this method](https://github.com/GitHawkApp/GitHawk/blob/b70e2649a01d3aa8f0ad93ad60a303b666aa1450/Classes/Issues/IssueManagingContextController.swift#L83) and you'll get the button to show up on Issues. 

Prob want the clear button to look similar to the current done button, possibly in the upper left hand corner. Although, currently Assignees has a leftHand nav bar item so maybe others have ideas about what the UI should look like? 

The done button is added in viewDidLoad() of each VC through [this method](https://github.com/GitHawkApp/GitHawk/blob/master/Classes/View%20Controllers/UIViewController+MenuDone.swift), so you can prob look there for styling cues. 

I could be wrong but I don't thinkkk you need to worry about sending any GraphQL mutations or network requests after clear is pressed or the context menu is dismissed, but you probably do need to re-set the selection state of the SectionControllers and update the UI.

rnystrom:
@BrianLitwin 🔥

PR #2288, opened 2018-10-14
Add clear button to action menus

Maxador:
This add a clear button as the left nav bar item for the `PeopleViewController`, `LabelsViewController` and the `MilestonesViewController`.
To keep the selected count in the `PeopleViewController`, I moved the counter in the title next to the "Assignees" or "Reviewers" title text.

#2288


Huddie:
@Maxador a screenshot of what it looks like would be helpful 

Classes/Labels/LabelsViewController.swift
```diff
@@ -10,7 +10,13 @@ import UIKit
 import IGListKit
 import Squawk
 
-final class LabelsViewController: BaseListViewController2<String>, BaseListViewController2DataSource {
+protocol LabelsViewControllerDelegate {
+    func didSelect()
+}
```
rnystrom: Can this protocol be removed? Looks unused.

Classes/People/PeopleViewController.swift
```diff
@@ -90,17 +109,16 @@ PeopleSectionControllerDelegate {
...
-        label.backgroundColor = .clear
-        label.textColor = Styles.Colors.Gray.light.color
-        label.text = "\(selected.count)/\(selectionLimit)"
-        label.sizeToFit()
-        navigationItem.leftBarButtonItem = UIBarButtonItem(customView: label)
+    private func updateTitle() {
+        let selectedCount = "\(selected.count)/\(selectionLimit)"
+        switch type {
+        case .assignee: title = "\(NSLocalizedString("Assignees", comment: "")) \(selectedCount)"
+        case .reviewer: title = "\(NSLocalizedString("Reviewers", comment: "")) \(selectedCount)"
```
rnystrom: I think we should keep the selection count formatting b/c there is a maximum for selected people (10)

Classes/Milestones/MilestonesViewController.swift
```diff
@@ -55,6 +59,27 @@ MilestoneSectionControllerDelegate {
         super.viewDidLoad()
         navigationController?.navigationBar.titleTextAttributes = [.foregroundColor: UIColor.white]
         addMenuDoneButton()
+        addMenuClearButton()
+    }
+
+    func addMenuClearButton() {
+        navigationItem.leftBarButtonItem = UIBarButtonItem(
+            title: NSLocalizedString("Clear", comment: ""),
```
rnystrom: Can we add this string to the `Constants.Strings` list since its used multiple times here?

Classes/Labels/LabelsViewController.swift
Maxador: Yes it is an artifact of my development, I will remove it.

Maxador:
Here are the screenshots:
img_1825
img_1824

The counter fits in the title for the `PeopleViewController` but it is a little tight. Since it is a really small view, I don't know if adding some kind of header or footer (or even a toolbar) to always show a clear button or the counter is a good idea.

One idea I had is to show a `Clear?` label when we tap on the counter (on top left like before). And if we tap again, the clear happen. Otherwise, after a short delay, the counter show up again. While this clear some space in the view, it makes the feature less discoverable.

rnystrom:
Ohh I see, duh, the left bar item is used for the clear button. Hmmm. Ok maybe we just go with putting it in the title for now.

Maxador:
I added the constants and fixed a bug I noticed. The Clear button state (enable or not) and the `PeopleViewController` title wasn't set to the correct initial selection value.

Maxador:
I think we can close #1942 now.

PR #2292, opened 2018-10-15
Enable SwiftLint and fix issues

rnystrom:
I'm preparing to do some more automation with the repo to speed up PR review and unify code formatting. First step is to enable SwiftLint to take care of basic issues.


Huddie:
@rnystrom ever considered [SwiftFormat](https://github.com/nicklockwood/SwiftFormat)

rnystrom:
Ya but seems like this is more mature, and using both would be pretty redundant right?

Huddie:
Ya ones a reminder and one just formats it all for you. So having both would essentially but just having SwiftFormat (unless you configured them differently)

You can place swiftFormat in bitrise so that it formats it according to a given standard whenever a PR is created but not on the local machines. So it gives devs to program however they desire but the product that gives merged will always be according to GitHawk style guide 

rnystrom:
@Huddie any pointers on how how to make that all work? I actually don’t mind making people address the linter vs me having to keep merging PRs for formatting 😛

Really just want to speed up PR turn around and spend less time nitpicking.

Huddie:
@rnystrom sure. I’ll try getting it up and running. 

Huddie:
@j-f1 You type everything on 1 line? Do you mind giving some more background? Are you suggesting prettier for GitHawk or just your experience with a formatter such as SwiftFormat?

I like the idea of SwiftFormat in Bitrise more than running each time you build simply because people program with different spacing etc and no need to force a style on you while programming. SwiftFormat in bitrise should format the code once you request a PR so that when @rnystrom  reviews it, it’s perfectly formatted according to GitHawk standards. 

I’ll admit, one downside of having it run on Bitrise is that each PR needs to be tested. Not sure how many tests we get to run since I’ve seen PRs without tests 

j-f1:
@Huddie I’m just providing my experience with formatters in general. Usually the difficulty I have with code styles is remembering to follow standards that are different from my own, but the differences aren’t as noticeable when reading the code. I’d suggest not auto-fixing issues in CI since that would add the noise of another commit to each PR.

Huddie:
@j-f1 🤔 From my experience:
Having it format on builds doesn’t allow you to program how you typically do (cause every build reformats) + its slows down the compile time 
And having swift lint is not an issue if there aren’t many issues but:
- Not allowing builds because of formatting is a PAIN 
- Warnings work but don’t enforce anything. people will ignore them. So you’ll have to report them in the CI or something which is fine I suppose. 
I don’t really see the downside of having the CI commit a formatting change. @j-f1  Thoughts?

j-f1:
The editor integrations I’ve used typically reformat the file when you save. This ensures you follow the style guide while not throwing errors in your face when you don’t. Also, it doesn’t slow builds down because you only format the one file you just saved and that’ll probably happen faster than you can start the build.
I agree that lint issues shouldn’t stop you from building — as long as the code is syntactically correct, it should run.

Huddie:
@j-f1 I’m not a fan of having code format every time I hit save tbh. i try to save often and code shifting each time can just feel weird. 
Again, programming styles differ and no need for the code to auto-format each and every save. Feels redundant and also just messes with programming “flow?”. 
I’d be more on board with a linter than forcing format each save. Thoughts?

Huddie:
@rnystrom regardless of swiftLint vs swiftFormat, I think it's important we get a https://github.com/danger/danger file in the project. 

Reasons:

- If swiftFormat occurs in the cloud
    - Easier with danger and allows for reporting back
- If swiftLint
    - We need a danger file to report the formatting issues when PR are created. Simply having swiftLint in the project does not make PR review easier because you can't see the lint warnings.

This will also be beneficial for advancing PR Reviews in other aspects

Side Point: Not sure PR thread is the best place to be discussing this :p

rnystrom:
Totally down with adding danger!

Huddie:
@rnystrom Thanks! I can set up the danger file with the longer/formatter once the dangerfile and not are in the repo 

BasThomas:
@Huddie @rnystrom fyi we have a ticket open to add Peril, which would mean we wouldn't need Danger: https://github.com/GitHawkApp/GitHawk/issues/1681

Huddie:
Oh well if you go with Peril I don’t think I can help out. @BasThomas Peril is server side Danger right? 

BasThomas:
Yep!

BasThomas:
It would require a server, but can be hosted on something like Heroku. See their [setup guide](https://github.com/danger/peril/blob/master/docs/setup_for_org.md).

rnystrom:
Alrighty, making a call to land this and start using SwiftLint so we can at least start having some standards in place. If we want to swap to SwiftFormat, that's fine too. Want to at least get this in before conflicts become a nightmare!

Huddie:
Sounds 😎 

PR #2294, opened 2018-10-15
Reduce bottom inset when manage button hidden

### LINKED ISSUE #2229: Extra space at bottom of pull request status view
Screenshot:

MichaelLutaaya:
## What This Does

This fixes #2229. Before this fix, the bottom inset for the collection view was the same whether or not the manage button was visible. After this fix, the bottom inset is smaller when the manage button is hidden.

**Before**
before

**After**
after

## How to Test

1. Visit an issue where the manage button is hidden.
    1. Once loaded, confirm that the inset and all other views appear correctly.
    2. Rotate the device. Confirm that the inset and all other views appear correctly.
    3. Pull to refresh. Confirm that the inset and that all other views appear correctly.
2. Repeat the above for an issue where the manage button is visible.

### DIFF
--- a/Classes/Issues/IssuesViewController.swift
+++ b/Classes/Issues/IssuesViewController.swift
@@ -44,13 +44,16 @@ final class IssuesViewController:
     private var bookmarkNavController: BookmarkNavigationController? = nil
     private var autocompleteController: AutocompleteController!
     private let manageController: IssueManagingContextController
-    private let threadInset = UIEdgeInsets(
-        top: Styles.Sizes.rowSpacing / 2,
-        left: Styles.Sizes.gutter,
-        bottom: 2 * Styles.Sizes.rowSpacing + Styles.Sizes.tableCellHeight,
-        right: Styles.Sizes.gutter
-    )
-
+    private var threadInset: UIEdgeInsets {
+        let insetForManageButton = manageController.manageButton.isHidden ? 0 : Styles.Sizes.tableCellHeight
+        return UIEdgeInsets(
+            top: Styles.Sizes.rowSpacing / 2,
+            left: Styles.Sizes.gutter,
+            bottom: 2 * Styles.Sizes.rowSpacing + insetForManageButton,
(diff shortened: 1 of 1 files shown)


Classes/Issues/IssuesViewController.swift
```diff
@@ -238,6 +241,9 @@ final class IssuesViewController:
                 y: messageView.frame.minY - manageButtonSize.height - Styles.Sizes.gutter),
             size: manageButtonSize
         )
+
+        // Recalculate the bottom content inset since the manageButton‘s visibility may have changed.
+        feed.collectionView.contentInset.bottom = threadInset.bottom
```
rnystrom: I've been a little burned by mucking w/ insets when layout happens. Also the inset method requires that `isHidden` be set _before_ creating a new inset, which is done in `IssueManagingContextController`, so now we've created a hidden dependency. Really easy to break!

Maybe it should just be dependent on the `manageButton`'s height instead?

MichaelLutaaya: Thanks for the feedback, @rnystrom. I agree with you about the hidden dependency and would like to find a way of avoiding it. 🙂

I'm not sure I understand your suggestion of making this depend on the button's height though. Right now, the height doesn't change when the button is hidden so relying on that to calculate the content inset wouldn't work (see line above where the size is set manually).

rnystrom: Err ya you're right. Ignore me! Ha.

j-f1: How about putting a transparent view with the appropriate height at the bottom of the scroll view?

rnystrom: @j-f1 and have it's height be determined by if the button is displayed?

j-f1: Yep!

Huddie:
@MichaelLutaaya What's the status on this?

rnystrom:
Closing as its stale

Issue #2263, opened 2018-10-10
Filter by label (Repo Issue List)


Huddie:
Currently you cannot tap labels in the repoIssueSectionList but can once you tap into a specific issue (it will redirect you)

When tapping on a label in Repo Issue List:
I think we should just add is:”\(label-name)” to the search so that it filters by that label

When tapping on a label inside a specific issue:
We should open the repo issue list in app and again append is:”\(label-name)”

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BrianLitwin:
makes sense @Huddie. I use the label tapping feature a lot on GitHub web to look for "help wanted" issues. Are you taking this? I'd be happy to work on it if you're not.

Huddie:
@BrianLitwin I’m a bit busy with school these next few days so I won’t be able to get to it. If you want to knock it out that would be awesome! 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BrianLitwin:
From Issue: 
&nbsp;
![ezgif com-video-to-gif 2](https://user-images.githubusercontent.com/26695477/46989536-6bc7e100-d0cb-11e8-9665-4b278088fda1.gif)

From deeper in view controller stack: 
&nbsp;
![ezgif com-video-to-gif 3](https://user-images.githubusercontent.com/26695477/46989625-b6495d80-d0cb-11e8-8b0d-151e005b9260.gif)

Huddie:
@BrianLitwin 
1. Looks awesome 
2. This also works on the repoIssueSectionController? 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BrianLitwin:
> This also works on the repoIssueSectionController?

@Huddie I'm going to PR this w/o that feature ->  These labels are nested pretty deep in views and I want to make sure the communication patterns I'm using meet expectations first. Then adding that in will be easier.

Huddie:
@BrianLitwin @rnystrom  I think clicking labels in repoIssueSectionController is the essential one just because that's where users are searching for issues and would want to to filter that search. Less likely someone will be nested deep somewhere are want to search by the label. 
I know at first I suggested we redirect to search when clicking a label outside the repoIssueSectionController but I think we should keep with Github practice on this one:

**Inside repoIssueSectionController** -> Adds label:xxxxx to the search bar
**Anywhere else** -> Shows label description or if one is not provided shows the label name.

[...]

PR #2309, opened 2018-10-18
Search for labels in-app 

BrianLitwin:
Connected to #2263

In most files, the changes are adding a `GithubClient` parameter so that it can be passed to the `LabelDetails` model, which is passed along to init and present a `RepositoryIssuesViewController`. 

Also, I removed the [RepositoryDetails](https://github.com/GitHawkApp/GitHawk/blob/master/Classes/Repository/RepositoryDetails.swift) parameter from the `RepositoryIssuesViewController` initializer and replaced it with `owner` and `name`. This was so that we didn't have to pass around a reference to `RepositoryDetails` as well, and also because we were only using the `name` and `owner` fields of `RepositoryDetails` anyway. 

### vid:

ezgif com-optimize

<br></br> 

I've checked manually that it works when labels are embedded in other contexts as well. Only place it doesn't work is on the main RepoIssuesSectionController list, which as @Huddie suggested, we should enable as well.

### DIFF
--- a/Classes/Models/LabelDetails.swift
+++ b/Classes/Models/LabelDetails.swift
@@ -10,11 +10,13 @@ import Foundation
 
 final class LabelDetails {
 
+    let client: GithubClient
     let owner: String
     let repo: String
     let label: String
 
-    init(owner: String, repo: String, label: String) {
+    init(client: GithubClient, owner: String, repo: String, label: String) {
+        self.client = client
         self.owner = owner
         self.repo = repo
         self.label = label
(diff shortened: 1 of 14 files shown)


Classes/Models/LabelDetails.swift
```diff
@@ -10,11 +10,13 @@ import Foundation
 
 final class LabelDetails {
 
+    let client: GithubClient
```
rnystrom: Is there another way to pass the client around beside attaching it to a model? This puts a **lot** of baggage in these models (the client object is basically a full app session!)

BrianLitwin: Re-tried this in a new PR - didn't do enough homework on how this is handled elsewhere in the app. Thanks for the feedback!

PR #2314, opened 2018-10-19
Search for labels in app (take two)

BrianLitwin:
My second go at #2309 -> see that PR for vid. Functionality is the same. The difference is instead of passing the `GithubClient` around as a parameter in the LabelDetails model, used a protocol to route the tap to a view controller with access to `GithubClient`. Also, in order to do that, re-wired the IssueLabel SectionControllers to be the `MarkdownStyledTextViewDelegate` so that they could receive those taps and re-route them to the delegate w/ access to `GithubClient`.

### DIFF
--- a/Classes/Issues/Labeled/IssueLabeledSectionController.swift
+++ b/Classes/Issues/Labeled/IssueLabeledSectionController.swift
@@ -38,7 +38,7 @@ final class IssueLabeledSectionController: ListGenericSectionController<IssueLab
     
     func didTap(cell: MarkdownStyledTextView, attribute: DetectedMarkdownAttribute) {
         guard case .label(let label) = attribute else { return }
-        tapDelegate?.didTapLabel(owner: label.owner, repo: label.repo, label: label.label)
+        tapDelegate?.didTapIssueLabel(owner: label.owner, repo: label.repo, label: label.label)
     }
     
 }
--- a/Classes/Issues/Labeled/IssueLabeledSectionController.swift
+++ b/Classes/Issues/Labeled/IssueLabeledSectionController.swift
@@ -18,8 +18,8 @@ final class IssueLabeledSectionController: ListGenericSectionController<IssueLab
 
     init(issueModel: IssueDetailsModel, tapDelegate: IssueLabelTapSectionControllerDelegate) {
         self.issueModel = issueModel
-        super.init()
         self.tapDelegate = tapDelegate
+        super.init()
     }
 
     override func sizeForItem(at index: Int) -> CGSize {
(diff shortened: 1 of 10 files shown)


Classes/Issues/Labeled/IssueLabeledSectionController.swift
```diff
@@ -9,13 +9,17 @@
...
+final class IssueLabeledSectionController: ListGenericSectionController<IssueLabeledModel>, MarkdownStyledTextViewDelegate {
 
     private let issueModel: IssueDetailsModel
+    private weak var tapDelegate: IssueLabelTapSectionControllerDelegate?
 
-    init(issueModel: IssueDetailsModel) {
+    init(issueModel: IssueDetailsModel, tapDelegate: IssueLabelTapSectionControllerDelegate) {
         self.issueModel = issueModel
         super.init()
+        self.tapDelegate = tapDelegate
```
rnystrom: _Technically_ should be set before init, but getting away w/ it since its `weak`

Classes/Issues/Labeled/IssueLabeledSectionController.swift
```diff
@@ -28,8 +32,13 @@ final class IssueLabeledSectionController: ListGenericSectionController<IssueLab
...
-        cell.delegate = viewController
+        cell.delegate = self
         return cell
     }
-
+    
+    func didTap(cell: MarkdownStyledTextView, attribute: DetectedMarkdownAttribute) {
+        guard case .label(let label) = attribute else { return }
+        tapDelegate?.didTapIssueLabel(owner: label.owner, repo: label.repo, label: label.label)
+    }
```
rnystrom: Does this break showing the profile when tapping on the username in the labeled cell?

Classes/Repository/RepositoryIssuesViewController.swift
```diff
@@ -51,10 +58,13 @@ SearchBarSectionControllerDelegate {
...
-        // set the frame in -viewDidLoad is required when working with TabMan
-        feed.collectionView.frame = view.bounds
-        feed.collectionView.contentInsetAdjustmentBehavior = .never
+        
+        let presentingInTabMan = label == nil
+        if presentingInTabMan {
+            // set the frame in -viewDidLoad is required when working with TabMan
+            feed.collectionView.frame = view.bounds
+            feed.collectionView.contentInsetAdjustmentBehavior = .never
+        }
```
rnystrom: Why do we need this?

BrianLitwin: The top of the view controller was hiding under the nav bar and that fixed it

Classes/Issues/Labeled/IssueLabeledSectionController.swift
BrianLitwin: Yes it does- updated to: 
```swift 
 if case .label(let label) = attribute {
      tapDelegate?.didTapIssueLabel(owner: label.owner, repo: label.repo, label: label.label)
 } else {
      viewController?.handle(attribute: attribute)
 }
```

PR #2346, opened 2018-10-23
"Try Beta" on Settings

### LINKED ISSUE #430: Add "Try Beta" to Settings
We could upsell how to add yourself as a beta tester in settings. Get more early feedback on stuff. Not sure how to collect emails tho. I've seen the fastlane tools to do this, but it requires a Heroku app. Maybe it just opens Mail.app to email me?

rizwankce:
Adds "Try Beta" cell on Settings screen, On tap opens the testflight invite link in safari.

Closes #430

### DIFF
--- a/Classes/Settings/Settings.storyboard
+++ b/Classes/Settings/Settings.storyboard
@@ -20,7 +20,7 @@
                         <color key="separatorColor" red="0.73725490196078436" green="0.73333333333333328" blue="0.75686274509803919" alpha="1" colorSpace="calibratedRGB"/>
                         <inset key="separatorInset" minX="16" minY="0.0" maxX="0.0" maxY="0.0"/>
                         <view key="tableFooterView" contentMode="scaleToFill" id="jkq-3p-p02">
-                            <rect key="frame" x="0.0" y="693.5" width="375" height="44"/>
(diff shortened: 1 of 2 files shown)


rizwankce:
I tried to hide or remove the "Try Beta" cell is the user is already in TF. But the solutions either didn't work or introduce some layout bugs. So far I have tried hiding the cell or removing from superview. Proper solution would be overriding via `cellForRowIndex` or set `minimumHeight` to 0. But it's not future safe as we are playing with index paths on a static table view.

I guess it's okay to show the cell even in TF builds. Clicking on it open safari web page and "start testing" from test light link just opens TF app.

Sherlouk:
@rizwankce We could just open up a super basic UIAlertController saying something like

"You're already using the TestFlight build" etc (and not open the Safari VC)

Can keep the button and not have layout issues, should be easy enough to do with a quick conditional! (See the FLEX setup as an example)

rnystrom:
Ya don't sweat showing the link in TF builds. Also we wouldn't really be able to test it until its in prod anyways.

rizwankce:
@Sherlouk Nice idea! We can probably show a Squawk message as well. I will send a new PR! 👍

j-f1:
Perhaps on TestFlight, the cell should show the current release (1540303812 right now) and open TestFlight when tapped?

Issue #2054, opened 2018-08-06
No way to edit issue title


Huddie:
Is this true? Or I’m missing it?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rnystrom:
Nope this has never existed. I think there’s an issue already for it. More a design issue than anything, I’m not sure how we’d do this yet. Maybe in the manage button menu?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Huddie:
That works. Another possibility could be a pencil like icon across from the title only visible to the user. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BasThomas:
#381

BasThomas:
You know what, I am actually going to close that other, old issue.

BrianLitwin:
If we could settle on a UI, could PR this. Seems important that if we let users create issue titles, they should be able to edit them too.

rnystrom:
@BrianLitwin how about:

- New item in manage button
  - Manage button appears if you can make updates, which would include editing title
  - Use [pencil icon](https://octicons.github.com/icon/pencil/)
  - "Edit Title"
- Show some sort of editing control
  - If we want to be really fancy... a new ContextMenu control with a text box. Focus keyboard on appearance. Done & cancel button.
  - For a quick win: `UIAlertController` with a text field

BrianLitwin:
Anybody have strong thoughts on light or dark background for this? I'm thinking try to make work with a dark bg since every other ContextMenu in the Manage menu is dark.

rnystrom:
Actually think light would be better for text editing, but def open to see it as light (hopefully not hard to change either way)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BrianLitwin:
I'm all ears if anybody has UI suggestions! This is what is currently looks like: 

<img width="287" alt="screenshot 2018-10-22 15 53 31" src="https://user-images.githubusercontent.com/26695477/47315554-eb493900-d612-11e8-9ad9-0dd8d68b753f.png">

BrianLitwin:
# Video: 

![ezgif com-video-to-gif 4](https://user-images.githubusercontent.com/26695477/47316466-349a8800-d615-11e8-9440-21196b66c554.gif)


<br> 
<br> 

[...]

PR #2348, opened 2018-10-23
[WIP] Adds edit issue title

BrianLitwin:
Connected to #2054
See that thread for video and pics.. I did adjust the height to 120 as was recommended there. 

Manually tested that editing the title updates three areas: 
- the issueTitle section controller  
- the timeline 
- the bookmark (when you go back to bookmarks, title is updated) 

Included unit tests as well. 
Used the viewerCanUpdate flag to determine editing privileges. 
I also manually tested super long titles.. the textview handles them well and that shouldn't be an issue. 

#### Update: 
- I realized this also needs to update the bookmark struct that the nav controller will use to set a new bookmark - eg currently, if you edit a title -> then bookmark the issue -> the bookmark will reference the original title, not the edited one. 

- @rnystrom Thanks for your review! I'm learning a lot. I'm going to update the Edit VC to follow a pattern more similar to People, Milestones, etc.. update everything optimistically and reset if there's a network failure. Was looking at the EditComment VC for cues, that was prob not the optimal apprach.

- Updating the Bookmark API seems like a non-trivial issue, might need a separate PR, for the time being i'll do my best.


BrianLitwin:
@Sherlouk Applied your suggestions.. thanks for the feedback!

Classes/Issues/IssuesViewController.swift
```diff
@@ -286,6 +288,13 @@ final class IssuesViewController: MessageViewController,
         activityController.popoverPresentationController?.barButtonItem = sender
         present(activityController, animated: trueUnlessReduceMotionEnabled)
     }
+    
+    var insetWidth: CGFloat {
+        // assumptions here, but the collectionview may not have been laid out or content size found
+        // assume the collectionview is pinned to the view's bounds
+        let contentInset = feed.collectionView.contentInset
+        return view.bounds.width - contentInset.left - contentInset.right
+    }
```
BrianLitwin: Moved this so that I could re-use it in the completion block after the fetch; see below 

```swift 
         if let current = strongSelf.result {
                    let issueResult = IssuesViewController.updateIssueResultModelTitle(
                        newTitle: newTitle,
                        oldTitle: strongSelf.currentIssueTitle ?? "",
                        username: strongSelf.client.userSession?.username ?? Constants.Strings.unknown,
                        issueResultModel: current,
                        width: strongSelf.insetWidth
                    )
                    strongSelf.client.cache.set(value: issueResult)
                }
```

BrianLitwin:
@Sherlouk - I'm doing something wrong = I've been having trouble adding files to local pods, and getting the new XCode beta seemed to help, so I pulled this PR down to check on it, and the V3EditIssueTitleRequest in GItHubAPI didn't compile (didn't actually show up in XCode at all) - so I amended the commit, re-added the file so that it compiled and pushed it back up, and now there's this huge diff in ` Pods/Pods.xcodeproj/project.pbxproj` and the check is failing. 

For starters, I don't actually know the correct way to add a file to a local pod. When I try to add the file, it says Group: "GitHubAPI", and for target I'm putting a checkmark next to GithubAPI-iOS, because XCode doesn't complain and everything compiles when I do that. But I'm not sure that's correct?

Classes/Issues/IssuesViewController.swift
```diff
@@ -642,5 +648,59 @@ final class IssuesViewController: MessageViewController,
...
+            viewController.dismiss(animated: true)
+        }
+    }
+    
+    var currentIssueTitle: String? {
+        return result?.title.string.allText
+    }
+    
+    var viewerCanUpdate: Bool {
+        return result?.viewerCanUpdate ?? false
```
rnystrom: `IssueManagingContextController` already has access to [IssueResult](https://github.com/GitHawkApp/GitHawk/blob/master/Classes/Issues/IssueManagingContextController.swift#L65). These don't have to be delegate methods right? Can't these values just be fetched from the model when the edit VC is initialized?

Classes/Issues/EditTitle/EditIssueTitleViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+//
+
+import UIKit
+import GitHubAPI
+import SnapKit
+
+protocol EditIssueTitleViewControllerDelegate: class {
+    func sendEditTitleRequest(newTitle: String, viewController: EditIssueTitleViewController)
+    var currentIssueTitle: String? { get }
+    var viewerCanUpdate: Bool { get }
```
rnystrom: See my notes about `IssueManagingContextController` already having access to the model. IMO these last 2 properties can be removed.

Classes/Issues/EditTitle/EditIssueTitleViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+//  Created by B_Litwin on 10/22/18.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import UIKit
+import GitHubAPI
+import SnapKit
+
+protocol EditIssueTitleViewControllerDelegate: class {
+    func sendEditTitleRequest(newTitle: String, viewController: EditIssueTitleViewController)
```
rnystrom: nit: delegate pattern has `self` param **first**, then other data params. Something like this would be more consistent:

```swift
func sendEditTitleRequest(from viewController: EditIssueTitleViewController, with title: String)
```

Classes/Issues/EditTitle/EditIssueTitleViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+    var currentIssueTitle: String? { get }
+    var viewerCanUpdate: Bool { get }
+}
+
+class EditIssueTitleViewController: UIViewController {
```
rnystrom: `final`

Classes/Issues/EditTitle/EditIssueTitleViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        preferredContentSize = CGSize(
+            width: Styles.Sizes.contextMenuSize.width,
+            height: 120
+        )
+        title = NSLocalizedString("Edit", comment: "")
```
rnystrom: String `"Edit"` [already exists](https://github.com/GitHawkApp/GitHawk/blob/c23945ac7dfe015f3f7218668b5caef2355ac7db/Classes/Issues/Comments/IssueCommentSectionController.swift#L119), mind tossing that in `Constants.swift`?

Classes/Issues/EditTitle/EditIssueTitleViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+    }
+    
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        textView.becomeFirstResponder()
+    }
+    
+    func setRightBarItemIdle() {
+        navigationItem.rightBarButtonItem = UIBarButtonItem(
+            title: NSLocalizedString("Save", comment: ""),
```
rnystrom: [Constants](https://github.com/GitHawkApp/GitHawk/blob/c23945ac7dfe015f3f7218668b5caef2355ac7db/Classes/Issues/EditComment/EditCommentViewController.swift#L106)?

Classes/Issues/IssueManagingContextController.swift
```diff
@@ -91,6 +93,16 @@ final class IssueManagingContextController: NSObject, ContextMenuDelegate {
             if result.pullRequest {
                 actions.append(.reviewers)
             }
+            
+            let viewerCanUpdate =
+                editIssueTitleViewControllerDelegate?
+                .viewerCanUpdate
+                ?? false
+            
+            if viewerCanUpdate {
```
rnystrom: Since the result of this `Bool` isn't used again, can it just all be in the `if` statement? No need for a variable

Classes/Issues/IssueManagingContextController.swift
```diff
@@ -264,6 +280,18 @@ final class IssueManagingContextController: NSObject, ContextMenuDelegate {
             delegate: self
         )
     }
+    
+    func presentEditTitleController() {
+        guard let viewController = viewController else { return }
+        guard let delegate = editIssueTitleViewControllerDelegate else { return }
```
rnystrom:
```swift
guard let viewController = viewController,
    let delegate = editIssueTitleViewControllerDelegate
    else { return }
```

Classes/Issues/IssuesViewController+UpdateIssueTitle.swift
```diff
@@ -0,0 +1,71 @@
...
+import UIKit
+
+extension IssuesViewController {
+    
+    static func updateBookmarkTitle(
+        newTitle: String,
+        bookmark: Bookmark,
+        bookmarkStore: BookmarkStore
+        )
+    {
```
rnystrom:
```swift
) {
```

(is SwiftLint not catching this?)

Classes/Issues/IssuesViewController+UpdateIssueTitle.swift
```diff
@@ -0,0 +1,71 @@
...
+            owner: bookmark.owner,
+            number: bookmark.number,
+            title: newTitle,
+            defaultBranch: bookmark.defaultBranch
+        )
+        
+        if let index = bookmarkStore.values.index(of: bookmark) {
+            bookmarkStore.values[index] = newBookmark
+            bookmarkStore.save()
+        }
```
rnystrom: Oof, we need a better way to do this... shouldn't have to create whole model then lookup based on index... `BookmarkStore` needs better APIs for finding, updating, and replacing stuff...

I guess we can tackle that later, but this should be the responsibility of the store.

Also `var values: [Bookmark]` is a publicly mutable array... 😬 this could use some cleanup....

Classes/Issues/IssuesViewController+UpdateIssueTitle.swift
```diff
@@ -0,0 +1,71 @@
...
+    }
+    
+    static func updateIssueResultModelTitle(
+        newTitle: String,
+        oldTitle: String,
+        username: String,
+        issueResultModel: IssueResult,
+        width: CGFloat
+        ) -> IssueResult
+    {
```
rnystrom:
```swift
) -> IssueResult {
```

Classes/Issues/IssuesViewController+UpdateIssueTitle.swift
```diff
@@ -0,0 +1,71 @@
...
+        let title = titleStringSizing(
+            title: newTitle,
+            contentSizeCategory: UIContentSizeCategory.preferred,
+            width: width
+        )
+        
+        let titleChangeString = IssueRenamedString(
+            previous: oldTitle,
+            current: newTitle,
+            contentSizeCategory: UIContentSizeCategory.preferred,
```
rnystrom: Can you capture `preferred` in a `let` since its used multiple times?

Classes/Issues/IssuesViewController+UpdateIssueTitle.swift
```diff
@@ -0,0 +1,71 @@
...
+            titleChangeString: titleChangeString
+        )
+        
+        let issueResult = issueResultModel.updated(
+            title: title,
+            timelinePages: issueResultModel.timelinePages(appending: [issueRenamedModel])
+        )
+        
+        return issueResult
+    }
```
rnystrom: This is great! And tests! 🎉

Classes/Issues/IssuesViewController.swift
```diff
@@ -286,6 +288,13 @@ final class IssuesViewController: MessageViewController,
         activityController.popoverPresentationController?.barButtonItem = sender
         present(activityController, animated: trueUnlessReduceMotionEnabled)
     }
+    
+    var insetWidth: CGFloat {
```
rnystrom: `private`?

Classes/Issues/IssuesViewController.swift
```diff
@@ -312,10 +321,7 @@ final class IssuesViewController: MessageViewController,
             }
         }
 
-        // assumptions here, but the collectionview may not have been laid out or content size found
-        // assume the collectionview is pinned to the view's bounds
-        let contentInset = feed.collectionView.contentInset
-        let width = view.bounds.width - contentInset.left - contentInset.right
+        let width = insetWidth
```
rnystrom: Just move `insetWidth` to the param? Doesn't need a `let` now right?

Classes/Issues/IssuesViewController.swift
```diff
@@ -642,5 +648,59 @@ final class IssuesViewController: MessageViewController,
...
+                    )
+                    strongSelf.client.cache.set(value: issueResult)
+                }
+                
+            case .failure(let error):
+                Squawk.show(error: error)
+            }
+            
+            viewController.setRightBarItemIdle()
+            viewController.dismiss(animated: true)
```
rnystrom: Note: this is strongly retaining the `viewController` param

Classes/Issues/IssuesViewController.swift
```diff
@@ -642,5 +648,59 @@ final class IssuesViewController: MessageViewController,
         guard let issueType = self.issueType else { return }
         presentLabels(client: client, owner: owner, repo: repo, label: label, type: issueType)
     }
+    
+    // MARK: EditIssueTitleViewControllerDelegate
 
+    func sendEditTitleRequest(newTitle: String, viewController: EditIssueTitleViewController) {
```
rnystrom: Why not just put all the sending stuff _inside_ the edit VC? Can include the cache updating stuff since the issue VC listens for model changes and will reload. Would totally eliminate the need for this delegate API and keep things really separated.

Classes/Issues/IssuesViewController.swift
```diff
@@ -642,5 +648,59 @@ final class IssuesViewController: MessageViewController,
...
+                        width: strongSelf.insetWidth
+                    )
+                    strongSelf.client.cache.set(value: issueResult)
+                }
+                
+            case .failure(let error):
+                Squawk.show(error: error)
+            }
+            
+            viewController.setRightBarItemIdle()
```
rnystrom: So the edit VC sets its right bar spinning, but the _issue VC_ is required to set it back? Adds a lot of implicit logic. Another +1 for just putting all this logic inside the edit VC...

rnystrom:
@BrianLitwin just lmk when this is ready for another look!

ijm8710:
Hey @BrianLitwin you still planning to jump back on this? No rush at all, just checking in since it seemed you were almost done with it. Obv I’m just excited for this awesome addition and let me know if I’m overstepping by f/u and I’ll make sure I refrain from doing so again in future...

rnystrom:
@BrianLitwin lmk if you have plans to revisit this. Def a high-impact feature! Leaving it open for now.

BrianLitwin:
@rnystrom Thx for leaving it open- I def plan to re-submit this in the near future 

BrianLitwin:
@rnystrom with the new Bookmark storage model, there's no reason to worry about updating the Bookmark Store anymore if you edited an Issue's title? Because we are not saving the bookmark title anymore, but saving the graphQL id and then fetching the title + other info from graphQL?

Issue #2355, opened 2018-10-26
Consecutive references doesn’t link


Huddie:
See #2354.

I tested this. If you reference two issues in a row, the second reference doesn’t link. #2354 #2353

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Issue #2357, opened 2018-10-26
Search in split view doesn’t fill screen


Huddie:
See photo. The search view is short on the right side
![GitHawk Upload by Huddie](https://i.imgur.com/8qxAb6F.jpg)

<details>
<summary>Bug Report Dump (Auto-generated)</summary>
<pre>
Version 1.23.0 (1540303812)
Device: iPhone XR (iOS 12.0)
TestFlight: true
</pre>
</details>

rizwankce:
Same bug in book mark as well.  It's iOS 12 Issue probably. Check https://github.com/GitHawkApp/GitHawk/issues/2251

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Issue #2356, opened 2018-10-26
Reaction emoji doesn’t update in split view


Huddie:
If your in split view and set a new default reaction in settings the emoji icon won’t update in the main view (settings). 

This is because the cells only update on viewDidAppear.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rizwankce:
Ah! Didn't notice that when I added the reaction vc to split view. I'll take care of this 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #2358, opened 2018-10-27
Fixes regex for consecutive shortlinks

BrianLitwin:
Connected to #2355  

Made a small change to the regex, and updated the unit tests to include the new test case and to accommodate testing more than one regex match in a single test string. 
I tested that this worked manually as well: 

screenshot 2018-10-27 10 51 33


ijm8710:
>Connected to #2355   ...

@BrianLitwin #2355 #2356 #2357 #2358 #2359

ijm8710:
> ...

@BrianLitwin looks like it doesn’t work^

PR #2359, opened 2018-10-27
Fixes Reaction emoji updates on split view

### LINKED ISSUE #2356: Reaction emoji doesn’t update in split view

rizwankce:
Closes #2356

Sry! oversight from mysids. This was introduced from my previous PR. 😦

### DIFF
--- a/Classes/Settings/DefaultReactionDetailController.swift
+++ b/Classes/Settings/DefaultReactionDetailController.swift
@@ -8,6 +8,10 @@
 
 import UIKit
 
+protocol DefaultReactionListener {
+    func didUpdateDefaultReaction()
+}
+
 class DefaultReactionDetailController: UITableViewController {
 
     @IBOutlet var thumbsUpCell: UITableViewCell!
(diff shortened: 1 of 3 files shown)


Classes/Settings/DefaultReactionDetailController.swift
```diff
@@ -18,6 +22,8 @@ class DefaultReactionDetailController: UITableViewController {
     @IBOutlet var heartCell: UITableViewCell!
     @IBOutlet var enabledSwitch: UISwitch!
 
+    var listener: DefaultReactionListener?
```
rnystrom: `weak` to avoid retain cycle! Also the protocol + var should just use the term "Delegate" b/c this is a 1:1 (not 1:n)

rizwankce: Done! Thx for pointing out.

Issue #2342, opened 2018-10-22
Feature request: clone repo in Working Copy


jdmarshall90:
Working Copy appears to have x-callback-url support for this

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rnystrom:
Love it! Could fit this in the “...” bar item

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

palmin:
The most suitable url to use is probably

    working-copy://show?remote=https%3A%2F%2Fgithub.com%2Fgit%2Fgit.git

that shows the given repository inside Working Copy if already cloned and will otherwise show a cloning dialog.

Sherlouk:
Will need to make sure we use can open to make sure the user actually has WC installed too (just a reminder)

palmin:
@rnystrom I would love to have this working and you can assign it to me if you like.

PR #2366, opened 2018-10-28
Allow showing/cloning repository in Working Copy when installed

palmin:
.../more button when looking at repository has extra action when Working Copy is installed
to show or clone repository in this app.

This should address #2342

### DIFF
--- a/Classes/Repository/RepositoryViewController.swift
+++ b/Classes/Repository/RepositoryViewController.swift
@@ -159,6 +159,20 @@ ContextMenuDelegate {
             .newIssue(issueController: newIssueViewController)
     }
 
+    func workingCopyAction() -> UIAlertAction? {
+        guard let remote = self.repoUrl.absoluteString.addingPercentEncoding(withAllowedCharacters: CharacterSet.alphanumerics) else { return nil}
+
+        guard let url = URL(string: "working-copy://show?remote=\(remote)") else { return nil }
+        guard UIApplication.shared.canOpenURL(url) else { return nil }
+
+        let title = NSLocalizedString("Working Copy", comment: "")
+        let action = UIAlertAction(title: title, style: .default,
+                                   handler: { _ in
+            UIApplication.shared.open(url)
+        })
+        return action
+    }
+
     @objc func onMore(sender: UIButton) {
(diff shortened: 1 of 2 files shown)


rnystrom:
This is so great!! Eventually I’d love to build a Routable out of this, but this is a GREAT way to get us started.

@palmin so this won’t display the action of Working Copy isn’t installed? 👌

Long term might be cool to add a page in settings for supported app links and add a link to download Working Copy.

ALSO I’ll ping you on twitter but I’m gonna change the routes lib a little to update the paths. Thought of a way to make that simpler but it’ll break any URLs if you started adding that to WC.

BasThomas:
Yeah would be cool to be able to download the app (a la Shortcuts) if it isn’t installed!

palmin:
@rnystrom There will be no action shown unless Working Copy is installed. 
This is handled by:

https://github.com/palmin/GitHawk/blob/8ce285d86bdc18f242bda0aeb824332aa8aa3a33/Classes/Repository/RepositoryViewController.swift#L166

Classes/Repository/RepositoryViewController.swift
```diff
@@ -159,6 +159,20 @@ ContextMenuDelegate {
             .newIssue(issueController: newIssueViewController)
     }
 
+    func workingCopyAction() -> UIAlertAction? {
+        guard let remote = self.repoUrl.absoluteString.addingPercentEncoding(withAllowedCharacters: CharacterSet.alphanumerics) else { return nil}
+
+        guard let url = URL(string: "working-copy://show?remote=\(remote)") else { return nil }
+        guard UIApplication.shared.canOpenURL(url) else { return nil }
```
freak4pc: @palmin Would this work without adding `working-copy` to `LSApplicationQueriesSchemes` ? 

https://developer.apple.com/documentation/uikit/uiapplication/1622952-canopenurl

> If your app is linked on or after iOS 9.0, you must declare the URL schemes you pass to this method by adding the LSApplicationQueriesSchemes key to your app's Info.plist file. This method always returns false for undeclared schemes, whether or not an appropriate app is installed. To learn more about the key, see LSApplicationQueriesSchemes.

Issue #697, opened 2017-10-23
Network stack mock


heshamsalman:
Ref #568 

way i see it, we won't be able to creep up that VC test rate until we have a mock for the client and all of the little guys it needs along the way. 
good mock should be able to to take a local path to a response and return that to the tests.

BasThomas:
Think we have this now?

rnystrom:
I never landed it 😞 I’ve been thinking about refactoring all the client stuff to be a protocol so a unit test or UI test can swap in its own stub

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BrianLitwin:
@rnystrom @BasThomas or any Upperclassmen: 

I've tried subclassing the [Client](https://github.com/GitHawkApp/GitHawk/blob/22ab3fbf19d47dc11c010247630e56ae4a4fa516/Local%20Pods/GitHubAPI/GitHubAPI/Client.swift#L26), I've tried moving the request methods into a protocol: 

<details><summary>ClientRequestPerformer</summary> 

```swift 
public protocol ClientRequestPerformer {
    func send<T: HTTPRequest>(_ request: T, completion: @escaping (Result<T.ResponseType>) -> Void)
    
    func query<T: GraphQLQuery, Q>(
        _ query: T,
        result: @escaping (T.Data) -> Q?,
        completion: @escaping (Result<Q>) -> Void
        ) -> Cancellable
    
    func mutate<T: GraphQLMutation, Q>(
        _ mutation: T,
        result: @escaping (T.Data) -> Q?,
        completion: @escaping (Result<Q>) -> Void
    ) -> Cancellable
}
```
</details> 
<br />

I'm not able to test anything using either strategy. I feel like I'm missing something basic (to do with GitHubAPI being in a separate pod?), b/c I can't even set and trip a breakpoint or print anything to the console in the Client's query methods, either in a test or running the app in the simulator. 

Would love to figure out how to do this so that we can write more tests.

BasThomas:
Does it? Cocoapods works for us at work (it didn't use to until the last beta)

BrianLitwin:
Good news! Getting the latest XCode beta worked! Good tip @BasThomas ! 

so the following code works for me to subclass the Client: 
```swift
extension GithubClient {
    init(mockClient: Client) {
        self.userSession = nil
        self.cache = FlatCache()
        self.bookmarksStore = nil
        self.client = mockClient
[...]

PR #2368, opened 2018-10-29
Client Mock & Search View Controller Tests

heshamsalman:
Special thanks to @BrianLitwin for getting this work going.
Ref #697 

I should probably note that this is the _simplest_ possible implementation of a mock -- it's not complete for anything but the simplest of cases. 

The next step after this would be to allow for stub data.

@rnystrom

### DIFF
--- a/Classes/History/Client+History.swift
+++ b/Classes/History/Client+History.swift
@@ -9,7 +9,7 @@
 import Foundation
 import GitHubAPI
 
-extension Client {
+extension ClientType {
 
     func fetchHistory(
         owner: String,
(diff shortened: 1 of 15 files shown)


heshamsalman:
Looks like my commit history got mangled here; gonna end up rebasing to clean this up

heshamsalman:
At present no, but when it allows for stubbed data, yes. 

We can model out the exact data payloads that a screen or component requests and then test that the UI generated from that payload matches our expectations.

heshamsalman:
Getting to the point where we can have stubbed data is gonna be tricky though; PATs are tough to reason about/mock if you don't want to start entering type-erasure / type-magic territory.

Issue #2393, opened 2018-11-04
Bringing templates to GitHawk


Huddie:
Working on this, if anyone has ideas or suggestions please lmk!

<img src=https://user-images.githubusercontent.com/10662471/47959045-1dae4b00-dfb0-11e8-9072-6a8a15069696.gif width=200>

rnystrom:
Yo this is so cool!! Would this work for the “new issue” feature in repos too?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Huddie:
@rnystrom I’m almost positive! I checked github and all issues are under .github/ISSUE_TEMPLATE 

Right now I fetch for specific files but my end plan is fetch all files under that directory, parse the text for the name / description 

Should work for all repos! 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Huddie:
@rnystrom Confirmed that it will work across the board. Done for the night 🌙

<img src=https://user-images.githubusercontent.com/10662471/47960194-3a0cb080-dfcd-11e8-9a09-8354aef96c8d.gif width=200>

Huddie:
@Sherlouk the old one being in root/docs/.github rather than strictly .github? I’m not so familiar with github templates, searching online It seemed that you could have as many templates as you want but must be in .github/ISSUE_TEMPLATE directory. Is this not true, can you have templates in other places? Quick note: I’m speaking specifically about issue templates not PR templates 

I think for now it’s fine to support this way as if no templates are found it simply opens up the new issue view like it used to. We can expand to support other templates in the future, if others exist.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Sherlouk:
@Huddie Yea so I think the difference is with the new system you can have multiple named templates but previously you could have a single template that was called `ISSUE_TEMPLATE.md` (and could be in a multitude of places) and would always use that one.

Essentially I think the chain of logic is:

1. Check for templates in `.github/ISSUE_TEMPLATE/` directory
2. Check for single template called `ISSUE_TEMPLATE.md` in either root or `.github` (name not case sensitive)
3. Just leave it empty

I think for the sake of this, just doing that first one is definitely great progress!

Huddie:
Thanks @Sherlouk I will PR number 1 and then maybe expand after that. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

[...]

PR #2400, opened 2018-11-04
WIP - Templates

Huddie:
Looking for preliminary feedback. I'm thinking of moving regex matching to an extension. I originally had it that the IssueTemplateHelper returned the templates to the caller and let it deal with it how it wished. I just then felt that since every view that wants to open a new issue needs to go through the same process that maybe its better to pass the view controller in and keep the code all in one place. Since it starts with an extension on the client the client isn't being passed around. 

Anyways let me know what to fix.

TODO:
- [ ] Add Spinner 
- [ ] Remove template details before passing it through (Text between the dashes ( --- ) at the beginning of a template file)

### DIFF
--- a/Classes/New Issue/IssueTemplates.swift
+++ b/Classes/New Issue/IssueTemplates.swift
@@ -1,0 +1,220 @@
+//
+//  IssueTemplates.swift
+//  Freetime
+//
+//  Created by Ehud Adler on 11/3/18.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+import GitHubAPI
+import GitHubSession
+import Squawk
+
+struct IssueTemplate {
+    let title:String
+    let template:String
+}
+
+enum IssueTemplateHelper {
+
+    static private func matchesForRegexInText(
+        regex: String!,
+        text: String!) -> [String] {
+
+        do {
+            let regex = try NSRegularExpression(pattern: regex, options: [])
+            let nsString = text as NSString
+
+            let results = regex.matches(
+                in: text,
+                options: [],
+                range: NSMakeRange(0, nsString.length))
+            return results.map { nsString.substring(with: $0.range)}
+
+        } catch let error as NSError {
+            print("invalid regex: \(error.localizedDescription)")
+            return []
+        }
+    }
+
+    static func getNameAndDescriptionFromTemplateFile(file: String) -> (name: String?, about: String?) {
+
+        let names = IssueTemplateHelper.matchesForRegexInText(regex: "(?<=name:).*", text: file)
+        let abouts = IssueTemplateHelper.matchesForRegexInText(regex: "(?<=about:).*", text: file)
+        let name = names.count > 0
+            ? names[0].trimmingCharacters(in: .whitespaces)
+            : nil
+        let about = abouts.count > 0
+            ? abouts[0].trimmingCharacters(in: .whitespaces)
+            : nil
+        return (name, about)
+    }
+
+    static func showIssueAlert(
+        with templates: [IssueTemplate],
+        owner: String,
+        repo: String,
+        session: GitHubUserSession?,
+        mainViewController: UIViewController,
+        delegate: NewIssueTableViewControllerDelegate) {
+
+        let alertView = UIAlertController(
+            title: NSLocalizedString("New Issue", comment: ""),
+            message: NSLocalizedString("Choose Template", comment: ""),
+            preferredStyle: .alert
+        )
+
+        for template in templates {
+            alertView.addAction(
+                UIAlertAction(
+                    title: template.title,
+                    style: .default,
+                    handler: { _ in
+
+                        guard let viewController = NewIssueTableViewController.createWithTemplate(
+                            client: GithubClient(userSession: session),
+                            owner: owner,
+                            repo: repo,
+                            template: template.template,
+                            signature: template.title == "Bug Report" ? .bugReport : .sentWithGitHawk
+                            ) else {
+                                Squawk.showGenericError()
+                                return
+                        }
+                        viewController.delegate = delegate
+                        let navController = UINavigationController(rootViewController: viewController)
+                        navController.modalPresentationStyle = .formSheet
+                        mainViewController.present(navController, animated: trueUnlessReduceMotionEnabled)
+                }))
+        }
+
+        alertView.addAction(
+            UIAlertAction(
+                title: "Dismiss",
+                style: UIAlertActionStyle.cancel,
+                handler: { _ in
+                    alertView.dismiss(animated: true, completion: nil)
+            })
+        )
+        mainViewController.present(alertView, animated: true, completion: nil)
+    }
+}
+
+extension GithubClient {
+
+    private func fetchTemplateFile(
+        owner: String,
+        repo: String,
+        filename: String,
+        completion: @escaping (Result<String>) -> Void
+        ) {
+
+        self.fetchFile(
+            owner: owner,
+            repo: repo,
+            branch: "master",
+            path: ".github/ISSUE_TEMPLATE/\(filename)") { (result) in
+                switch result {
+                case .success(let file):
+                    completion(.success(file))
+                case .error(let error):
+                    completion(.error(error))
+                case .nonUTF8:
+                    completion(.error(nil))
+                }
+        }
+    }
+
+    private func createTemplateWith(
+        owner: String,
+        repo: String,
+        filename: String,
+        completion: @escaping (Result<IssueTemplate>) -> Void
+        ) {
+        fetchTemplateFile(owner: owner, repo: repo, filename: filename) { (result) in
+            switch result {
+            case .success(let file):
+                let nameAndDescription = IssueTemplateHelper.getNameAndDescriptionFromTemplateFile(file: file)
+                if let name = nameAndDescription.name {
+                    completion(.success(IssueTemplate(title: name, template: file)))
+                } else {
+                    completion(.error(nil))
+                }
+            case .error(let error):
+                completion(.error(error))
+            }
+        }
+    }
+
+    func createNewIssue(
+        owner: String,
+        repo: String,
+        session: GitHubUserSession?,
+        mainViewController: UIViewController,
+        delegate: NewIssueTableViewControllerDelegate) {
+
+        var templates: [IssueTemplate] = []
+        let templateGroup = DispatchGroup()
+
+        self.fetchFiles(
+            owner: owner,
+            repo: repo,
+            branch: "master",
+            path: ".github/ISSUE_TEMPLATE") { (result) in
+                switch result {
+                case .success(let files):
+                    for file in files {
+                        templateGroup.enter()
+                        self.createTemplateWith(owner: owner, repo: repo, filename: file.name, completion: {
+                            switch $0 {
+                            case .success(let template):
+                                templates.append(template)
+                            case .error(let error):
+                                if let err = error {
+                                    Squawk.show(error: err)
+                                }
+                            }
+                            templateGroup.leave()
+                        })
+                    }
+                case .error(let error):
+                    Squawk.show(error: error)
+                }
+
+                // Wait for async calls in for-loop to finish up
+                templateGroup.notify(queue: .main) {
+
+                    // Sort lexicographically
+                    let sortedTemplates = templates.sorted(by: {$0.title < $1.title })
+
+                    if sortedTemplates.count > 0 {
+                        // Templates exists...
+                        IssueTemplateHelper.showIssueAlert(
+                            with: sortedTemplates,
+                            owner: owner,
+                            repo: repo,
+                            session: session,
+                            mainViewController: mainViewController,
+                            delegate: delegate
+                        )
+                    } else {
+                        // No templates exists, show blank new issue view controller
+                        guard let viewController = NewIssueTableViewController.create(
+                            client: GithubClient(userSession: session),
+                            owner: owner,
+                            repo: repo,
+                            signature: .sentWithGitHawk
+                            ) else {
+                                Squawk.showGenericError()
+                                return
+                        }
+                        viewController.delegate = delegate
+                        let navController = UINavigationController(rootViewController: viewController)
+                        navController.modalPresentationStyle = .formSheet
+                        mainViewController.present(navController, animated: trueUnlessReduceMotionEnabled)
+                    }
+                }
+        }
+    }
+}
--- a/Classes/New Issue/NewIssueTableViewController.swift
+++ b/Classes/New Issue/NewIssueTableViewController.swift
@@ -51,6 +51,7 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
     @IBOutlet var titleField: UITextField!
     @IBOutlet var bodyField: UITextView!
 
+    private var template: String = ""
     private var client: GithubClient!
     private var owner: String!
     private var repo: String!
@@ -85,6 +86,26 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         return viewController
     }
 
+    static func createWithTemplate(
+        client: GithubClient,
+        owner: String,
+        repo: String,
+        template: String,
+        signature: IssueSignatureType? = nil) -> NewIssueTableViewController? {
+
+        let storyboard = UIStoryboard(name: "NewIssue", bundle: nil)
+
+        let viewController = storyboard.instantiateInitialViewController() as? NewIssueTableViewController
+        viewController?.hidesBottomBarWhenPushed = true
+        viewController?.client = client
+        viewController?.owner = owner
+        viewController?.repo = repo
+        viewController?.template = template
+        viewController?.signature = signature
+
+        return viewController
+    }
+
     override func viewDidLoad() {
         super.viewDidLoad()
 
@@ -106,6 +127,7 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         // Setup markdown input view
         bodyField.githawkConfigure(inset: false)
         setupInputView()
+        bodyField.text = template
 
         // Update title to use localization
         title = Constants.Strings.newIssue
--- a/Classes/Repository/RepositoryViewController.swift
+++ b/Classes/Repository/RepositoryViewController.swift
@@ -142,21 +142,16 @@ ContextMenuDelegate {
     }
 
     func newIssueAction() -> UIAlertAction? {
-        guard let newIssueViewController = NewIssueTableViewController.create(
-            client: client,
-            owner: repo.owner,
-            repo: repo.name,
-            signature: .sentWithGitHawk)
-        else {
-            Squawk.showGenericError()
-            return nil
+        let action = UIAlertAction(title: "New Issue", style: UIAlertActionStyle.default) { _ in
+            self.client.createNewIssue(
+                owner: self.repo.owner,
+                repo: self.repo.name,
+                session: nil,
+                mainViewController: self,
+                delegate: self
+            )
         }
-
-        newIssueViewController.delegate = self
-        weak var weakSelf = self
-
-        return AlertAction(AlertActionBuilder { $0.rootViewController = weakSelf })
-            .newIssue(issueController: newIssueViewController)
+        return action
     }
 
     func workingCopyAction() -> UIAlertAction? {
--- a/Classes/Settings/SettingsViewController.swift
+++ b/Classes/Settings/SettingsViewController.swift
@@ -162,19 +162,17 @@ NewIssueTableViewControllerDelegate, DefaultReactionDelegate {
     }
 
     func onReportBug() {
-        guard let viewController = NewIssueTableViewController.create(
-                client: GithubClient(userSession: sessionManager.focusedUserSession),
-                owner: "GitHawkApp",
-                repo: "GitHawk",
-                signature: .bugReport
-            ) else {
-                Squawk.showGenericError()
-                return
-        }
-        viewController.delegate = self
-        let navController = UINavigationController(rootViewController: viewController)
-        navController.modalPresentationStyle = .formSheet
-        present(navController, animated: trueUnlessReduceMotionEnabled)
+        showTemplateOptions()
+    }
+
+    func showTemplateOptions() {
+        client.createNewIssue(
+            owner: "GitHawkApp",
+            repo: "GitHawk",
+            session: sessionManager.focusedUserSession,
+            mainViewController: self,
+            delegate: self
+        )
     }
 
     func onViewSource() {
--- a/Freetime.xcodeproj/project.pbxproj
+++ b/Freetime.xcodeproj/project.pbxproj
@@ -2161,6 +2161,7 @@
 			children = (
 				98B5A0811F6C73D6000617D6 /* NewIssue.storyboard */,
 				98647DF21F758CCF00A4DE7A /* NewIssueTableViewController.swift */,
+				C0DEB1DC218E7BFB00B6FEFD /* IssueTemplates.swift */,
 			);
 			path = "New Issue";
 			sourceTree = "<group>";
@@ -2948,6 +2949,7 @@
 				29F3A18820CBFF8700645CB7 /* CreateNotificationTitle.swift in Sources */,
 				292EB08D1F1FF58D0046865D /* IssueMilestoneEventModel.swift in Sources */,
 				292EB0911F1FF72D0046865D /* IssueMilestoneEventSectionController.swift in Sources */,
+				C0DEB1DD218E7BFB00B6FEFD /* IssueTemplates.swift in Sources */,
 				DCA5ED141FAEE8030072F074 /* Bookmark.swift in Sources */,
 				299F4A89204CEDDC004BA4F0 /* Client+AccessToken.swift in Sources */,
 				2931892F1F539C0E00EF0911 /* IssueMilestoneSectionController.swift in Sources */,


Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+import GitHubAPI
+import GitHubSession
+import Squawk
+
+struct IssueTemplate {
+    let title:String
```
BasThomas: Nit: spacing after `:`

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+    let template:String
+}
+
+enum IssueTemplateHelper {
+
+    static private func matchesForRegexInText(
+        regex: String!,
```
BasThomas: What about `matches(forRegex:inText:)`?
And why are string types force unwrapped?

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+        text: String!) -> [String] {
+
+        do {
+            let regex = try NSRegularExpression(pattern: regex, options: [])
```
BasThomas: FYI you can omit the `options` parameter as that has a default expression of `[]`.

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+            let nsString = text as NSString
+
+            let results = regex.matches(
+                in: text,
+                options: [],
```
BasThomas: Here too

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+                range: NSMakeRange(0, nsString.length))
+            return results.map { nsString.substring(with: $0.range)}
+
+        } catch let error as NSError {
```
BasThomas: The compiler likes this? Cool.

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+            print("invalid regex: \(error.localizedDescription)")
+            return []
+        }
+    }
+
+    static func getNameAndDescriptionFromTemplateFile(file: String) -> (name: String?, about: String?) {
```
BasThomas: `getNameAndDescription(fromTemplateFile:)`?

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+
+        let names = IssueTemplateHelper.matchesForRegexInText(regex: "(?<=name:).*", text: file)
+        let abouts = IssueTemplateHelper.matchesForRegexInText(regex: "(?<=about:).*", text: file)
+        let name = names.count > 0
```
BasThomas: What about `names.first?.trimmingCharacters(in: .whitespaces)`? Then you skip some checks.

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+            ? names[0].trimmingCharacters(in: .whitespaces)
+            : nil
+        let about = abouts.count > 0
+            ? abouts[0].trimmingCharacters(in: .whitespaces)
+            : nil
+        return (name, about)
+    }
+
+    static func showIssueAlert(
```
BasThomas: It would be cool if a function creates the alert and returns it, so it can be tested.

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+        owner: String,
+        repo: String,
+        session: GitHubUserSession?,
+        mainViewController: UIViewController,
+        delegate: NewIssueTableViewControllerDelegate) {
+
+        let alertView = UIAlertController(
+            title: NSLocalizedString("New Issue", comment: ""),
+            message: NSLocalizedString("Choose Template", comment: ""),
+            preferredStyle: .alert
```
BasThomas: Not sure if we want to have this be an alert instead of an action sheet?

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+                    handler: { _ in
+
+                        guard let viewController = NewIssueTableViewController.createWithTemplate(
+                            client: GithubClient(userSession: session),
+                            owner: owner,
+                            repo: repo,
+                            template: template.template,
+                            signature: template.title == "Bug Report" ? .bugReport : .sentWithGitHawk
+                            ) else {
+                                Squawk.showGenericError()
```
BasThomas: Can we show a custom error with Squawk? Maybe add an error message / code to debug?

BasThomas: Or at least an `assert` is this is not expected to happen.

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+                                return
+                        }
+                        viewController.delegate = delegate
+                        let navController = UINavigationController(rootViewController: viewController)
```
BasThomas: Not sure if we want to pull out this presenter logic. cc @rnystrom

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+                        navController.modalPresentationStyle = .formSheet
+                        mainViewController.present(navController, animated: trueUnlessReduceMotionEnabled)
+                }))
+        }
+
+        alertView.addAction(
+            UIAlertAction(
+                title: "Dismiss",
+                style: UIAlertActionStyle.cancel,
```
BasThomas: You can omit `UIAlertActionStyle`; it is inferred.

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+                        viewController.delegate = delegate
+                        let navController = UINavigationController(rootViewController: viewController)
+                        navController.modalPresentationStyle = .formSheet
+                        mainViewController.present(navController, animated: trueUnlessReduceMotionEnabled)
+                }))
+        }
+
+        alertView.addAction(
+            UIAlertAction(
+                title: "Dismiss",
```
BasThomas: This should be localized.

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+                style: UIAlertActionStyle.cancel,
+                handler: { _ in
+                    alertView.dismiss(animated: true, completion: nil)
```
BasThomas: You can omit the `completion` parameter; it defaults to `nil`.

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+            })
+        )
+        mainViewController.present(alertView, animated: true, completion: nil)
```
BasThomas: Same here.

BasThomas: And make them `trueUnlessReduceMotionEnabled` maybe? :)

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+        owner: String,
+        repo: String,
+        filename: String,
+        completion: @escaping (Result<String>) -> Void
+        ) {
+
+        self.fetchFile(
+            owner: owner,
+            repo: repo,
+            branch: "master",
```
BasThomas: Is this expected to be hardcoded?

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+                    completion(.success(file))
+                case .error(let error):
+                    completion(.error(error))
+                case .nonUTF8:
+                    completion(.error(nil))
+                }
+        }
+    }
+
+    private func createTemplateWith(
```
BasThomas: Neat!

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+                } else {
+                    completion(.error(nil))
+                }
+            case .error(let error):
+                completion(.error(error))
+            }
+        }
+    }
+
+    func createNewIssue(
```
BasThomas: Maybe split some of this up to make it more readable; it does a lot, making it hard to parse.

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,220 @@
...
+                    Squawk.show(error: error)
+                }
+
+                // Wait for async calls in for-loop to finish up
+                templateGroup.notify(queue: .main) {
+
+                    // Sort lexicographically
+                    let sortedTemplates = templates.sorted(by: {$0.title < $1.title })
+
+                    if sortedTemplates.count > 0 {
```
BasThomas: This logic can be pulled out, for instance.

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -51,6 +51,7 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
     @IBOutlet var titleField: UITextField!
     @IBOutlet var bodyField: UITextView!
 
+    private var template: String = ""
```
BasThomas: Nit: you can omit the `: String`; it is inferred.

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -85,6 +86,26 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         return viewController
     }
 
+    static func createWithTemplate(
+        client: GithubClient,
+        owner: String,
+        repo: String,
+        template: String,
+        signature: IssueSignatureType? = nil) -> NewIssueTableViewController? {
```
BasThomas: Nit: newline for the return type?

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -85,6 +86,26 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
...
+
+        let storyboard = UIStoryboard(name: "NewIssue", bundle: nil)
+
+        let viewController = storyboard.instantiateInitialViewController() as? NewIssueTableViewController
```
BasThomas: Should we exit scope if we can't do this? Can't imagine that that would be the case unless somebody removes the storyboard or changes its name 😛

Classes/Repository/RepositoryViewController.swift
```diff
@@ -142,21 +142,16 @@ ContextMenuDelegate {
...
     func newIssueAction() -> UIAlertAction? {
-        guard let newIssueViewController = NewIssueTableViewController.create(
-            client: client,
-            owner: repo.owner,
-            repo: repo.name,
-            signature: .sentWithGitHawk)
-        else {
-            Squawk.showGenericError()
-            return nil
+        let action = UIAlertAction(title: "New Issue", style: UIAlertActionStyle.default) { _ in
```
BasThomas: `"New Issue"` should be localized.

BasThomas: And you can omit the `UIAlertActionStyle` again

Classes/New Issue/IssueTemplates.swift
Huddie: Ya I'll switch to assert

Classes/Repository/RepositoryViewController.swift
Huddie: I changed this in both createWithTemplate(..) and create(..), I'm going to move all the localized strings to the Constants file as well

Classes/New Issue/IssueTemplates.swift
Huddie: Let me know. I can switch

BasThomas: @rnystrom?

Classes/New Issue/IssueTemplates.swift
BasThomas: I don't think so. I think you can even not have a `master` branch; can we get the default branch somehow? cc @rnystrom

Classes/New Issue/NewIssueTableViewController.swift
BasThomas: Would be awesome! Can also do this in a follow up sometime.

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,233 @@
...
+                    title: template.title,
+                    style: .default,
+                    handler: { _ in
+
+                        guard let viewController = NewIssueTableViewController.createWithTemplate(
+                            client: GithubClient(userSession: details.session),
+                            owner: details.owner,
+                            repo: details.repo,
+                            template: template.template,
+                            signature: template.title == "Bug Report" ? .bugReport : .sentWithGitHawk
```
Sherlouk: How does everyone feel about this check?

Is this essentially saying that if *anybody* creates a template in their repository called "Bug Report" then it will use the bug report template? Or is this limited just to the GitHawk repo?

I ask this because the Bug Report signature is pretty specific to GitHawk even going as far to include TestFlight information which isn't really applicable if it's a bug report on _pluck random repo out of the air_ equally the device info probably isn't that important.

Might be missing something here though!

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,233 @@
...
+        repo: String,
+        filename: String,
+        completion: @escaping (Result<String>) -> Void
+        ) {
+
+        self.fetchFile(
+            owner: owner,
+            repo: repo,
+            branch: "master",
+            path: ".github/ISSUE_TEMPLATE/\(filename)") { (result) in
```
Sherlouk: Have we double checked if GitHub is specific about the caps on this? I know previously when I investigated this I had way too many failures causes by the fact GitHub was case insensitive when checking for this!

(This could be different now though!)

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,233 @@
...
+            session: session,
+            viewController: mainViewController,
+            delegate: delegate
+        )
+
+        self.fetchFiles(
+            owner: owner,
+            repo: repo,
+            branch: "master",
+            path: ".github/ISSUE_TEMPLATE") { (result) in
```
Sherlouk: Same comment as previous, just would like to test case sensitivity on this

Classes/New Issue/IssueTemplates.swift
Huddie: Ya that a good point. I meant to revisit this. Maybe check if repo =  GitHawk or owner = GitHawkApp

Classes/New Issue/IssueTemplates.swift
Sherlouk: Also think @BasThomas picked up on the hardcoded branch here, maybe needs to be the default branch?

Super nit: should we move these constants to somewhere together? (The two path constants)

Sherlouk:
Think this is an ace addition to the app, making it super easy to make issues can only help productivity! Good work!

Classes/New Issue/IssueTemplates.swift
BasThomas: Yep, that seems fair. Maybe create a `GitHawkApp` repo constant somewhere and check that. Didn't think of that.

Classes/New Issue/IssueTemplates.swift
Huddie: Good point. I remember saying it must be under ISSUE_TEMPLATE but ill recheck.

BasThomas: It is case insensitive; we should check that.

Huddie:
> Think this is an ace addition to the app, making it super easy to make issues can only help productivity! Good work!

Thanks! I really appreciate that

BasThomas:
Yessss super cool!

Classes/New Issue/IssueTemplates.swift
Huddie: @BasThomas as in you can have .github/issue_template ? Or the current code is case insensitive?

Classes/Settings/SettingsViewController.swift
```diff
@@ -162,19 +162,17 @@ NewIssueTableViewControllerDelegate, DefaultReactionDelegate {
...
-        }
-        viewController.delegate = self
-        let navController = UINavigationController(rootViewController: viewController)
-        navController.modalPresentationStyle = .formSheet
-        present(navController, animated: trueUnlessReduceMotionEnabled)
+        showTemplateOptions()
+    }
+
+    func showTemplateOptions() {
+        client.createNewIssue(
```
Sherlouk: To make sure I'm clued up on what's happening here, so tapping on "Report a Bug" is now going to ask you which template to use "Bug Report" or "Feature Request", is that correct?

I'm easy but is this the functionality we want? Maybe we should rename the button or maybe default to the Bug Report template, or if people are fine with this we can keep the same!

Classes/New Issue/IssueTemplates.swift
Huddie: Ya I'll create that. 👍

Classes/New Issue/IssueTemplates.swift
Huddie: @BasThomas "If you create templates on other branches, they will not be available for collaborators to use."
[Link](https://help.github.com/articles/about-issue-and-pull-request-templates/)

Classes/New Issue/IssueTemplates.swift
Sherlouk: Could go fully overpowered and add it as an option to the markdown table at the top of template 😉but nah think the simple and easy solution is as above!

Classes/New Issue/IssueTemplates.swift
Huddie: Actually, if someone knows how to get default branch easily that would be good. I know we have access to it in the branch switcher, not sure if its a quick get or not

Classes/Settings/SettingsViewController.swift
Huddie: I don't have a preference,  switching to Report Bug/ Feature request sounds good or defaulting to bug. Currently, we don't have a quick link for feature request in settings. Maybe now we just offer both, bug report + feature request.

Classes/New Issue/IssueTemplates.swift
Huddie: Pulled it out, is it better?

Classes/Settings/SettingsViewController.swift
Sherlouk: It's a weird one, as a user I want to be able to do both easily! I'm thinking a button rename would be the best, but I'm lost on the naming for it!

Classes/New Issue/IssueTemplates.swift
Huddie: Haha @Sherlouk if we added more details into the bug report that could be a cool way to allows other repos to make use of the general bug report info. But i don't know what useful data we could add to the bug report that benefits any repo.

Classes/New Issue/IssueTemplates.swift
BasThomas: As in you can have `.github/issue_template`

Classes/Settings/SettingsViewController.swift
BasThomas: "Create an issue"?

Classes/New Issue/IssueTemplates.swift
Huddie: Figured this out, got it from the RepositoryDetails

Classes/New Issue/IssueTemplates.swift
Huddie: @BasThomas @Sherlouk I did a bare bones test adding a file under .github/ISSUE_TEMPLATE and then under .github/issue_template. First one worked, the second one did not. I think it may have to be under ISSUE_TEMPLATE unless someone knows of a counter example

Sherlouk: So for context my logic was coming from [this official article](https://help.github.com/articles/about-issue-and-pull-request-templates/) but is more about the singular templates as apposed to the new pick a template chooser (as I mentioned in the accompanying issue, this is step 2).

> Issue template filenames are not case sensitive, and can have an extension such as .md or .txt.

If you've tested it we might be safe for this iteration, worst case we raise a bug and fix it at a later date! Ship and iterate!

Sherlouk: Think this get's a bit confusing because it depends on exactly what the repo is doing but I think as per my comment in the accompanying issue ("step 2") is that single issue templates are not case sensitive and have a lot of different formats but potentially the new multi issue templates require the all caps directory?

https://help.github.com/articles/manually-creating-a-single-issue-template-for-your-repository/

Good article above on the topic showing that the markdown files can be in a bunch of places, and I believe are definitely not case sensitive

Huddie: @Sherlouk Ah I think I understand. The actual filename isn't an issue here because we look at all files in the folder and request those filenames, so they will always match perfectly. The only issue would be the ISSUE_TEMPLATE folder as that's the only hardcoded path. I agree that maybe we ship and fix if there's a bug report. Waiting on Ryan for a few things before I remove the WIP and finish up

Huddie:
* Uses the repos default branch now
* Moved path into constants

Classes/New Issue/IssueTemplates.swift
Huddie: @Sherlouk Just saw the link, so there's a difference between single issues vs multiple issues it appear (As you mentioned above). I still agree that maybe we go with this and as we learn more about what works etc we fix. If we can't find your templates we still show a blank New Issue screen so it won't have negative effects.

Huddie:
Added completion block so the caller knows when the template UIAlertView is about to be displayed. This allows for showing a spin indicator in the calling view. I implemented that in the settings view.

ijm8710:
Curious, so the way it works is to provide a framework on what you’re reporting? Is it the user’s job to replace the template with their report? Is there any way to set it up so it prompts you, this is what you enter here without the need to delete? This is super cool either way but was just curious so that the template is still a reference but also streamlines the input

Huddie:
>Curious, so the way it works is to provide a framework on what you’re reporting? Is it the user’s job to replace the template with their report? Is there any way to set it up so it prompts you, this is what you enter here without the need to delete? This is super cool either way but was just curious so that the template is still a reference but also streamlines the input

@ijm8710 if you could move this comment to #2393 this thread is more of the release of the feature, that thread is about the feature itself

Classes/New Issue/IssueTemplates.swift
Huddie: Switched checking titles to checking repo owner, now only GitHawkApp will have the bug report dump.

Classes/New Issue/IssueTemplates.swift
```diff
@@ -0,0 +1,230 @@
...
+        with templates: [IssueTemplate],
+        details: IssueTemplateDetails) -> UIAlertController {
+
+        let alertView = UIAlertController(
+            title: NSLocalizedString("New Issue", comment: ""),
+            message: NSLocalizedString("Choose Template", comment: ""),
+            preferredStyle: .alert
+        )
+
+        for template in templates {
```
Sherlouk: Correct me if I'm wrong, there appears to be no way to create an empty issue anymore? Is that correct?

As per the image below, GitHub has a button which allows you to just create an empty issue without a template!

image

Classes/New Issue/NewIssueTableViewController.swift
Sherlouk: Would we not just rather this be nil if there is no template? Feels like a tidier solution in my opinion

Huddie: @Sherlouk Updated!

Sherlouk: @Huddie Swish! If you give us a shout once you're looking for another review and I can do a full sweep at that point, hard to keep track of all these individual discussions at this point 😅

Huddie:
@Sherlouk Looking for a second review. If you could give some regex feedback I'd appreciate it. I'm pretty new to it. Maybe @BrianLitwin has some improvement suggestions on that.

Huddie:
@rnystrom I think Bas pinged you on some reviews above. I believe the 2 topics were:

- AlertView vs ActionSheet
   - https://github.com/GitHawkApp/GitHawk/pull/2400#discussion_r230915222
- If we want to pull out presenter logic
   - https://github.com/GitHawkApp/GitHawk/pull/2400#discussion_r230888506
Feedback would be appreciated when you get a chance as I think this is close to being ready

Classes/Settings/SettingsViewController.swift
```diff
@@ -154,19 +154,25 @@ NewIssueTableViewControllerDelegate, DefaultReactionDelegate {
...
-                owner: "GitHawkApp",
-                repo: "GitHawk",
-                signature: .bugReport
-            ) else {
-                Squawk.showGenericError()
-                return
+        showTemplateOptions()
+    }
+
+    func showTemplateOptions() {
```
jdisho: In my understanding, you are not using this `func` anywhere else except `onReportBug`. Why not putting the implementation in `onReportBug`?

Classes/Utility/String+Regex.swift
```diff
@@ -0,0 +1,34 @@
...
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+
+extension String {
+    func matches(
+        regex: String) -> [String] {
```
jdisho: IMO, the new line makes it less unreadable.

Classes/Views/SpinnerTableCell.swift
```diff
@@ -0,0 +1,35 @@
...
+//
+//  Created by Ehud Adler on 11/9/18.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import UIKit
+
+final class SpinnerTableCell: StyledTableCell {
+
+    let activity = UIActivityIndicatorView(activityIndicatorStyle: .gray)
```
jdisho: Are you using this outside this class? If not, make it private :)

Classes/Views/SpinnerTableCell.swift
```diff
@@ -0,0 +1,35 @@
...
+    var indicatorHolder: UITableViewCellAccessoryType?
```
jdisho: Are you using this outside this class? If not, make it private :)

Classes/Views/SpinnerTableCell.swift
```diff
@@ -0,0 +1,35 @@
...
+        super.awakeFromNib()
+        contentView.addSubview(activity)
+        activity.snp.makeConstraints { make in
+            make.top.bottom.equalTo(self)
+            make.trailing.equalTo(self).inset(20)
+        }
+        activity.hidesWhenStopped = true
+    }
+
+    public func showSpinner() {
```
jdisho: Why `public` and not `internal` (default) ?

Classes/Views/SpinnerTableCell.swift
```diff
@@ -0,0 +1,35 @@
...
+        indicatorHolder = self.accessoryType
+        self.accessoryType = .none
+        activity.startAnimating()
+    }
+    public func stopSpinner() {
```
jdisho: Why `public` and not `internal` (default) ?

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -69,17 +70,22 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
...
-        let storyboard = UIStoryboard(name: "NewIssue", bundle: nil)
+    static func create(
+        client: GithubClient,
+        owner: String,
+        repo: String,
+        template: String? = nil,
+        signature: IssueSignatureType? = nil
+        ) -> NewIssueTableViewController? {
+        
+        let storyboard = UIStoryboard(name: NSLocalizedString("NewIssue", comment: ""), bundle: nil)
```
Sherlouk: Storyboard name definitely shouldn't be localised!

Classes/Views/Constants.swift
```diff
@@ -52,5 +71,7 @@ enum Constants {
         static let reviewers = NSLocalizedString("Reviewers", comment: "")
         static let reviewGitHubAccess = NSLocalizedString("Review GitHub Access", comment: "")
         static let clear = NSLocalizedString("Clear", comment: "")
+        static let emptyString = NSLocalizedString("", comment: "")
```
Sherlouk: Empty string constant shouldn't be localized either

Classes/Utility/String+Regex.swift
```diff
@@ -0,0 +1,32 @@
...
+            let str = self as NSString
+            let results = regex.matches(
+                in: self,
+                range: NSRange(
+                    location: 0,
+                    length: str.length
+                )
+            )
+            return results.map { str.substring(with: $0.range) }
+        } catch let error as NSError {
```
Sherlouk: Unnecessary cast to NSError here

Classes/Repository/RepositoryViewController.swift
```diff
@@ -147,21 +147,17 @@ ContextMenuDelegate {
...
-            repo: repo.name,
-            signature: .sentWithGitHawk)
-        else {
-            Squawk.showGenericError()
-            return nil
-        }
-
-        newIssueViewController.delegate = self
-        weak var weakSelf = self
+        let action = UIAlertAction(title: "New Issue", style: .default) { _ in
```
Sherlouk: Title should be localized

Classes/New Issue/GithubClient+Template.swift
```diff
@@ -0,0 +1,130 @@
...
+        repo: RepositoryDetails,
+        filename: String,
+        completion: @escaping (Result<String>) -> Void
+        ) {
+
+        self.fetchFile(
+            owner: repo.owner,
+            repo: repo.name,
+            branch: repo.defaultBranch,
+            path: "\(Constants.Filepaths.githubIssue)/\(filename)") { (result) in
```
Sherlouk: If this gets rebased onto `master` I believe @rnystrom has created a URLBuilder to make this a bit nicer

Classes/New Issue/IssueTemplateHelper.swift
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    static func showIssueAlert(
+        with templates: [IssueTemplate],
+        details: IssueTemplateDetails) -> UIAlertController {
+
+        let alertView = UIAlertController(
+            title: NSLocalizedString("New Issue", comment: ""),
+            message: NSLocalizedString("Choose Template", comment: ""),
+            preferredStyle: .alert
```
Sherlouk: I'm thinking action sheet might be better here, but this is personal preference

Classes/Settings/Settings.storyboard
```diff
@@ -125,7 +125,7 @@
                             </tableViewSection>
                             <tableViewSection headerTitle="🦅 Githawk" id="3QH-oa-l2C">
                                 <cells>
-                                    <tableViewCell clipsSubviews="YES" contentMode="scaleToFill" selectionStyle="default" accessoryType="disclosureIndicator" indentationWidth="10" textLabel="kdx-s1-NXl" style="IBUITableViewCellStyleDefault" id="WwO-m1-s3q" customClass="StyledTableCell" customModule="Freetime" customModuleProvider="target">
+                                    <tableViewCell clipsSubviews="YES" contentMode="scaleToFill" selectionStyle="default" accessoryType="disclosureIndicator" indentationWidth="10" textLabel="kdx-s1-NXl" style="IBUITableViewCellStyleDefault" id="WwO-m1-s3q" customClass="SpinnerTableCell" customModule="Freetime" customModuleProvider="target">
```
Sherlouk: Title here hasn't been changed, did we want to do that?

Huddie: I subclassed StyledTableCell to allow an activity indicator.

Classes/New Issue/IssueTemplateHelper.swift
Huddie: Bas also mentioned this, I'll switch for now and we can switch back if need be

Classes/New Issue/GithubClient+Template.swift
Huddie: I think URLBuilder makes a full url (Https/http:// ... ). I'll check it out

Classes/Settings/Settings.storyboard
Sherlouk: I was more referring to the label inside of the cell, but can't add a review to a line outside of the change context 😉

Huddie: Oh haha! Ya ill fix that 😄Sorry

Classes/New Issue/GithubClient+Template.swift
Huddie: Ya URLBuilder seems to require a scheme. I also don't see why URLBuilder is better than a constant. Mind if I keep this for now?

Sherlouk: Not at all, hadn't looked into the URLBuilder stuff and wondered if it would've worked! All good for now

Sherlouk:
This looks good to me! 🎉 Great work!

Huddie:
@Sherlouk Thanks for the help!

jdisho:
Great work @Huddie 🎉

Huddie:
@rnystrom  I'm moving strings out of constants that aren't used elsewhere but I kind of like having the different regex stored in one place so we can see them easily. I know we've updated a few regex already and having them in one place could be nice. Should I:

- Move all regex out of constants back into the code
- Keep in the constants file
- Move to a new file where all regex is held (enum Regex .. )

?

rnystrom:
Hm. If the regret isn’t used again I say keep it inline. IMO its _more_ readable when it’s lives where it’s used.

Huddie:
@rnystrom Once I updated constants to remove
the unnecessary stuff I mentioned above is this good to go?

Huddie:
This is the updated gif.

BrianLitwin:
@Huddie Sorry I just saw your comment above- 
Tbh, I rely on unit tests with the regex's, not a deep understanding of regular expressions or Swift's implementation. 
These files all follow a similar pattern: 

[String+GithubEmoji](https://github.com/GitHawkApp/GitHawk/blob/master/Classes/Issues/Comments/Markdown/String+GitHubEmoji.swift)

[String+Shortlink](https://github.com/GitHawkApp/GitHawk/blob/master/Classes/Issues/Comments/Markdown/String+Shortlink.swift)

[String+StripHTMLComments](https://github.com/GitHawkApp/GitHawk/blob/master/Classes/Issues/Comments/Markdown/String+StripHTMLComments.swift)

[String + CustomBuilding](https://github.com/GitHawkApp/GitHawk/blob/master/Classes/Issues/Comments/Markdown/String+CustomBuilding.swift)

You'll see the pattern quickly: a public String method that returns what you need, eg `var templateNameAndDescription`, plus a single file with private properties and functions that contain all the logic. 

```Swift 

private let nameRegex = Try! NSRegularExpression(pattern: )
private let aboutRegex = Try! NSRegularExpression(pattern: )

extension String {
        var templateNameAndDescription: (name: String, description: String)? {
               // logic 
         }
}
```

Huddie:
Cool. I’m going to whip up a few unit tests and check out those files 

Huddie:
Does this need updating?

rnystrom:
I haven't been following along b/c of the other eyeballs that have been on this. Lmk if/when I should give it a look.

Huddie:
@rnystrom if you don’t mind taking a look through I’d appreciate it. 

Huddie:
@rnystrom Anything fixes required?

rnystrom:
Haven’t gotten to a review yet 😞

Huddie:
@rnystrom all good. Had to fix merge problems anyways 

FreetimeTests/TemplateRegexTests.swift
```diff
@@ -0,0 +1,123 @@
...
+        let template = """
+        -----
+        name: foo
+        about: bar
+        -----
+        Lorem ipsum dolor sit amet, mei no ignota commodo, dolorem luptatum in qui. Iusto regione inermis ea quo, te eos consul utroque scaevola. Ea per possim salutandi. At tota aperiam ius, facilisi recusabo accommodare vel cu. Cibo omnium utamur et pro, :apple: temporibus eu cum.\nCum paulo definitionem at, vel atqui tempor ad. Ex minimum percipit erroribus vis, qui erroribus persequeris te, et augue :chart_with_upwards_trend: mediocritatem eum. Et mei cibo dicta, vis reque :dancers: at, suas mutat corrumpit eos eu. Sint eruditi eos ei, ad pri esse omnes, dolorem blandit voluptatibus an ius.\nOption eligendi id eum, ne sea porro :small_red_triangle_down: assueverit, sed modo diam in. Id vim eripuit intellegebat, ad quo consul constituto, at vix natum interpretaris. Id his tritani vituperata, ea minim quidam aliquip vis, animal pertinacia ad sit. Mel impedit
+        """
+
+        var cleanedFile = ""
+        if let textToClean = template.matches(regex: "([-]{3,})([\\s\\S]*)([-]{3,})").first {
```
rnystrom: Regex should be shared w/ actual implementation!

Classes/New Issue/GithubClient+Template.swift
```diff
@@ -0,0 +1,150 @@
...
+struct IssueTemplateRepoDetails {
+    let repo: RepositoryDetails
+    let defaultBranch: String
+}
+
+struct IssueTemplateDetails {
+    let repoDetails: IssueTemplateRepoDetails
+    let session: GitHubUserSession?
+    let viewController: UIViewController
+    weak var delegate: NewIssueTableViewControllerDelegate?
```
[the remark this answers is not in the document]
Huddie: @rnystrom We can't reduce the networking calls because the title and description of files are at the top of the files. The file names are not equal to the template names. Let me know if that makes sense. I will make the changes regarding passing the templates to the VC and letting it take care of displaying the alert.

Huddie:
@rnystrom I'll take care of the nits and feedback. I replied above in regads to the number of networking calls. I don't think it can be reduced unfortuantly. 
If no templates are found we just open NewIssueViewController with nothing passed in. Basically a blank issue.

Classes/New Issue/IssueTemplateHelper.swift
```diff
@@ -0,0 +1,108 @@
...
+                            template: template.template,
+                            signature: details.repoDetails.repo.owner == Constants.GitHawk.owner ? .bugReport : .sentWithGitHawk
+                            ) else {
+                                assertionFailure("Failed to create NewIssueTableViewController")
+                                return
+                        }
+                        viewController.delegate = details.delegate
+                        let navController = UINavigationController(rootViewController: viewController)
+                        navController.modalPresentationStyle = .formSheet
+                        details.viewController.route_present(to: navController)
```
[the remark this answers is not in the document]
Huddie: @rnystrom Can you explain this a bit? If I switch to `route_present(to: viewController)` no navigation bar shows in the pushed view. I think I may not fully understand.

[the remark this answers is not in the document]
Huddie: Also separate question: Should it be route_detail? on iPad it pushes the new VC over the entire screen not into the detail section

Huddie:
@rnystrom Ready for another review whenever. Thanks for the last one!

Huddie:
@rnystrom 

Classes/New Issue/GithubClient+Template.swift
```diff
@@ -0,0 +1,115 @@
...
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+import GitHubAPI
+import GitHubSession
+import Squawk
+
+private let githubIssueURL = ".github/ISSUE_TEMPLATE"
+let tempalateRegex = ".github/ISSUE_TEMPLATE"
```
rnystrom: `private`?

Classes/New Issue/GithubClient+Template.swift
```diff
@@ -0,0 +1,115 @@
...
+
+    private func fetchTemplateFile(
+        repoDetails: IssueTemplateRepoDetails,
+        filename: String,
+        testingFile: String? = nil,
+        completion: @escaping (Result<String>) -> Void
+        ) {
+
+        // For Testing.
+        if let testingFile = testingFile { completion(.success(testingFile)) }
```
rnystrom: Let's newline the closure execution here to stay consistent w/ other `if`s

Classes/New Issue/GithubClient+Template.swift
```diff
@@ -0,0 +1,115 @@
...
+            path: "\(githubIssueURL)/\(filename)") { result in
+                switch result {
+                case .success(let file): completion(.success(file))
+                case .error(let error): completion(.error(error))
+                case .nonUTF8: completion(.error(nil))
+                }
+        }
+    }
+
+    func createTemplate(
```
rnystrom: nit: this needs a new name, I keep thinking that it's actually creating something. Can it just use the "fetch" naming used elsewhere?

Classes/New Issue/GithubClient+Template.swift
```diff
@@ -0,0 +1,115 @@
...
+        fetchFiles(
+            owner: repoDetails.owner,
+            repo: repoDetails.name,
+            branch: repoDetails.defaultBranch,
+            path: githubIssueURL) { result in
+                switch result {
+                case .success(let files):
+                    for file in files {
+                        templateGroup.enter()
+                        self.createTemplate(repoDetails: repoDetails, filename: file.name, completion: { result in
```
rnystrom: I'm going to block this PR on the multiple-requests part. It's just too wasteful IMO as-is. Some options:

1. Do a fetch for the directory and just show the file names in the list instead of the template title. Then fetch the contents of the template once its been selected.
2. Only fetch this info once then cache it. Always use the cache but allow the user to pull-to-refresh the templates list.

I favor option 1 personally.

Classes/New Issue/GithubClient+Template.swift
Sherlouk: also nit on the spelling 😄

Classes/New Issue/GithubClient+Template.swift
Sherlouk: Sorry to jump in on here @rnystrom, and I know it's a massive shift in direction but could I suggest a third option here?

If we use the v4 API (ooh bois he's said the danger word) we can actually do this with a single request which returns the contents of every template within the directory?

This request: (Notably, we would need to replace the default branch and repo owner/name)
```gql
query {
  repository(owner: "GitHawkApp", name:"GitHawk") {
    object(expression: "master:.github/ISSUE_TEMPLATE") {
      ... on Tree {
        entries {
          name
          object {
            ... on Blob {
              text
            }
          }
        }
      }
    }
  }
}
```

Returns us this response:
```json
{
  "data": {
    "repository": {
      "object": {
        "entries": [
          {
            "name": "bug_report.md",
            "object": {
              "text": "---\nname: Bug report  (etc etc, I've truncated this)"
            }
          },
          {
            "name": "feature_request.md",
            "object": {
              "text": "---\nname: Feature request (etc etc, I've truncated this)"
            }
          }
        ]
      }
    }
  }
}
```

One request, all the information? Opinion Ryan?

rnystrom: Lovd it!

PR #2414, opened 2018-11-10
Query for label on tap in RepoIssuesVC

BrianLitwin:
Connected to #2263

## Summary: 
In `RepositoryIssuesViewController`, when a user taps a label, the label is appended to the searchBar text and the query is fetched. 

## Test Notes: 
Manually: 
* tapping on a label adds the appropriate query to the search bar and fetches the appropriate result. 
* tapping on a label that is already contained in the query doesn't append duplicate labels 
* tapping a point beside the last label in the LabelListView is passed through and an IssuesVC is pushed. Ie, the LabelListView doesn't eat touches inside it's frame where there are no label cells. 

## How: 
There's a lot of chaining delegates involved. Reviewers: it will help to look at this view hierarchy:  

* RepositoryIssuesViewController ( BaseListViewController )
  * SearchBarSectionController ( ListSwiftSectionController )
    * SearchBarCell ( UICollectionViewCell )
  * RepositorySummarySectionController ( ListSwiftSectionController )
    * RepositorySummaryCell ( SelectableCell )
      * LabelListView ( UICollectionView )

The label is inside the `LabelListView`. It tells `RepositorySummaryCell`, which tells `RepositorySummarySectionController`, which tells `RepositoryIssuesViewController` that it has been tapped. 
Then: 
The `RepoIssueViewController` then tells `SearchBarSectionController` to tell `SearchBarCell` to set the searchBar text appropriately. Then the fetch proceeds as it currently would, as though the user entered the text in the cell. My hope is that this preserves a "one-way data flow".


BrianLitwin:
## Manual Test Screen Vid
<br> 
It's kinda hard to see but this tests the items in Test Notes above:
<br> 

ezgif com-video-to-gif 4

Classes/Repository/RepositoryIssuesViewController.swift
```diff
@@ -151,5 +160,12 @@ SearchBarSectionControllerDelegate {
         }
         return "repo:\(owner)/\(repo) \(typeQuery) \(previousSearchString)".lowercased()
     }
-
+    
+    // Mark: LabelListViewTapDelegate
+    
+    func didTap(label: String) {
```
jdisho: I would suggest to put this in an extension.

Classes/Section Controllers/SearchBar/SearchBarCell.swift
```diff
@@ -53,6 +52,14 @@ final class SearchBarCell: UICollectionViewCell, UISearchBarDelegate {
         searchBar.text = query
         searchBar.placeholder = placeholder
     }
+    
+    // MARK: SetSearchBarTextDelegate
```
jdisho: Separate extension?

Classes/Section Controllers/SearchBar/SearchBarSectionController.swift
```diff
@@ -13,12 +13,17 @@ protocol SearchBarSectionControllerDelegate: class {
     func didChangeSelection(sectionController: SearchBarSectionController, query: String)
 }
 
-final class SearchBarSectionController: ListSwiftSectionController<String>, SearchBarCellDelegate {
+protocol SetSearchBarTextDelegate: class {
+    func append(query: String)
```
jdisho: Why not calling `didAppend(query:)` ? `append(query:)` (as a name) confuses me a bit.

Classes/Section Controllers/SearchBar/SearchBarSectionController.swift
```diff
@@ -45,5 +51,10 @@ final class SearchBarSectionController: ListSwiftSectionController<String>, Sear
         self.query = query
         self.delegate?.didChangeSelection(sectionController: self, query: query)
     }
-
+    
+    // MARK: SetSearchBarTextDelegate
+    
+    func append(query: String) {
```
jdisho: Separate extension?

BrianLitwin:
@jdisho Thanks for your review! I'm not 100% sure why this is, but I don't see the pattern of conforming to protocols in extensions used elsewhere in the app - I'm just trying to stay consistent with the rest of the codebase; I don't have any strong opinion on the issue. See [IssuesViewController](https://github.com/GitHawkApp/GitHawk/blob/master/Classes/Issues/IssueTextActionsView.swift) for an example of conforming to a lot of delegates inside the class.

rnystrom:
+1 on not using `extension` everywhere. I hit compiler errors early on in GitHawk's development from over-using them and lost a ton of time debugging + refactoring.

Classes/Views/LabelListView.swift
```diff
@@ -63,6 +67,18 @@ UICollectionViewDelegateFlowLayout {
         collectionView.frame = bounds
     }
 
+    override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+        guard let lastCell = collectionView.cellForItem(at: [0, labels.count - 1])
+            else { return self }
+        return point.x > lastCell.frame.maxX ? self : super.hitTest(point, with: event)
+    }
```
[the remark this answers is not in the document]
BrianLitwin: screenshot 2018-11-11 11 57 13

I highlighted the `LabelListView` frame in orange. The UicollectionView's frame is set to it's bounds. 

If you click inside the orange border on an area that's not a LabelCell, nothing happens. In the HitTest override, I'm trying to process taps that land on a label cell, and pass through taps to the right of the right-most label cell. 

Perhaps setting the CollectionView's frame is more appropriate, so that it's only side wide as it needs to be given the number of labels it presents?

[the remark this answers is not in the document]
BrianLitwin: Fudging around, something like this seems to send the taps to the right places: 

```Swift
// this is copied/pasted from existing methods to find width 
private var labelTextTotalWidth: CGFloat {
        let interitemSpacing = labels.count > 1 ? CGFloat(labels.count - 1) * Styles.Sizes.labelSpacing : 0.0
        return labels.reduce(0, { $0 + LabelListCell.size($1.name).width }) + interitemSpacing
    }

override func layoutSubviews() {
        super.layoutSubviews()
        collectionView.frame = CGRect(
            x: bounds.minX,
            y: bounds.minY,
            width: labelTextTotalWidth,
            height: bounds.height
        )
}

```

Classes/Section Controllers/SearchBar/SearchBarCell.swift
```diff
@@ -53,6 +52,14 @@ final class SearchBarCell: UICollectionViewCell, UISearchBarDelegate {
...
+    
+    func append(query: String) {
+        searchBar.text?.append(query)
+        guard let text = searchBar.text else { return }
+        delegate?.didChangeSearchText(cell: self, query: text)
```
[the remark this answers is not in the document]
BrianLitwin: No it doesn't appear to be triggered automatically: the text in the bar changes but no fetch or update happens.

Classes/Section Controllers/SearchBar/SearchBarSectionController.swift
```diff
@@ -14,11 +14,12 @@ protocol SearchBarSectionControllerDelegate: class {
 }
 
 final class SearchBarSectionController: ListSwiftSectionController<String>, SearchBarCellDelegate {
-
+    
+    static let listSwiftId = "header"
```
BrianLitwin: Okay with this name?

Classes/Section Controllers/SearchBar/SearchBarCell.swift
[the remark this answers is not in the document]
BrianLitwin: Yeah the delegate method triggered here on it's own. The text in the search bar changes but no fetch/update happens.

BrianLitwin:
Wanted to add this screenshot with the updated method used to calculate the CollectionView's frame. Here, the CollectionView's border is Orange: 

screenshot 2018-11-11 15 27 58

RomanSimSalabim:
**```
```
~~**``**~~
```
```**

RomanSimSalabim:
#2434 

kh00:
~~~~

PR #2427, opened 2018-11-12
Improve review-access prompt in inbox

rnystrom:


Huddie:
Why the removal from constants?

rnystrom:
We should only add to `Constants` when a string is reused

Huddie:
Cool that makes sense. Thanks 

PR #2510, opened 2018-12-01
Change UISwitches colour to blue in Settings 💅🏻

jdisho:
In this PR, I change the UISwitches colour in Settigns to blue. I think it makes more sense than green since the labelings of GitHawk are blue-ish. Additionally, I refactored some code in that group 😄

### DIFF
--- a/Classes/Settings/DefaultReactionDetailController.swift
+++ b/Classes/Settings/DefaultReactionDetailController.swift
@@ -22,16 +22,46 @@ class DefaultReactionDetailController: UITableViewController {
     @IBOutlet var heartCell: UITableViewCell!
     @IBOutlet var enabledSwitch: UISwitch!
 
-    weak var delegate: DefaultReactionDelegate?
+    weak var delegate: DefaultReactionDelegate? {
+        didSet {
+            tableView.reloadData()
+        }
+    }
 
     override func viewDidLoad() {
         super.viewDidLoad()
+
+        configureSwitch(withColor: UIColor.fromHex(Styles.Colors.Blue.medium))
         checkCurrentDefault()
-        tableView.reloadData()
     }
 
     override func numberOfSections(in tableView: UITableView) -> Int {
         return enabledSwitch.isOn ? 2 : 1
+    }
+
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: trueUnlessReduceMotionEnabled)
+        let cell = tableView.cellForRow(at: indexPath)
+
+        switch cell {
+        case thumbsUpCell: updateDefault(reaction: .thumbsUp)
+        case thumbsDownCell: updateDefault(reaction: .thumbsDown)
+        case laughCell: updateDefault(reaction: .laugh)
+        case hoorayCell: updateDefault(reaction: .hooray)
+        case confusedCell: updateDefault(reaction: .confused)
+        case heartCell: updateDefault(reaction: .heart)
+        default: return
+        }
+    }
+
+    // MARK: Private API
+    @IBAction private func toggleDefaultReaction(_ sender: Any) {
+        enabledSwitch.isOn ? updateDefault(reaction: .thumbsUp) : disableReaction()
+        updateSections()
+    }
+
+    private func configureSwitch(withColor color: UIColor) {
+        enabledSwitch.onTintColor = color
     }
 
     private func checkCurrentDefault() {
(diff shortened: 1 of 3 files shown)


rnystrom:
@jdisho can you share a screenshot?

jdisho:
simulator screen shot - iphone xs max - 2018-12-01 at 09 17 24
simulator screen shot - iphone xs max - 2018-12-01 at 09 17 32

Classes/Settings/DefaultReactionDetailController.swift
```diff
@@ -22,18 +22,48 @@ class DefaultReactionDetailController: UITableViewController {
     @IBOutlet var heartCell: UITableViewCell!
     @IBOutlet var enabledSwitch: UISwitch!
 
-    weak var delegate: DefaultReactionDelegate?
+    weak var delegate: DefaultReactionDelegate? {
+        didSet {
+            tableView.reloadData()
```
rnystrom: Why is this necessary?

Classes/Settings/SettingsViewController.swift
```diff
@@ -109,39 +110,38 @@ GitHubSessionListener {
...
+        default: return
         }
     }
 
     // MARK: Private API
 
-    func updateDefaultReaction() {
+    private func configureSwitches(withColor color: UIColor) {
+        [badgeSwitch, pushSwitch, signatureSwitch, markReadSwitch].forEach {
+            $0?.onTintColor = color
```
rnystrom: Could we just update this to use `UIAppearance` instead? Like these:

https://github.com/GitHawkApp/GitHawk/blob/master/Classes/Views/Styles.swift#L124-L136

Classes/Settings/SettingsViewController.swift
```diff
@@ -258,41 +258,41 @@ GitHubSessionListener {
         }
     }
 
-    @IBAction func onBadgeChanged() {
+    private func updateActiveAccount() {
+        accountsCell.detailTextLabel?.text = sessionManager.focusedUserSession?.username ?? Constants.Strings.unknown
+    }
```
rnystrom: why move?

Classes/Settings/SettingsViewController.swift
```diff
@@ -109,39 +110,38 @@ GitHubSessionListener {
...
+        case reviewAccessCell: onReviewAccess()
+        case accountsCell: onAccounts()
+        case githubStatusCell: onGitHubStatus()
+        case reviewOnAppStoreCell: onReviewOnAppStore()
+        case reportBugCell: onReportBug()
+        case viewSourceCell: onViewSource()
+        case setDefaultReaction: onSetDefaultReaction()
+        case signOutCell: onSignOut()
+        case tryTestFlightBetaCell: onTryTestFlightBeta()
+        default: return
```
rnystrom: Nice, I dig this. Can the default just `break` though? In case someone adds beneath it, might be a pain to debug.

Classes/Settings/DefaultReactionDetailController.swift
```diff
@@ -110,12 +104,11 @@ class DefaultReactionDetailController: UITableViewController {
...
     private func updateSections() {
-        tableView.performBatchUpdates({
-            if enabledSwitch.isOn {
-                self.tableView.insertSections(IndexSet(integer: 1), with: .top)
-            } else {
+        tableView.performBatchUpdates({ [weak self] in
+            guard let `self` = self else { return }
+            enabledSwitch.isOn ?
+                self.tableView.insertSections(IndexSet(integer: 1), with: .top) :
                 self.tableView.deleteSections(IndexSet(integer: 1), with: .top)
```
rnystrom: Can we leave this as an `if`? This is much harder to read

Classes/Settings/DefaultReactionDetailController.swift
```diff
@@ -110,12 +104,11 @@ class DefaultReactionDetailController: UITableViewController {
     }
 
     private func updateSections() {
-        tableView.performBatchUpdates({
-            if enabledSwitch.isOn {
-                self.tableView.insertSections(IndexSet(integer: 1), with: .top)
-            } else {
+        tableView.performBatchUpdates({ [weak self] in
+            guard let `self` = self else { return }
```
rnystrom: There's no retain cycle w/ `performBatchUpdates`, no need for `weak`

Also if there were, can avoid the `guard` with

```swift
self?.tableView.insertSections(...)
```

Much cleaner

(but should just remove the `weak` here)

jdisho:
I applied the changes as it was discussed @rnystrom 😊

PR #2532, opened 2018-12-06
remove "remove bookmarks" part of the signout message

### LINKED ISSUE #2530: Remove “bookmarks will be removed” on logout
<details>
<pre>
Version 1.26.0 (1543766544)
Device: iPhone X (iOS 12.1)
TestFlight: true
</pre>
</details>

romankl:
Right now the logout mentions that the bookmarks are removed during the logout process, but the `sessionManager` just signs out

fixes #2530

### DIFF
--- a/Classes/Settings/SettingsViewController.swift
+++ b/Classes/Settings/SettingsViewController.swift
@@ -203,8 +203,8 @@ GitHubSessionListener {
 
     private func onSignOut() {
         let title = NSLocalizedString("Are you sure?", comment: "")
-        let message = NSLocalizedString("All of your accounts will be signed out, and their bookmarks will be removed. Do you want to continue?", comment: "")
-        let alert = UIAlertController.configured(title: title, message: message, preferredStyle: .alert)
+        let message = NSLocalizedString("You will be signed out from all of your accounts. Do you want to sign out?", comment: "")
+        let alert = UIAlertController.configured(title: title, message: message, preferredStyle: .actionSheet)
         alert.addActions([
             AlertAction.cancel(),
             AlertAction(AlertActionBuilder {


Classes/Settings/SettingsViewController.swift
```diff
@@ -203,7 +203,7 @@ GitHubSessionListener {
 
     private func onSignOut() {
         let title = NSLocalizedString("Are you sure?", comment: "")
-        let message = NSLocalizedString("All of your accounts will be signed out, and their bookmarks will be removed. Do you want to continue?", comment: "")
+        let message = NSLocalizedString("All of your accounts will be signed out. Do you want to continue?", comment: "")
```
jdisho: Suggestion: _You will be signed out from all of your accounts. Do you want to sign out?_ 

The two buttons are `Cancel` and `Sign out` and  imo, the message  _Do you want to sign out?_ reflects the `Sign out` action.

romankl: @jdisho thanks for the help! I've changed it

Classes/Settings/SettingsViewController.swift
```diff
@@ -203,7 +203,7 @@ GitHubSessionListener {
 
     private func onSignOut() {
         let title = NSLocalizedString("Are you sure?", comment: "")
-        let message = NSLocalizedString("All of your accounts will be signed out, and their bookmarks will be removed. Do you want to continue?", comment: "")
+        let message = NSLocalizedString("You will be signed out from all of your accounts. Do you want to sign out?", comment: "")
         let alert = UIAlertController.configured(title: title, message: message, preferredStyle: .alert)
```
j-f1: Should this be an action sheet instead, like the delete-photo action sheet in Photos?

Huddie: 🤔 that does fit the app style and will look better when we add the (sign out, sign out of all,
Cancel) once we are able to do selective sign out

romankl: @Huddie should I prepare this part for an ActionSheet style control?

Huddie: Ya, I think ActionSheet makes sense. Fits the style and works better with the future selective sign-out option.

romankl: @Huddie I've changed it!
image
What do you think? It was just an idea to change the primary action string to `sign out of all`

Huddie: I’d leave it as “Sign out” for now

romankl: @Huddie I've reverted the change

PR #2535, opened 2018-12-07
Prefer Static String

Huddie:


rnystrom:
I actually want to start using one of those libraries that gen resource structs like Android R. This is fine, just weird a weird Xcode-only thing.

Huddie:
@rnystrom am I closing this? Sorry just a bit confused 

PR #2538, opened 2018-12-09
Open external links in Safari

### LINKED ISSUE #2517: Open external links in Safari
It would be handy to have a setting to open all external links (or indeed GitHub links that GitHawk don’t handle through native UI) in Safari instead of in-app WebView.
This is a setting that you can find in many iOS apps and no doubt would be useful addition to many users.

ismetanin:
Fixes #2517 

**Changes:**
Added setting for external links opening in Safari

**Notes:**
I'm not sure that `func presentSafari(url: URL)` is a proper name for our method. And also I have some doubts about `Defaults+ExternalLinks.swift` implementation

### DIFF
--- a/Classes/Settings/Defaults+ExternalLinks.swift
+++ b/Classes/Settings/Defaults+ExternalLinks.swift
@@ -10,7 +10,7 @@ import Foundation
 
 extension UserDefaults {
 
-    private static let defaultKey = "com.whoisryannystrom.freetime.is-need-open-external-links-in-safari"
+    private static let defaultKey = "com.whoisryannystrom.freetime.should-open-external-links-in-safari"
 
     var shouldOpenExternalLinksInSafari: Bool {
         get {
--- a/Classes/Settings/Defaults+ExternalLinks.swift
+++ b/Classes/Settings/Defaults+ExternalLinks.swift
@@ -1,0 +1,24 @@
+//
+//  Defaults+ExternalLinks.swift
+//  Freetime
+//
+//  Created by Ivan Smetanin on 08/12/2018.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+extension UserDefaults {
+
+    private static let defaultKey = "com.whoisryannystrom.freetime.is-need-open-external-links-in-safari"
+
+    var isNeedOpenExternalLinksInSafari: Bool {
+        get {
+            return bool(forKey: UserDefaults.defaultKey)
+        }
+        set {
+            set(newValue, forKey: UserDefaults.defaultKey)
+        }
+    }
+
+}
--- a/Classes/Settings/Defaults+ExternalLinks.swift
+++ b/Classes/Settings/Defaults+ExternalLinks.swift
@@ -12,7 +12,7 @@ extension UserDefaults {
 
     private static let defaultKey = "com.whoisryannystrom.freetime.is-need-open-external-links-in-safari"
 
-    var isNeedOpenExternalLinksInSafari: Bool {
+    var shouldOpenExternalLinksInSafari: Bool {
         get {
             return bool(forKey: UserDefaults.defaultKey)
         }
--- a/Classes/Settings/SettingsViewController.swift
+++ b/Classes/Settings/SettingsViewController.swift
@@ -46,7 +46,6 @@ GitHubSessionListener {
     @IBOutlet weak var pushSwitch: UISwitch!
     @IBOutlet weak var pushCell: UITableViewCell!
     @IBOutlet weak var pushSettingsButton: UIButton!
-    @IBOutlet weak var openExternalLinksLabel: SettingsLabel!
     @IBOutlet weak var openExternalLinksSwitch: UISwitch!
 
     override func viewDidLoad() {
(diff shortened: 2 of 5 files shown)


Classes/Settings/SettingsViewController.swift
```diff
@@ -46,6 +46,8 @@ GitHubSessionListener {
     @IBOutlet weak var pushSwitch: UISwitch!
     @IBOutlet weak var pushCell: UITableViewCell!
     @IBOutlet weak var pushSettingsButton: UIButton!
+    @IBOutlet weak var openExternalLinksLabel: SettingsLabel!
```
Huddie: Is this needed? Do you use this variable anywhere?

Classes/Settings/Defaults+ExternalLinks.swift
```diff
@@ -0,0 +1,24 @@
...
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+extension UserDefaults {
+
+    private static let defaultKey = "com.whoisryannystrom.freetime.is-need-open-external-links-in-safari"
+
+    var isNeedOpenExternalLinksInSafari: Bool {
```
Huddie: Maybe “shouldOpenExternal...” isNeed sounds a bit off to me.

Huddie (commented):
What will happen before any user defaults are set (before the user switches the settings updating the user default value)? Will it be true/false/nil?

ismetanin:
@Huddie by default bool value in UserDefaults in **false**. 

> Return Value
 The Boolean value associated with the specified key. If the specified key doesn‘t exist, this method returns false.

Details see [here](https://developer.apple.com/documentation/foundation/userdefaults/1416388-bool).

Classes/Settings/SettingsViewController.swift
ismetanin: Nope
Maybe it will be needed later for localization. But okay, now I remove it

Huddie:
@ismetanin cool 😎! 

Huddie:
@ismetanin you added a file but didn’t upload the project .pbxproj file. 

ismetanin:
@Huddie fixed

Classes/Settings/Defaults+ExternalLinks.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+//  Created by Ivan Smetanin on 08/12/2018.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+extension UserDefaults {
+
+    private static let defaultKey = "com.whoisryannystrom.freetime.is-need-open-external-links-in-safari"
```
BasThomas: Why don't we name this `should-open-external-links-in-safari` too? Not that big of a deal of course, just wondering why.

Classes/View Controllers/UIViewController+Safari.swift
```diff
@@ -12,8 +12,12 @@ import SafariServices
 extension UIViewController {
 
     func presentSafari(url: URL) {
-		guard let safariViewController = try? SFSafariViewController.configured(with: url) else { return }
-        route_present(to: safariViewController)
+        if UserDefaults.standard.shouldOpenExternalLinksInSafari {
+            UIApplication.shared.open(url)
```
BasThomas: Does this not need a `canOpenURL(_ url:)` check?

Classes/View Controllers/UIViewController+Safari.swift
```diff
@@ -12,8 +12,12 @@ import SafariServices
...
+        } else {
+            guard let safariViewController = try? SFSafariViewController.configured(with: url) else { return }
```
BasThomas: What about putting this in a `do/catch` and asserting if this fails?

Classes/View Controllers/UIViewController+Safari.swift
```diff
@@ -12,8 +12,12 @@ import SafariServices
...
+            route_present(to: safariViewController)
```
BasThomas: @rnystrom: why is this using snake case? 😮

Classes/View Controllers/UIViewController+Safari.swift
ismetanin: Good suggestion!

Classes/Settings/Defaults+ExternalLinks.swift
ismetanin: Forgot to fix

Classes/View Controllers/UIViewController+Safari.swift
ismetanin: It won't crash but in general, it would be good to check
Maybe put assertion here too?

BasThomas: Sounds good

Classes/View Controllers/UIViewController+Safari.swift
rnystrom: experimenting w/ prefix for a utils extension

ismetanin: @rnystrom Should I fix it?

rnystrom: Not a big deal right now IMO

rnystrom:
Love this. Has been requested multiple times!

CryptoGodfatherVA (commented):
#

Issue #2546, opened 2018-12-11
Deprecated status page


DaniSpringer:
Please link app to new GitHub status page

<details>
<summary>Bug Report Dump (Auto-generated)</summary>
<pre>
Version 1.25.0 (1542806489)
Device: iPhone XR (iOS 12.1.1)
TestFlight: false
</pre>
</details>

romankl:
The url changed to https://www.githubstatus.com/
Can I send a PR for this?

Huddie:
@romankl ya!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BasThomas:
Weird move 🤔

BasThomas:
[This](https://github.com/GitHawkApp/GitHawk/blob/083155077e06347f944e690f10385d793327ebca/Classes/Settings/SettingsViewController.swift#L147) would now be `https://www.githubstatus.com/`, without adding the `messages` path.

[This](https://github.com/GitHawkApp/GitHawk/blob/8452d73ab9aab9db3891dce102c4b1574acc95d0/Local%20Pods/GitHubAPI/GitHubAPI/GitHubAPIStatusRequest.swift#L20) one I think can stay; I don't see that the API path changed.

j-f1:
The old site will be removed in 7 days.

PR #2547, opened 2018-12-12
switch from the now deprecated to the new status page

romankl:
the "old" status page https://status.github.com is no deprecated and GitHub switched to a statuspage.io implementation with https://githubstatus.com. The page mentions the deprecation.

This PR switches the url to the new status page and changes the status API model for the new REST- response.

The new API replaced the `good` state with the state `none`. And has more fields, that are probably 
 not relevant for the App.

Previous response:
```json
{"status":"good","last_updated":"2018-12-06T17:09:57Z"}
```

New response:
```json
{"page":{"id":"kctbh9vrtdwd","name":"GitHub","url":"https://www.githubstatus.com","time_zone":"Etc/UTC","updated_at":"2018-12-12T05:38:30.535Z"},"status":{"indicator":"none","description":"All Systems Operational"}}
```

New page in SafariVC:
image

fixes: #2546

### DIFF
--- a/Classes/Settings/SettingsViewController.swift
+++ b/Classes/Settings/SettingsViewController.swift
@@ -82,21 +82,13 @@ GitHubSessionListener {
                 strongSelf.apiStatusView.isHidden = true
                 strongSelf.apiStatusLabel.text = NSLocalizedString("error", comment: "")
             case .success(let response):
-                let text: String
+                let text = response.data.status.description
                 let color: UIColor
                 switch response.data.status.indicator {
-                case .none:
-                    text = response.data.status.description
-                    color = Styles.Colors.Green.medium.color
-                case .minor:
-                    text = response.data.status.description
-                    color = Styles.Colors.Yellow.medium.color
-                case .major:
-                    text = response.data.status.description
-                    color = Styles.Colors.Red.medium.color
-                case .critical:
-                    text = response.data.status.description
-                    color = Styles.Colors.Red.medium.color
+                case .none: color = Styles.Colors.Green.medium.color
+                case .minor: color = Styles.Colors.Yellow.medium.color
+                case .major: color = Styles.Colors.Red.medium.color
+                case .critical: color = Styles.Colors.Red.medium.color
                 }
                 strongSelf.apiStatusView.isHidden = false
                 strongSelf.apiStatusView.backgroundColor = color
--- a/Classes/Settings/SettingsViewController.swift
+++ b/Classes/Settings/SettingsViewController.swift
@@ -139,7 +139,7 @@ GitHubSessionListener {
     }
 
     private func onGitHubStatus() {
-        guard let url = URLBuilder(host: "githubstatus.com").url
+        guard let url = URLBuilder(host: "www.githubstatus.com").url
             else { return }
         presentSafari(url: url)
     }
--- a/Classes/Settings/SettingsViewController.swift
+++ b/Classes/Settings/SettingsViewController.swift
@@ -88,7 +88,7 @@ GitHubSessionListener {
                 case .none: color = Styles.Colors.Green.medium.color
                 case .minor: color = Styles.Colors.Yellow.medium.color
                 case .major: color = Styles.Colors.Red.medium.color
-                case .critical: color = Styles.Colors.Red.medium.color
+                case .critical: color = Styles.Colors.Red.dark.color
                 }
                 strongSelf.apiStatusView.isHidden = false
                 strongSelf.apiStatusView.backgroundColor = color
--- a/Classes/Settings/SettingsViewController.swift
+++ b/Classes/Settings/SettingsViewController.swift
@@ -84,15 +84,18 @@ GitHubSessionListener {
             case .success(let response):
                 let text: String
                 let color: UIColor
-                switch response.data.status {
-                case .good:
-                    text = NSLocalizedString("Good", comment: "")
+                switch response.data.status.indicator {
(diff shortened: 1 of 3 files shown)


Classes/Settings/SettingsViewController.swift
```diff
@@ -84,8 +84,8 @@ GitHubSessionListener {
             case .success(let response):
                 let text: String
                 let color: UIColor
-                switch response.data.status {
-                case .good:
+                switch response.data.status.indicator {
+                case .none:
                     text = NSLocalizedString("Good", comment: "")
```
j-f1: Do you want to use the provided `description` key here instead?

romankl: I've changed it to the `description` JSON property:

image

Classes/Settings/SettingsViewController.swift
```diff
@@ -84,15 +84,18 @@ GitHubSessionListener {
             case .success(let response):
                 let text: String
                 let color: UIColor
-                switch response.data.status {
-                case .good:
-                    text = NSLocalizedString("Good", comment: "")
+                switch response.data.status.indicator {
+                case .none:
+                    text = response.data.status.description
                     color = Styles.Colors.Green.medium.color
```
viktorgardart: You could add a computed property on `StatusType` called color and this `switch` could be removed.

Classes/Settings/SettingsViewController.swift
Huddie: How does this look on iPhone 5SE?

Classes/Settings/SettingsViewController.swift
```diff
@@ -84,15 +84,18 @@ GitHubSessionListener {
             case .success(let response):
                 let text: String
                 let color: UIColor
-                switch response.data.status {
-                case .good:
-                    text = NSLocalizedString("Good", comment: "")
+                switch response.data.status.indicator {
+                case .none:
+                    text = response.data.status.description
```
j-f1: This should be able to be pulled out of the switch (DRY).

romankl: Sorry for this mistake! Will fix it and thank you for the review!

Classes/Settings/SettingsViewController.swift
```diff
@@ -84,15 +84,18 @@ GitHubSessionListener {
...
-                    text = NSLocalizedString("Minor", comment: "")
+                    text = response.data.status.description
                     color = Styles.Colors.Yellow.medium.color
                 case .major:
-                    text = NSLocalizedString("Major", comment: "")
+                    text = response.data.status.description
+                    color = Styles.Colors.Red.medium.color
+                case .critical:
+                    text = response.data.status.description
                     color = Styles.Colors.Red.medium.color
```
BasThomas: Should we introduce / can we use another color to disambiguate between `.major` and `.critical`?

Local Pods/GitHubAPI/GitHubAPI/GitHubAPIStatusRequest.swift
```diff
@@ -9,15 +9,22 @@
...
-    public enum StatusType: String, Codable {
-        case good, minor, major
+    public let status: Status
+}
+
+public struct Status: Codable {
+    public let indicator: StatusType
+    public let description: String
+    
+    public enum StatusType: String,Codable, CodingKey {
```
BasThomas: Super nit: no space after `String,`

BasThomas: Can we rename the `case none` to `case normal` or similar? `.none` can silently be interpreted as `Optional.none` when working with an optional enum of this type, which can be hard to debug. Would require manually handling the decoding though.

BasThomas:
Nice work @romankl!

BasThomas:
Hi @romankl, do you want to take another look at the comments? Can I help out?

BasThomas:
From @zhongwuzw:

> https://github.com/GitHawkApp/GitHawk/pull/2547/files#diff-9f0bae268086ea5dfc986d1a77f28abaL147, change `githubstatus.com` to `www.githubstatus.com` may more better, can reduce HTTP redirect.

Sherlouk:
@romankl Any interest in finishing this little bit of work of as per the above remarks?

@rnystrom @BasThomas Just as a reminder, the deadline is moving in quickly (current page shuts down on Monday)

Huddie:
@BasThomas @Sherlouk What’s left on this? Just www?

BasThomas:
That and some more comments. Will work on that this weekend and merge. 

Huddie:
Okay haha, I was gonna offer the same. Just to get it out. Enjoy!

Classes/Settings/SettingsViewController.swift
BasThomas: I decided not to do this as this would need `StatusType` to import `UIKit`, which feels unnecessary, and a weird dependency in the networking layer.

Local Pods/GitHubAPI/GitHubAPI/GitHubAPIStatusRequest.swift
BasThomas: Decided against changing this for now as it would also "change" the API, which might be confusing, and certainly would be more confusing in most cases.

BasThomas:
Changes addressed, thanks so much for the initial push @romankl!

Issue #198, opened 2017-07-28
Dark mode / night theme


BasThomas:
Would love this. Definitely not too important though.

shaps80:
Haha I wondered how long it would take for someone to request this.

BasThomas:
This could go together with a dark app icon as well

Sherlouk:
@Huddie @BrianLitwin @rnystrom 

Anybody got a mockup design or palette they think would work nice as a dark mode? Need some inspiration before I go rainbow city on the app :trollface:

kocheck:
@Sherlouk This is a huge interest of mine 👍 I might spin something up this weekend to play with.

Sherlouk:
I spent a week turning my Slack into dark mode for a reason 😂Dark mode should be a requirement for all apps :trollface: 

Can't wait to see what you make!

rnystrom:
Very very interested. The hardest part for us will be text colors. Might have to update StyledTextKit to support themes or something so we can regen attributed strings without rebuilding.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ggliv:
I understand that this is an ancient thread but what’s the status on this? I’d love to see this happen.

BasThomas:
Hey @ggliv, thanks for bringing it up! Let's wait and see what Apple announces on Monday... and see where to go from there.

There's also still a work-in-progress pull request for a dark mode [here](https://github.com/GitHawkApp/GitHawk/pull/2553).

Huddie:
Seems like dark mode will kinda come for free with iOS 13. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jdisho:
I guess, we still need to support dark mode for older iOS versions, right? 🤔

Sherlouk:
Didn't take you folks long to start updating this 😅 Personally I don't think we should actively try to support dark mode on iOS 12-

Dark mode exclusive to iOS 13+ using OS systems sounds like a simpler solution and less maintenance longer term.

iDjay:
Would love to see one too :)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ideologysec:
Definitely throwing my vote in for a dark mode in iOS 13+ (previous versions not so much; let's aim for the MVP of getting it implemented, and if there's enough demand it can maybe be backported...)

ggliv:
@BasThomas Any news?

[...]

PR #2553, opened 2018-12-17
Dark Mode [WIP]

nathantannar4:
**Please note this is still a WIP**

This PR fixes issue #198

## Description

These changes will allow views/objects to subscribe to theme changes so their colours can be changed dynamically. I have started doing some basic theming so if you run the app you will be able to flip the "Dark Mode" toggle in the Settings and see things become more night friendly. What I have yet to implement is dynamic colour changing for the `StyledTextView` as there seemed to be some chatter in #198 that `StyledTextKit` may need some updates first to support it.

## What Needs to be Discussed
- Should the theme changes be visible right away for require app restart? (then text does not need to be redrawn)
- What colour palette should dark mode have? (I just went with black and dark grey for now)
- Is `ThemeChangeListener` the best approach if we do want dynamic changes?

## Key Changes

Aside from changing some background and text colours the only noteworthy addition is `ThemeChangeListener` and `Appearance` found in `Theme.swift`

When a view/object conforms to `ThemeChangeListener` they can call `registerForThemeChanges` in their `init` for `viewDidLoad()`. This allows for dynamic theme changes across views through Apples `NotificationCenter` API. Each object that registers for theme changes will have a `ThemeChangeNotifier` set through the use of `objc_setAssociatedObject`

```swift
protocol ThemeChangeListener: AnyObject {
    func themeDidChange(_ theme: Theme)
}
```

The theme can then be changed with `Appearance.setCurrentTheme(_ theme: Theme, animated: Bool)` or read with `Appearance.currentTheme`


Classes/Issues/IssuesViewController.swift
```diff
@@ -239,6 +239,12 @@ IssueManagingContextControllerDelegate {
         )
     }
 
+    func themeDidChange(_ theme: Theme) {
+        view.backgroundColor = Styles.Colors.background
```
ismetanin: In my opinion, this code looks heavy
I've used this library for setup themes in my recent project: https://github.com/jiecao-fm/SwiftTheme
Maybe we will take some ideas from there?
It will be more readable, for example, we can simply write something like this:
```swift
view.themeBackgroundColor = Theme.Colors.background
```

nathantannar4: What does the `themeBackgroundColor` property do?
If your talking about where the color was derived from, I was following the preexisting constants that are in the `Styles`

ismetanin: `themeBackgroundColor` sets the background color of a view 🙂
I mean in your implementation we have to listen to the theme change in each file and write code for this. With SwiftTheme we won't need this

Classes/Utility/Theme.swift
```diff
@@ -0,0 +1,92 @@
...
+            options: [.transitionCrossDissolve],
+            animations: {
+                Appearance.currentTheme = theme
+                UserDefaults.standard.set(theme.rawValue, forKey: kPreferredThemeRawValueKey)
+                NotificationCenter.default.post(name: kThemeDidChange, object: theme)
+        })
+    }
+}
+
+private let kThemeDidChange = NSNotification.Name("kThemeDidChange")
```
jdisho: Group private properties together. Add this after `kPreferredThemeRawValueKey`

Classes/Utility/Theme.swift
```diff
@@ -0,0 +1,92 @@
...
+    @objc private func notificationReceived(_ notification: Notification?) {
+        block(notification)
+    }
+
+    deinit {
+        NotificationCenter.default.removeObserver(self)
+    }
+}
+
+protocol ThemeChangeListener: AnyObject {
```
jdisho: Why not putting `ThemeChangeListener` to a separate file?

Classes/View Controllers/NavigationController.swift
```diff
@@ -0,0 +1,30 @@
...
+//  NavigationController.swift
+//  Freetime
+//
+//  Created by Nathan Tannar on 2018-12-16.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import UIKit
+
+final class NavigationController: UINavigationController, ThemeChangeListener {
```
jdisho: Suggestion: What about calling `NavigationController`, `GHNavigationController`? (GH for GitHawk)

Classes/View Controllers/TabBarController.swift
```diff
@@ -0,0 +1,28 @@
...
+//  TabBarController.swift
+//  Freetime
+//
+//  Created by Nathan Tannar on 2018-12-16.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import UIKit
+
+final class TabBarController: UITabBarController, ThemeChangeListener {
```
jdisho: Suggestion: `GHTabBarController`?

jdisho (commented):
Some suggestions from my side. I will check more thoroughly later.

Classes/Utility/Theme.swift
nathantannar4: That was the plan, and for all your suggestions about this file. II didn't want to invest too heavy in formatting these if architecturally it was decided to use another method of theming

Classes/View Controllers/NavigationController.swift
nathantannar4: I didn't see GH[ClassName] anywhere else so I didn't think this was a convention that was being followed

Classes/Systems/LocalNotificationsCache.swift
```diff
@@ -19,7 +19,7 @@ final class LocalNotificationsCache {
 
     private let path: String
     private lazy var queue: FMDatabaseQueue = {
-        return FMDatabaseQueue(path: self.path)
+        return FMDatabaseQueue(path: self.path)!
```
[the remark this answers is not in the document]
nathantannar4: Yes I'll need to revert this, I think one of my pods was ahead a version or two and it was an optional initializer so I needed this to compile

Classes/Issues/IssuesViewController.swift
nathantannar4: @ismetanin The problem is even with `SwiftTheme` we would have to listen for theme changes, as GitHawk uses a lot of attributedStrings and so some redrawing of those might be required on theme change

YangW0223:
- - - - 

YangW0223:
````

**Comment by rnystrom** [comment]
Very cool! @nathantannar4 what's the plan for sequencing changes here and making updates to text rendering?

**Comment by nathantannar4** [comment]
I'm going to keep chipping away at making all the views light/dark themed but wanted to hear your input on the text colour theming.  

I see two approaches:
1. Re-Build each attributed string on theme change (less optimal)
2. You mentioned something in #198 about "regen attributed strings without rebuilding" with `StyledTextKit`, I haven't dove too much into that framework so I'm not sure what that would entail. But I presume this to be more optimal.

File: Classes/Utility/Theme.swift [hunk 7]
```diff
@@ -0,0 +1,92 @@
...
+    case light = 0
+    case dark = 1
+
+    static var `default`: Theme {
+        let rawValue = UserDefaults.standard.integer(forKey: kPreferredThemeRawValueKey)
+        return Theme(rawValue: rawValue) ?? .light
+    }
+}
+
+final class Appearance {
```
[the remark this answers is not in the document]
nathantannar4: Open to name changes if the general consensus is that its more clear.

Classes/Login/LoginSplashViewController.swift
```diff
@@ -68,6 +76,15 @@ final class LoginSplashViewController: UIViewController {
         super.viewDidLoad()
         state = .idle
         signInButton.layer.cornerRadius = Styles.Sizes.cardCornerRadius
+        view.backgroundColor = Appearance.currentTheme == .light
```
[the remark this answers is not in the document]
nathantannar4: 👍

Classes/Utility/Theme.swift
```diff
@@ -0,0 +1,92 @@
+//
+//  Theme.swift
+//  Freetime
+//
+//  Created by Nathan Tannar on 2018-12-15.
+//  Copyright © 2018 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
```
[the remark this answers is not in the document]
nathantannar4: To keep some properties private, or in this case fileprivate they need to be in the same file

Grinnerbren:
```

```

Shadyfire2020:
##2895 

Shadyfire2020:

F3000:
when can we expect this feature ?

Shadyfire2020:
```

```

Shadyfire2020:

Issue #2537, opened 2018-12-08
Stray oddities from bookmark refactor/iCloud integration


ijm8710:
Relating to #2483 bookmark refactor, we’ve seen cases where bookmarks are:

1) losing prior ordering
2) new bookmarks add to bottom , not top of list (if intended, 👌, just curious if you intentionally wanted to reverse that)
3) lost bookmarks-perhaps mostly coinciding with private repos

Sherlouk:
Ok so I downloaded an older build off of TF and a few things of note:

1. Should we delete the local storage of bookmarks once they're migrated? One big nit I have with some apps is when they store files that can't be deleted without reinstalling the app and risk losing other things. I know the file is small in this case but yano?

2. I've just checked which ones disappeared and I think it looked more dramatic because I have an issue bookmarked which wraps onto 10 lines so takes up a lot of the screen and has since been deleted (Fully understand this one being gone)! 

The main thing though is the removal of one of my private repos stored in a 2FA org.

On v1.26 this same repo is unavailable, it says 'There was an error loading the repository.' which is likely why it's not showing with the new live updating logic. However on v1.25 I can access this repo with no issues - wonder what change caused this as it doesn't seem related to bookmarks refactor? @rnystrom

rnystrom:
It’s related. I moved the repo “has issues enabled” and “default branch” settings to a fetch in the repo VC. This API auth difference between 3 and 4 is really frustrating.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rnystrom:
Details:

https://github.community/t5/GitHub-API-Development-and/Getting-quot-OAuth-App-access-restrictions-quot-for-GraphQL/m-p/14556#M262

rnystrom:
I'm honestly thinking about **requiring** PAT login as the only option.

Sherlouk:
Hmm yeah that's really frustrating!

I mean I'm completely up for just using PAT if it gives us more flexibility and less issues - but I do think we need to work out a nice way to explain this to users when shipping the update.

Wouldn't jump to my mind to switch to using PAT and don't want it to look like we've just logged users out or something!

rnystrom:
Nope def wouldn't do that. Another option would be that when we get an OAuth error, show a more detailed alert on how to add a PAT to your account (requires updating account settings so you can override current login w/ a PAT tho).

[...]

PR #2557, opened 2018-12-24
Alert on partial migration for bookmarks

rnystrom:
Partly addresses #2537 to at least inform the user that something went wrong

### DIFF
--- a/Classes/Bookmark/BookmarkCloudMigrator.swift
+++ b/Classes/Bookmark/BookmarkCloudMigrator.swift
@@ -11,6 +11,8 @@ import Foundation
 enum BookmarkCloudMigratorClientResult {
     case noMigration
     case success([String])
+    // when oauth restrictions fail some migrations
+    case partial(success: [String], errors: Int)
     case error(Error?)
 }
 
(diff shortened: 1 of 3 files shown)


ijm8710:
Who would this affect going fwd? Wouldn’t all new users not be in-scope for the migration and one would assume almost all existing TF users already migrated?

ijm8710:
Does this mean that the pre-bookmark list is still retained locally (ie when Sherlock said he had deleted bookmarks he would actually get an alert now saying that some bookmarks were not fully migrated?)
Assuming you do get the alert is there any way to down the line see what those not migrated bookmarks were?
Is there any way to retrieve bookmark order pre-migration or is that lost?

ijm8710:
@rnystrom little confused. Are you saying with new testflight we should install the App Store version on top of Tf and current bkmks should be retained before re-migrating?

Do you expect the former bookmark order and those “hidden” bookmarks to show in both the old AppStore and this new testflight version after?

rnystrom:
Yup install App Store then TestFlight to make sure migration works or alerts when something is lost

ijm8710:
Did this. Went from 41 to 38 bookmarks back to 41. 

1. Did bookmarks showing when public githawk was installed on testflight retain the bookmarks order before testflight versions went into affect (meaning retain a save-state from a month or so ago?)
2. More importantly is there any way to get back the prior order (at least now I have screenshot so I can do manual but just curious)

Most importantly, the migration seemed to work but I did have no alerts pop up

PR #2558, opened 2018-12-24
Changing request review message to self when the same user

LucianoPAlmeida:
Hey guys :))

This is a simple improvement on the `IssueRequest` changing the message when the users are the same. The same message that appears on the GitHub web interface.
From:
simulator screen shot - iphone x - 2018-12-24 at 12 22 44
To:
simulator screen shot - iphone x - 2018-12-24 at 12 23 39

Also, some warnings removed :)

### DIFF
--- a/Classes/Issues/Merge/MergeButton.swift
+++ b/Classes/Issues/Merge/MergeButton.swift
@@ -37,7 +37,7 @@ final class MergeButton: UIView {
         layer.cornerRadius = Styles.Sizes.avatarCornerRadius
 
         let image = UIImage(named: "chevron-down").withRenderingMode(.alwaysTemplate)
-        let optionButtonWidth = (image.size.width ?? 0) + (2 * Styles.Sizes.gutter)
+        let optionButtonWidth = image.size.width + (2 * Styles.Sizes.gutter)
         optionIconView.contentMode = .center
         optionIconView.image = image
         optionIconView.isAccessibilityElement = true
--- a/Classes/Issues/Request/IssueRequestModel.swift
+++ b/Classes/Issues/Request/IssueRequestModel.swift
@@ -42,6 +42,63 @@ final class IssueRequestModel: ListDiffable {
         self.date = date
         self.event = event
 
+        let styledString: StyledTextString
+        if actor == user {
+            styledString = IssueRequestModel.buildSelfString(actor: actor,
+                                                             user: user,
+                                                             date: date,
+                                                             event: event)
+        } else {
+            styledString = IssueRequestModel.buildString(actor: actor,
+                                                         user: user,
+                                                         date: date,
+                                                         event: event)
+        }
+
+        self.string = StyledTextRenderer(
+            string: styledString,
+            contentSizeCategory: contentSizeCategory,
+            inset: UIEdgeInsets(
+                top: Styles.Sizes.inlineSpacing,
+                left: 0,
+                bottom: Styles.Sizes.inlineSpacing,
+                right: 0
+            )
+        ).warm(width: width)
+    }
+
+    static private func buildSelfString(actor: String,
+                                        user: String,
+                                        date: Date,
+                                        event: Event) -> StyledTextString {
+        let phrase: String
+        switch event {
+        case .assigned: phrase = NSLocalizedString(" self-assigned this", comment: "")
+        case .unassigned: phrase = NSLocalizedString(" removed their assignment", comment: "")
+        case .reviewRequested: phrase = NSLocalizedString(" self-requested a review", comment: "")
+        case .reviewRequestRemoved: phrase = NSLocalizedString(" removed their request for review", comment: "")
+        }
+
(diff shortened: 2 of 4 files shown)


Classes/Issues/Request/IssueRequestModel.swift
```diff
@@ -42,6 +42,63 @@ final class IssueRequestModel: ListDiffable {
         self.date = date
         self.event = event
 
+        let styledString: StyledTextString
+        if actor == user {
+            styledString = IssueRequestModel.buildSelfString(actor: actor,
```
Huddie: nit: newline this like
```
buildSelfString(
       actor: actor,
       ...
)
```

Classes/Issues/Request/IssueRequestModel.swift
```diff
@@ -42,6 +42,63 @@ final class IssueRequestModel: ListDiffable {
...
+                                                             user: user,
+                                                             date: date,
+                                                             event: event)
+        } else {
+            styledString = IssueRequestModel.buildString(actor: actor,
```
Huddie: same as above

Classes/Issues/Request/IssueRequestModel.swift
```diff
@@ -42,6 +42,63 @@ final class IssueRequestModel: ListDiffable {
...
+            inset: UIEdgeInsets(
+                top: Styles.Sizes.inlineSpacing,
+                left: 0,
+                bottom: Styles.Sizes.inlineSpacing,
+                right: 0
+            )
+        ).warm(width: width)
+    }
+
+    static private func buildSelfString(actor: String,
```
Huddie: newline this (same as above) 
```
function(
    param1: ...,
    param2 ...,
) -> return {
```

Classes/Issues/Request/IssueRequestModel.swift
```diff
@@ -42,6 +42,63 @@ final class IssueRequestModel: ListDiffable {
...
+                ])
+            ))
+            .restore()
+            .add(text: phrase)
+            .add(text: " \(date.agoString(.long))", attributes: [MarkdownAttribute.details: DateDetailsFormatter().string(from: date)])
+
+        return builder.build()
+    }
+
+    static private func buildString(actor: String,
```
Huddie: same as above

Classes/Issues/Request/IssueRequestModel.swift
```diff
@@ -42,6 +42,63 @@ final class IssueRequestModel: ListDiffable {
...
+                top: Styles.Sizes.inlineSpacing,
+                left: 0,
+                bottom: Styles.Sizes.inlineSpacing,
+                right: 0
+            )
+        ).warm(width: width)
+    }
+
+    static private func buildSelfString(actor: String,
+                                        user: String,
```
Huddie: `user` is never used inside this function so no need to pass it in. Instead of passing in `actor` you may want to pass in `user` and switch the use of `actor` later on to `user`. I think it reads better as `user`

Classes/Issues/Merge/MergeButton.swift
```diff
@@ -37,7 +37,7 @@ final class MergeButton: UIView {
         layer.cornerRadius = Styles.Sizes.avatarCornerRadius
 
         let image = UIImage(named: "chevron-down").withRenderingMode(.alwaysTemplate)
-        let optionButtonWidth = (image.size.width ?? 0) + (2 * Styles.Sizes.gutter)
+        let optionButtonWidth = image.size.width + (2 * Styles.Sizes.gutter)
```
Huddie: This isn't relevant to this PR

Classes/Issues/Request/IssueRequestModel.swift
```diff
@@ -42,6 +42,63 @@ final class IssueRequestModel: ListDiffable {
...
+                                                             date: date,
+                                                             event: event)
+        } else {
+            styledString = IssueRequestModel.buildString(actor: actor,
+                                                         user: user,
+                                                         date: date,
+                                                         event: event)
+        }
+
+        self.string = StyledTextRenderer(
```
Huddie: nit: You don't need `self.` here right?

LucianoPAlmeida: Don't need, but I think is better for consistency since all other members being initialized require `self.` So doing all members even this with it :))

Huddie: They need `self.` because the parameter names match the variable names

LucianoPAlmeida: Yeah, but what I mean by consistency is
```swift
        self.id = id
        self.actor = actor
        self.user = user
        self.date = date
        self.event = event
        string = ... // Also a member 
        // Even it not needing, to maintain consistency in all memeber initialization 
       self.string = ... 

```
But no problem changing if it is a problem :)

Issue #2376, opened 2018-11-01
[VISION] Auth, login & bookmark refactor


ijm8710:
- [ ] Selective logout #2302
- [ ] Fix login scopes #2165
- [ ] Add PAT to auth #2165
- [ ] Bookmarks
  - [ ] iCloud sync #2377
  - [ ] Sort/reordering/foldering #1886
  - [ ] States to update open/closed icon and title edits #2379
  - [ ] Reflect total count in section header #2441
  - [ ] Integrate with Github’s “save for later” (pending API) #2377
-   [ ] return of search

The following referenced tickets can be closed:
#2379 #2377 #1886 #2441 #2302 #2165

ijm8710:
@rnystrom I would actually perhaps rate this as higher than 2375 because of feasible, and if my analysis on credential ticket is also accurate, this provides a solution without your data loss concern

rnystrom:
Oh interesting. I bet we could bundle all of it in the next update:

- Bookmark sync
- Selective logout
- Fix login scopes
- Add PAT to auth

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ijm8710:
Yeah that would be beautiful!

ijm8710:
Hey @rnystrom hopefully a tag and revisit here is ok :)
Do you think this might be a good idea to try to action for 1.25 or 1.26?

rnystrom:
Ya I’d like to get the login enhancements in a future version once the refactor noise quiets down. Can always use help.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ijm8710:
@rnystrom I laid everything out more condensed and on one ticket. Hopefully you agree this makes it easier to follow

rob-nash:
I found [this repository](https://github.com/intellum/neeman) today by @intellum. It might be useful to you.

Their recent work on this repo has broken two example projects nested in the root dir. So I submitted a PR to get the 'intermediate' example project running. 

Checkout the code [here](https://github.com/intellum/neeman/pull/20) and navigate to Root/Examples/Intermediate and run `carthage bootstrap` there.

![untitled 2018-11-28 16_46_31](https://user-images.githubusercontent.com/14126999/49167454-4a465000-f32d-11e8-9bc1-b79f44617360.gif)

ijm8710:
@rob-nash is that a wrapper of the mobile site or an app?

rob-nash:
[...]

Issue #2165, opened 2018-09-14
Credentials Issue


Huddie:
**Describe the bug**
I'm trying to fix my PR #2157.

My example gif shows a "something went wrong" error. At first, I assumed it was some maybe network issue or something since the error is the generic one (so who really knows). I just assumed it didn't code I changed since I didn't really change anything that I can see would ever effect it. 

I printed the result in `addCommentClient.swift` under the `case: .faliure` 

the result is: 

> failure(Optional(Although you appear to have the correct authorization credentials,
>the `GitHawkApp` organization has enabled OAuth App access restrictions, meaning that data
>access to third-parties is limited. For more information on these restrictions, including
>how to whitelist this app, visit
>https://help.github.com/articles/restricting-access-to-your-organization-s-data/
>))

I guessed it may have been my Github Client ID or Secret but 

1. The app shouldn't sign me in if those aren't valid (pretty sure)
2. I added them again to make sure it was right and I'm pretty sure they are fine.

Note: Using the Testflight version of githawk works fine.

**To Reproduce**
Umm.. unsure.

**Expected behavior**
For the comment to post successfully 

**Screenshots**
See PR with gif

Anyone have an idea of stuff I can try? Did I just forget to do something obvious? Does my code in the PR seem to mess something up in this regard? (I really don't think its the code cause it works for my other repos
 
Any help is appreciated

rnystrom:
So is this a bug with the latest App Store version? Do we need to disable OAuth restrictions?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Huddie:
@rnystrom no this post is from the App Store version.
Not sure what is causing the error on my specific branch

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jsg2021:
i’m also getting “something went wrong” on various projects.

jsg2021:
I don’t have any. Just what the public has.

rnystrom:
I thought I saw some weird inconsistencies w/ auth between PAT and normal login b/c of third-party scopes. Is that what this is from?

Huddie:
I logged in normally. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ijm8710:
@rnystrom since we closed my issue ticket today, do you have any idea what may cause it.

Also expanding on op that I’ve received this error when:
- Posting comments
- Opening/closing tickets

[...]

PR #2566, opened 2018-12-29
Enable adding a PAT from settings

rnystrom:
Part of #2376, #2165

simulator screen shot - iphone xs - 2018-12-29 at 14 56 29


ijm8710:
I also think there should be guidance on what scopes you recommend for user ie public repo scope access. I know this is just a partial merge but just to keep in mind :)

PR #2572, opened 2018-12-30
Retain bookmark order when syncing

### LINKED ISSUE #2537: Stray oddities from bookmark refactor/iCloud integration

rnystrom:
Fixes #2537

### DIFF
--- a/Classes/Bookmark/GitHubClient+Bookmarks.swift
+++ b/Classes/Bookmark/GitHubClient+Bookmarks.swift
@@ -16,9 +16,10 @@ private extension String {
     }
 }
 
-private func bookmarkGraphQLIDs(from json: [String: Any]) -> [String] {
-    return json.compactMap {
-        guard let item = $0.value as? [String: Any] else { return nil }
+private func bookmarkGraphQLIDs(from json: [String: Any], originalKeys: [String]) -> [String] {
+    // fetch by originalKeys to retain original ordering
+    return originalKeys.compactMap {
+        guard let item = json[$0] as? [String: Any] else { return nil }
         if let issueOrPullRequest = item["issueOrPullRequest"] as? [String: Any] {
             return issueOrPullRequest["id"] as? String
         }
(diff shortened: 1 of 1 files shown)


ijm8710:
Thank you Ryan! You’re the best!

ijm8710:
@rnystrom did this hit the build you just pushed? Must we do AppStore overwrite then Tf, since the order didn’t sync when I just updated a second ago

rnystrom:
Yup need to update from an old version again

ijm8710:
@rnystrom the orders not syncing over from the App Store version (just did a double-install on top)

ijm8710:
Hey Ryan just hoping you can give this another look when you get a chance before 1.26 public is pushed and we lose ability to revert

rnystrom:
To test this:

- On a TF version, un-bookmark everything so the iCloud store is empty
- Install App Store version
- Bookmark a few things (take a screenshot to remember order)
- Install TF version
- Check that bookmarks exist and are in order

ijm8710:
Thanks Ryan; looks like it does sync if you unbookmark everything first, I had just kept what was there thinking it would merge/overwrite.

The bookmark refactor does place newest bookmarked items at bottom though so my newest item pre-refactor and next item added since are added at opposite ends of list.

I also noticed search is gone as well.

Not sure if those are intentional design changes, unintentional, or temporary as you expand upon the refactor now. 

Ideally I’d still love just a simple count to reflect how many bookmarks you have in the header bar, ability to reorder the bookmarks and some other items I had aggregated in #2376 but are now stray tickets.

ijm8710:
@rnystrom did you have a chance to look at the above comment. One of the things I mentioned:

Old behavior : new bookmarks populate at top
Current behavior: new bookmarks populate at bottom 

The old behavior makes more sense. But if this was an intentional change, it creates a non continuous feed of bookmarks whereby you have relatively new bookmarks at the top (the newest pre bookmark refactor), but the absolute newest bookmarks at the bottom

PR #2616, opened 2019-01-20
Support rocket & eyes emojis

### LINKED ISSUE #2605: Support new reactions
GitHub [just released](https://github.blog/changelog/2019-01-18-more-reaction-emojis/) support for the 🚀 (“rocket”) and 👀 (“eyes”) reactions. It would be nice to be able to support these in GitHawk.

(the web UI now splits the reactions into two rows of four emoji on both desktop and mobile)

viktorasl:
Closes #2605

Added two new emojis to GQL schema, emoji picker & default emoji selection screens. Moreover, split emoji picker cells into two rows, though the order of them is not preserved from Github Web UI.

|simulator screen shot - iphone xr - 2019-01-20 at 11 13 14|
simulator screen shot - iphone xr - 2019-01-20 at 11 19 40|
|-|-|

1. Unknown emojis are still defaulted to 👍 , not which approach is better and if it's worth solving that at all;
2. Static UITableView for default emoji selection is a bit repetitive now, I suggest making dynamic list either as part of this PR or as a separate. Any thoughts?

### DIFF
--- a/Classes/Issues/Comments/Reactions/ReactionsMenuViewController.swift
+++ b/Classes/Issues/Comments/Reactions/ReactionsMenuViewController.swift
@@ -42,13 +42,13 @@ final class ReactionsMenuViewController: UICollectionViewController,
     private let sectionedReactions: [[ReactionContent]] = [
         [
             .thumbsUp,
-            .hooray,
             .thumbsDown,
-            .heart
+            .laugh,
+            .hooray
         ],
         [
-            .laugh,
             .confused,
+            .heart,
             .rocket,
             .eyes
         ]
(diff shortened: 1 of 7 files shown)


j-f1:
I agree we should default unknown ones to something else. Some ideas: ⭕️❓⁉️

Also, let’s take this opportunity to match the order to the website.

Sherlouk:
Great work picking this up! Well done on the first contribution ❤️

Classes/Issues/Comments/Reactions/ReactionsMenuViewController.swift
```diff
@@ -69,22 +76,27 @@ final class ReactionsMenuViewController: UICollectionViewController,
         collectionView?.register(EmojiCell.self, forCellWithReuseIdentifier: reuseIdentifier)
         collectionView?.reloadData()
         collectionView?.layoutIfNeeded()
+        let reactionsInRow = sectionedReactions.first?.count ?? 0
```
BasThomas: Can we at least throw an `assertionFailure` if `first` here is `nil`? It wouldn't make sense for this to have a `0` width in that case...

viktorasl: Fixed 👍

inkoperative:
wongie in the houuuuuse 🐆😬🤫

Issue #2587, opened 2019-01-04
Can not see all branches - too many branches


MohamedElPrince:
I have a repo with too many branches, when switching to a specific branch I cannot find it listed.
And there’s no search or pagination or even manual branch name entry

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rnystrom:
Ah the branches view doesn’t support pagination. This would be a great feature to add.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #2620, opened 2019-01-21
Pagination for repo branches list

viktorasl:
Closes (#2587)

Possibly quite an edge case to have more than 50 branches but it was fun working w/ GQL and pagination implementation here 🎆 

Follows pagination convention from `RepositoryIssuesViewController.swift` & `RepositoryClient.swift`, though I didn't understand why `RepoSearchPagesQuery` conforms to `RepositoryQuery` protocol instead of having extension for `RepoSearchPagesQuery.Data`.

|simulator screen shot - iphone xr - 2019-01-21 at 14 33 59|
|-|

### DIFF
--- a/Classes/Repository/RepositoryBranches/RepositoryBranchesViewController.swift
+++ b/Classes/Repository/RepositoryBranches/RepositoryBranchesViewController.swift
@@ -69,22 +69,24 @@ RepositoryBranchSectionControllerDelegate {
     }
 
     override func fetch(page: String?) {
-        client.fetchRepositoryBranches(owner: owner,
-                                 repo: repo
-            ) {  [weak self] result in
+        client.fetchRepositoryBranches(
+            owner: owner,
+            repo: repo,
+            nextPage: page as String?
+        ) {  [weak self] result in
             guard let strongSelf = self else { return }
             switch result {
-            case .success(let branches):
+            case .success(let payload):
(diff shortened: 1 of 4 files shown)


Classes/Repository/RepositoryBranches/RepositoryBranchesViewController.swift
```diff
@@ -69,22 +69,24 @@ RepositoryBranchSectionControllerDelegate {
...
+        client.fetchRepositoryBranches(
+            owner: owner,
+            repo: repo,
+            nextPage: page as String?
+        ) {  [weak self] result in
             guard let strongSelf = self else { return }
             switch result {
-            case .success(let branches):
+            case .success(let payload):
                 self?.branches = RepositoryBranchesViewController.arrangeBranches(
```
BasThomas: Any reason not to use `strongSelf` here and in other places where `self?` is used?

viktorasl: Didn’t notice the discrepancy in that code block. No reason at all, I can change and use `strongSelf`.

BasThomas: Yeah that'd be nice. Then I will merge after :)

viktorasl:
Sorry it took so long. Fixed 🔨

PR #2622, opened 2019-01-22
Add handoff support for code blob views

viktorasl:
I'm using handoff quite extensively (it's a great feature for me). Thus I did code blob view handoff just because sometimes it's easier to read code in a bigger screen.

There were no feature requests and from #331 I see that it's not expected that people will use handoff much. So if it looks like not a useful feature feel free to close it.

### DIFF
--- a/Classes/Utility/GithubURL.swift
+++ b/Classes/Utility/GithubURL.swift
@@ -1,0 +1,21 @@
+//
+//  GithubURL.swift
+//  Freetime
+//
+//  Created by Viktoras Laukevicius on 20/02/2019.
+//  Copyright © 2019 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+enum GithubURL {
+    static func codeBlob(repo: RepositoryDetails, branch: String, path: FilePath) -> URL? {
+        let builder = URLBuilder.github()
+            .add(path: repo.owner)
+            .add(path: repo.name)
+            .add(path: "blob")
+            .add(path: branch)
+        path.components.forEach { builder.add(path: $0) }
+        return builder.url
+    }
+}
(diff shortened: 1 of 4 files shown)


Huddie:
@viktorasl Just to clarify, if your using Githawk you can handoff and open the issue/pr/whatever on github.com on your mac?

viktorasl:
Yes, but probably the handoff can be done only to Safari. Issue/PR handoff was implemented with #331, this PR allows to open code blob view.

j-f1:
Yeah, Handoff, as it works now, is great! Would like to see it everywhere though.

viktorasl:
First of all, URL building for code blobs was already there and I didn't have to introduce `externalURL`.
Nontheless I've introduced `GithubURL` which should be used for concrete URL buildings. The idea is to move all Github URL building there & cover w/ tests.

Classes/Utility/GithubURL.swift
```diff
@@ -0,0 +1,21 @@
...
+//  GithubURL.swift
+//  Freetime
+//
+//  Created by Viktoras Laukevicius on 20/02/2019.
+//  Copyright © 2019 Ryan Nystrom. All rights reserved.
+//
+
+import Foundation
+
+enum GithubURL {
```
BasThomas: Oh cool, I like this idea. :)

BasThomas:
> Nontheless I've introduced `GithubURL` which should be used for concrete URL buildings. The idea is to move all Github URL building there & cover w/ tests.

Want to open an issue for that?

viktorasl:
> > Nontheless I've introduced `GithubURL` which should be used for concrete URL buildings. The idea is to move all Github URL building there & cover w/ tests.
> 
> Want to open an issue for that?

Absolutely!

viktorasl:
Raised an issue #2647;
Merged `master` & solved conflicts.

Issue #1184, opened 2017-12-04
Add fork count to repo


Sherlouk:
https://developer.github.com/v4/changelog/2017-12-02-schema-changes/

New to the GQL response! 😄 

Similar to #876 (Star count on repo)

PR #2624, opened 2019-01-23
Repo overview header including watchers, stars & forks

### LINKED ISSUE #876: View star count on repos
Put this as a cell above the readme?

viktorasl:
Closes #876 and #1184 + adds watchers count.
Picked quite simplified styling, hopefully it's good enough.
Not sure if implemented `StyledTextBuilder` correctly.

simulator screen shot - iphone xr - 2019-01-23 at 22 32 52

### DIFF
--- a/Classes/Repository/RepositoryOverviewHeaderViewModels.swift
+++ b/Classes/Repository/RepositoryOverviewHeaderViewModels.swift
@@ -1,0 +1,62 @@
+//
+//  RepositoryOverviewHeaderViewModels.swift
+//  Freetime
+//
+//  Created by Viktoras Laukevicius on 23/01/2019.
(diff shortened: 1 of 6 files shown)


ijm8710:
I honestly think what would make the most sense is to just create an isolated icon similar to the bookmark one top right. Honestly a star/watch should probably just replace the bookmark functionality as the combination of either of them basically are just a better implementation of a bookmarked repo anyway.

viktorasl:
Yes, but that will require to think more about:
1. Bookmark icon should be changed to?
2. Bookmarks tab should contain watched & starred repos probably. I guess it could have tabbed UI like repo overview.

I would be more than happy to implement that.

PR #2625, opened 2019-02-03
Adds sound option so device can vibrate and ring

flovilmart:
I was wondering why I never have any sound / vibration on notifications, while this is pleasant, we can let the user configure it in the notifications settings.


BasThomas:
Neat!

flovilmart:
Thanks @BasThomas I have commit access as well but wanted to make sure it was all good.

PR #2645, opened 2019-02-18
Migrate GitHawk to Swift 4.2

BasThomas:


Classes/Issues/Comments/Reactions/IssueReactionCell.swift
```diff
@@ -35,7 +35,7 @@ final class IssueReactionCell: UICollectionViewCell {
 
     override init(frame: CGRect) {
         super.init(frame: frame)
-        accessibilityTraits |= UIAccessibilityTraitButton
+        accessibilityTraits.insert(.button)
```
BasThomas: None of these were fixed by the migrator 😭

agisilaos: @BasThomas Highly recommend opening an issue about that. When I had issues with the migrator I opened an issue and they fixed it with the next release of Swift.

BasThomas: Thanks, that's a good idea! Found https://bugs.swift.org/browse/SR-8181.

BasThomas:
BitRise 404'd, can't see why the build is failing... also do not see any configuration file that we can change. @rnystrom?

Sherlouk:
@BasThomas Still think it's worth landing this, or should we just straight to Swift 5?

BasThomas:
Hm so this is going to be a tricky one I think, as we rely on a [`IGListKit` branch](https://github.com/GitHawkApp/GitHawk/blob/5bd1721d9a3df7737bfff856d8e450cacbc85404/Podfile#L29) that needs an update, presumably.

BasThomas:
Ha, nice timing @Sherlouk.

Let's see. It shouldn't add too much overhead to update to both, I think. But we can also go straight to 5.0.

PR #2650, opened 2019-02-22
Update CocoaPods and pods

BasThomas:


Classes/Issues/IssuesViewController.swift
```diff
@@ -572,7 +572,7 @@ IssueManagingContextControllerDelegate {
         case .item(let item):
             guard item is IssueResult else { break }
             updateAndScrollIfNeeded()
-        case .list: break
+        case .list, .clear: break
```
BasThomas: This is a new case in `FlatCache.Update`.

Classes/Systems/LocalNotificationsCache.swift
```diff
@@ -19,7 +19,7 @@ final class LocalNotificationsCache {
 
     private let path: String
     private lazy var queue: FMDatabaseQueue = {
-        return FMDatabaseQueue(path: self.path)
+        return FMDatabaseQueue(path: self.path)!
```
BasThomas: This now returns an optional; force unwrapping for now, as it seemingly worked in the past without error. :)

Local Pods/GitHubAPI/GitHubAPI.podspec
```diff
@@ -9,6 +9,6 @@ Pod::Spec.new do |spec|
   spec.source_files = 'GitHubAPI/*.swift'
   spec.ios.deployment_target     = '11.0'
   spec.watchos.deployment_target = '3.0'
-  spec.dependency 'Alamofire', '~> 4.4.0'
+  spec.dependency 'Alamofire', '~> 4.8'
```
BasThomas: I was a bit confused that we need to update this in the `podspec` _and_ the [Podfile](https://github.com/GitHawkApp/GitHawk/pull/2650/files#diff-3083edf0ef607bdec8f7c6a7c2ee663eR8) 😅

Feels like one of the two might not be necessary?

Classes/Issues/IssuesViewController.swift
BasThomas: I'm assuming this is a "correct" implementation here @rizwankce? :)

Issue #2649, opened 2019-02-22
"Mentioned" issues are not sorted by date


BasThomas:
Referencing #2612

**Describe the bug**
Take a look at only mentioned issues, and observe they are not sorted by date

**To Reproduce**
1. Go to "Inbox"
2. Click on Inbox
3. Click on Mentioned

**Expected behavior**
A list of issues that is sorted by date, latest on top

**Screenshots**
![screenshot](https://user-images.githubusercontent.com/4190298/53221907-d4bbf880-366b-11e9-8522-4caf673f2f6b.jpeg)

Request:

```
curl -v -X GET 'https://api.github.com/issues?access_token=123filter=mentioned&page=1&per_page=50' -H 'User-Agent: Freetime/1.27.1 (com.whoisryannystrom.freetime; build:1546622353; iOS 12.1.4) Alamofire/4.4.0' -H 'Accept-Language: en-DE;q=1.0, de-DE;q=0.9, nl-DE;q=0.8' -H 'Authorization: token 7e713728cee1b78872817efad46a940ce65af232' -H 'Accept: application/vnd.github.merge-info-preview+json' -H 'Accept-Encoding: gzip;q=1.0, compress;q=0.5' -H 'Cookie: logged_in=no;' 
```

akarataev:
I'll take a look at this, if nobody is against.

ijm8710:
@BasThomas I think this is slightly different then the ticket I opened that you referenced. My older ticket focused on the aggregation of all the issues within a specific filter type, aka created or mentioned.

Your ticket revolves around time stamps. Assuming we are identifying the same thing, what I noticed is that GitHub issue tickets will not fall in chronological order specifically when you yourself are the only one post a new comment. Unless someone else comments after, the timestamp will update to newer but since there is no new activity for you to review, it doesn’t bump it to the top. I actually wish it sorted chronologically regardless but it’s not an app error but just how GitHub handles things. I have even emailed them directly about this requesting that they added an option to sort chronologically for all tickets, even if you are the only updater. 

If we are on the same page here and you agree, I would love you to send GitHub another ticket on this to add visibility to what I wrote and hopefully influence them more to accommodate. If it’s something different, apologies and carry on.

BasThomas:
Interesting, @ijm8710. It feels like it would be behaving as expected then, as of now? It’s a bit confusing though. I’m not sure what the UX would look like, but GitHawk could add it as an option to sort by “last updated”? What do you think?

[...]

PR #2663, opened 2019-03-05
 Fix for "Mentioned issues are not sorted by date"

akarataev:
Issue mentioned in #2649   
I've implement sorting `Array<InboxDashboardModel>` by date for mentioned filter type.


Classes/Notifications/NotificationModelController.swift
```diff
@@ -24,6 +24,14 @@ extension NotificationViewModel {
     }
 }
 
+// sorting Array<InboxDashboardModel> by date for mentioned filter type
+extension Array where Element == InboxDashboardModel {
+    func sortedMentionedByDate(_ predicate: Bool = true, filter: V3IssuesRequest.FilterType) -> [Element] {
```
BasThomas: Hm, what's the reason for the `predicate`? When would you want to call this with `false`? That would just no-op, if I understand this correctly 😄

Classes/Notifications/NotificationModelController.swift
```diff
@@ -24,6 +24,14 @@ extension NotificationViewModel {
     }
 }
 
+// sorting Array<InboxDashboardModel> by date for mentioned filter type
+extension Array where Element == InboxDashboardModel {
```
BasThomas: It would also be 💯 if we can add a test for this one!

BasThomas:
Thanks for picking this up @akarataev!

Classes/Notifications/NotificationModelController.swift
akarataev: I thought it was a good place for [feature toggle](https://en.wikipedia.org/wiki/Feature_toggle). I removed this argument.

Classes/Notifications/NotificationModelController.swift
akarataev: I added tests for this functionality.

akarataev:
Hello @BasThomas, 
 
I made a changes, please see it.

FreetimeTests/SortingMentionedByDayTests.swift
```diff
@@ -0,0 +1,54 @@
...
+        let assigned = V3IssuesRequest.FilterType.assigned
+        // MARK: - when
+        let mentionedSorted = parsed.sortedMentionedByDate(filter: mentioned)
+        let createdSorted = parsed.sortedMentionedByDate(filter: created)
+        let assignedSorted = parsed.sortedMentionedByDate(filter: assigned)
+        // MARK: - then
+        XCTAssert(zip(mentionedSorted,
+            mentionedSorted.sorted(by: {$0.date > $1.date}))
+            .map { $0.date == $1.date }
+            .reduce(true) { $0 && $1 }, "Mentioned is not sorted by day")
```
BasThomas: You can use [SE-0207](https://github.com/apple/swift-evolution/blob/master/proposals/0207-containsOnly.md)'s [`allSatisfy`](https://developer.apple.com/documentation/swift/array/2994715-allsatisfy) here ;)

FreetimeTests/SortingMentionedByDayTests.swift
```diff
@@ -0,0 +1,54 @@
...
+class InboxDashboardModelBuilder {
+    let model: InboxDashboardModel
+    private init(date: Date) {
+        self.model = InboxDashboardModel(owner: "", name: "", number: 0, date: date,
+            text: StyledTextRenderer(string: StyledTextString(styledTexts: []),
+            contentSizeCategory: UIContentSizeCategory(rawValue: "")),
+            isPullRequest: false, state: .open
+        )
+    }
+    static func provide(number: Int, with interval: TimeInterval = TimeInterval(86400)) -> [InboxDashboardModel] {
```
BasThomas: Neat :)

FreetimeTests/SortingMentionedByDayTests.swift
akarataev: Wow 😲 
I did not know about it. Fixed it.

akarataev:
@BasThomas 

I can help design UI for this option. Do you have any suggestions?

BasThomas:
What do you think about a filter (a là Mail.app on iOS) in the top right? Would only be there for `mentioned` then. Don't think we should show it but leave it disabled in other views.

akarataev:
@BasThomas 

I can prepare several design options. How do you look at closing this PR under feature flag? And the development of the interface to make a separate task.

FreetimeTests/SortingMentionedByDayTests.swift
akarataev: Hm.. I thought it was a linter configuration. 

Снимок экрана 2019-03-10 в 19 49 38

BasThomas:
Hm, I don't think we've used feature flags for that before. Not opposed to it if we keep it documented in a document as well.

Huddie:
I don’t think giving sorted/unsorted options removes the issue Ryan was worried about since if a user selected the Sorted option we would be sorting on the client and possibly causing some out of sync issue with GitHub. Not sure why sorting would cause an issue but Ryan thought it might. Maybe wait for Ryan before we officially merge ?

akarataev:
@Huddie,
Ryan said what exactly should be out of sync?

ijm8710:
I agree with @huddie, [this was Ryan’s comment on it](https://github.com/GitHawkApp/GitHawk/issues/2574#issuecomment-451292917). I believe i was first to bring this all up, so trust me I’m all in favor of the sort in itself but sounds like Ryan wanted to keep things as dumbed down as possible so hopefully we can touch base with him on this before merging when he returns.

BasThomas:
OK, let's leave this one open for now then 👍

PR #2667, opened 2019-03-08
Fix for “Disallow Empty Bug Reports” Issue # 2615

wayni208:
Added a guard to the OnSend function in NewIssueTableViewController.swift that verifies the bodyField.text is not nil, only contains whitespace, or contains no text at all before submitting issue.

Called Squawk to display an error message in the bodyField view container as it had been displaying behind the keyboard.

Lastly, changed the Squawk being used when an empty titleText.text field is detected to also display in the bodyField container as it would also be hidden by the keyboard.

### DIFF
--- a/Classes/New Issue/NewIssueTableViewController.swift
+++ b/Classes/New Issue/NewIssueTableViewController.swift
@@ -145,11 +145,6 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
             return
         }
 
-        guard let text = bodyField.text, !text.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines).isEmpty else {
-            Squawk.showError(message: NSLocalizedString("You must provide an issue description!", comment: "Invalid body field when sending new issue"))
-            return
-        }
-
         titleField.resignFirstResponder()
         bodyField.resignFirstResponder()
         setRightBarItemSpinning()
--- a/Classes/New Issue/NewIssueTableViewController.swift
+++ b/Classes/New Issue/NewIssueTableViewController.swift
@@ -141,10 +141,17 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
     /// Attempts to sends the current forms information to GitHub, on success will redirect the user to the new issue
     @objc func onSend() {
         guard let titleText = titleText else {
-            Squawk.showError(message: NSLocalizedString("You must provide a title!", comment: "Invalid title when sending new issue"))
+            Squawk.shared.show(in: bodyField, config: Squawk.Configuration(text: NSLocalizedString("You must provide a title!", comment: "Invalid title when sending new issue")))
             return
         }
+        
+        guard let text = bodyField.text,!text.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines).isEmpty else {
+            Squawk.shared.show(in: bodyField, config: Squawk.Configuration(text: NSLocalizedString("You must provide a issue description!", comment: "Invalid description when sending new issue")))
 
+            return
+        
+        }
+        
         titleField.resignFirstResponder()
         bodyField.resignFirstResponder()
         setRightBarItemSpinning()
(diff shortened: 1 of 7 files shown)


wayni208:
Hey fellas. I'm Wayne. 

This is my first contribution. 

I'm just figuring out how GitHub acts with Xcode and all of that. I thought that I could store my environment variables in my fork and then pick only NewIssueTVC two use for a pull request but can't. So I reverted the original changes and recommitted. So sorry if this is a mess with the three commit business. If you would like, I can reinstall everything and redo it all. I'm not sure if that is all necessary or not. Just let me know.

BasThomas:
Hey, that’s awesome! Don’t worry about the commits 😊

wayni208:
Thanks!

It looked to me that the guard statement I OnSend for the titleField.text contents are unnecessary. Function titleFieldEditingChanged enables the Submit button only when titleField.text has content. Do we really need that guard statement or is iit accomplishing more than what I understand?

Sherlouk:
I think we want to approach this in a two fold system.

1. We should follow a similar strategy to the titleText in which we only enable the `Submit` button if both the title and description fields contain text.

We will probably want to introduce a new function `updateSubmitButtonState()` which is something a little like this:

```
func updateSubmitButtonState() {
    if titleText == nil {
        rightBarButtonItem.isEnabled = false
        return
    }

    if bodyText == nil {
        rightBarButtonItem.isEnabled = false
        return
    }

    rightBarButtonItem.isEnabled = true
}
```

(Not perfect, just written in GitHub)

We would then need to call that function from a few different places, such as when the title text field is changed, or the body text area is changed.

2. We should add a guard to the `onSend` which ensures bodyField has text as a final defence. You've already got this one done, but we can make an improvement here by taking advantage of `bodyText` which already trims and checks for an empty contents - we just need to make a guard similar to the one above!

```
        guard let bodyText = bodyText else {
            Squawk.shared.show(in: bodyField, config: Squawk.Configuration(text: NSLocalizedString("You must provide an issue description!", comment: "Invalid description when sending new issue")))
            return
        }
```

You can also improve the code a little bit further down in that same function where we go from `body: (bodyText ?? "") + signature)` to `body: bodyText + signature`

wayni208:
Thanks for the direction. I’ll get back into it and revise. 

BasThomas:
Let us know if there's something we can help with, @wayni208!

wayni208:
>Let us know if there's something we can help with, @wayni208!

@BasThomas 

As a matter of fact...

How should I set up git to exclude or ignore project, entitlement, and plist files after setting up my group and bundle info? Should i place all of these (there are six) into the global Xcode .gitignore?

BasThomas:
That's a good question. Are you referring to GitHawk? You shouldn't need to set that up for simulator builds at least.

I don't know at the top of my head which files it generates, but it might indeed be worth adding these to a `.gitignore` to avoid checking them in, at least in case these are files that are user-specific and wouldn't be "stable" in a git repository with multiple people working on it.

wayni208:
Good morning Bas! 

I'm sorry, I should have explained better. I'm referring to my local Xcode working copy. 

After running all of the setup instructions to be able to build and run correctly, I have these files modified:

Freetime.xcodeproj/project.pbxproj
FreetimeWatch/FreetimeWatch.entitlements
FreetimeWatch/Info.plist
FreetimeWatch Extension/FreetimeWatch Extension.entitlements
FreetimeWatch Extension/Info.plist
Resources/Freetime.entitlements

I can work around it.I just figured that maybe you guys had already worked around it on your local setup and could give me some direction. I've tried adding these to my exudes file but it's still picking them up.

BasThomas:
That's weird — for simulator builds, I've never seen the entitlements and plists being updated. That also means I've not worked around it. What are the changes (and what changes) the files?

wayni208:
These are just changes to group and bundle identifiers. I think it’s just the byproduct of following setup.md and Xcode. How are you doing it? 

All good now, though. When I created a working branch to checkout and make my changes on, it finally processed my exclude file and are hiding these files from me. 

Issue #2615, opened 2019-01-20
Disallow Empty Bug Reports


Sherlouk:
Noticed there's been a number of empty bug reports that end up just being closed down or things that don't really make sense. Probably not a huge issue as they just get closed down and move on with our day.

Having said that, is it worth adding a guard in the 'submit bug report' page to prevent empty bug reports? Or at least a confirmation?

Happy to close if no one really minds but thought I'd raise a quick one for it

wayni208:
Was this addressed? I’m unable to submit an empty report and looking at NewIssueTableViewController.swift it seems this is normal. How are you reproducing this?

Sherlouk:
Hadn't personally tried/experienced it, but historically we received a lot of empty reports. Though this does seem to have died down...

May no longer be an issue

wayni208:
I see what you are talking about, actually. I’ll take care of this.

PR #2676, opened 2019-03-10
Fix for "Disallow Empty Bug Reports" Issue #2615

wayni208:
Added new function in Squawk+GitHawk.swift, showIssueError, to handle error reporting through Squawk. The previous handling would display the error beneath the keyboard, on this form, and not be seen. Not that this should ever be seen.

Added a function in NewIssueTableViewController.swift that checks both the UITextField and the UITextView for content before enabling the Submit button. After the Submit button is enabled and used, we will check once again to verify content, throwing visible errors via Squawk in the TextView if no text is present, before sending.

### DIFF
--- a/Classes/New Issue/NewIssueTableViewController.swift
+++ b/Classes/New Issue/NewIssueTableViewController.swift
@@ -44,7 +44,7 @@ protocol NewIssueTableViewControllerDelegate: class {
     func didDismissAfterCreatingIssue(model: IssueDetailsModel)
 }
 
-final class NewIssueTableViewController: UITableViewController, UITextFieldDelegate {
+final class NewIssueTableViewController: UITableViewController, UITextFieldDelegate, UITextViewDelegate {
 
     weak var delegate: NewIssueTableViewControllerDelegate?
 
@@ -103,6 +103,9 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         titleField.delegate = self
         titleField.font = Styles.Text.body.preferredFont
 
+        // Setup the description textView to report if it has been edited
+        bodyField.delegate = self
+
         // Setup markdown input view
         bodyField.githawkConfigure(inset: false)
         setupInputView()
@@ -141,7 +144,12 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
     /// Attempts to sends the current forms information to GitHub, on success will redirect the user to the new issue
     @objc func onSend() {
         guard let titleText = titleText else {
-            Squawk.showError(message: NSLocalizedString("You must provide a title!", comment: "Invalid title when sending new issue"))
+            Squawk.showIssueError(message: NSLocalizedString("You must provide a title!", comment: "Invalid title when sending new issue"), view: bodyField)
+            return
+        }
+
+        guard let bodyText = bodyText else {
+            Squawk.showIssueError(message: NSLocalizedString("You must provide an issue description!", comment: "Invalid description when sending new issue"), view: bodyField)
             return
         }
 
@@ -155,7 +163,7 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
             owner: owner,
             repo: repo,
             title: titleText,
-            body: (bodyText ?? "") + signature)
+            body: bodyText + signature)
         ) { [weak self] result in
             guard let strongSelf = self else { return }
             strongSelf.setRightBarItemIdle()
@@ -205,6 +213,20 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         bodyField.inputAccessoryView = actions
     }
 
+    func updateSubmitButtonState() {
+        if titleText == nil {
+            navigationItem.rightBarButtonItem?.isEnabled = false
+            return
+        }
+
+        if bodyText == nil {
+            navigationItem.rightBarButtonItem?.isEnabled = false
+            return
+        }
+
+        navigationItem.rightBarButtonItem?.isEnabled = true
+    }
+    
     // MARK: UITextFieldDelegate
 
     /// Called when the user taps return on the title field, moves their cursor to the body
@@ -213,11 +235,18 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         return false
     }
 
+    // MARK: UITextViewDelegatre
+
+    /// Called when editing changed on the body field, enable/disable submit button based on title and body
+    func textViewDidChange(_ bodyField: UITextView) {
+        updateSubmitButtonState()
+    }
+
     // MARK: Actions
 
-    /// Called when editing changed on the title field, enable/disable submit button based on title text
+    /// Called when editing changed on the title field, enable/disable submit button based on title and body
     @IBAction func titleFieldEditingChanged(_ sender: Any) {
-        navigationItem.rightBarButtonItem?.isEnabled = titleText != nil
+        updateSubmitButtonState()
     }
 
 }
--- a/Classes/Systems/Squawk+GitHawk.swift
+++ b/Classes/Systems/Squawk+GitHawk.swift
@@ -83,4 +83,9 @@ extension Squawk {
         triggerHaptic()
     }
 
+    static func showIssueError(message: String, view: UIView?) {
+        Squawk.shared.show(in: view, config: errorConfig(text: message))
+        triggerHaptic()
+    }
+
 }


wayni208:
Hey I'm learning quite a bit about git lately. 😀

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -205,6 +213,20 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         bodyField.inputAccessoryView = actions
     }
 
+    func updateSubmitButtonState() {
+        if titleText == nil {
```
Huddie: Maybe combine these 2 ifs with an or ( `||`) or all three with a ternary. Something like

 navigationItem.rightBarButtonItem?.isEnabled = ( titleText == nil || bodyText == nil ) ? false : true

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -213,11 +235,18 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         return false
     }
 
+    // MARK: UITextViewDelegatre
```
Huddie: nit: Spelling - UITextViewDelegate

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -141,7 +144,12 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
     /// Attempts to sends the current forms information to GitHub, on success will redirect the user to the new issue
     @objc func onSend() {
         guard let titleText = titleText else {
-            Squawk.showError(message: NSLocalizedString("You must provide a title!", comment: "Invalid title when sending new issue"))
+            Squawk.showIssueError(message: NSLocalizedString("You must provide a title!", comment: "Invalid title when sending new issue"), view: bodyField)
```
Huddie: Thoughts on rewording this alert? Maybe something kinder like:

Title: "Title Required" or "Issue requires title"
Body: "Please provide a title and try again"

Classes/New Issue/NewIssueTableViewController.swift
BasThomas: Well the `false : true` is unnecessary, of course. Although you should then flip the result. 😊

Huddie: Hahahahah! Yes correct on both points. Totally forget the ternary

wayni208: Ok. I can change this around. How should I resubmit? I close and then do a new PR right?

Huddie: @wayni208 no need to close. Once you make a new commit to your branch it will update here

wayni208: There we go! Thanks for the feedback and direction on that function. Such a more elegant way to get it done.

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -205,6 +213,10 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         bodyField.inputAccessoryView = actions
     }
 
+    func updateSubmitButtonState() {
+        navigationItem.rightBarButtonItem?.isEnabled = ( titleText == nil || bodyText == nil ) ? false : true
```
Huddie: Like @BasThomas mentioned, you actually can just leave it as 

`navigationItem.rightBarButtonItem?.isEnabled = !(titleText == nil || bodyText == nil)

wayni208: I see what you did there. 😀

Classes/New Issue/NewIssueTableViewController.swift
BasThomas: Awesome! Was about to say you don't have to reopen every time. 💯

Classes/New Issue/NewIssueTableViewController.swift
BasThomas: Yep, I like title option 2... and I like the body. Mind updating that, @wayni208?

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -141,7 +144,12 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
     /// Attempts to sends the current forms information to GitHub, on success will redirect the user to the new issue
     @objc func onSend() {
         guard let titleText = titleText else {
-            Squawk.showError(message: NSLocalizedString("You must provide a title!", comment: "Invalid title when sending new issue"))
+            Squawk.showIssueError(message: NSLocalizedString("An issue title is required. Please add a title and try again.", comment: "Invalid title when sending new issue"), view: bodyField)
+            return
+        }
+
+        guard let bodyText = bodyText else {
+            Squawk.showIssueError(message: NSLocalizedString("Please provide as much of a detailed description possible and try again.", comment: "Invalid description when sending new issue"), view: bodyField)
```
BasThomas: Hm, I think this might be a bit confusing. Of course a good issue description is welcome, but not required. GitHub itself allows an empty body; I think we shouldn't fight that.

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -205,6 +213,10 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         bodyField.inputAccessoryView = actions
     }
 
+    func updateSubmitButtonState() {
+        navigationItem.rightBarButtonItem?.isEnabled = !( titleText == nil || bodyText == nil )
```
BasThomas: I think we should split this up a bit, with the goal of making it easier to read (and again, not requiring a body).

What do you think about the following?

```swift
let hasTitleText = titleText != nil
let isTitleTextEmpty = titleText?.isEmpty == true // This would be additional. We could even check for whitespace only in the future if we want to.
navigationItem.rightBarButtonItem?.isEnabled = hasTitleText && isTitleTextEmpty == false
```

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -213,11 +225,18 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         return false
     }
 
+    // MARK: UITextViewDelegate
+
+    /// Called when editing changed on the body field, enable/disable submit button based on title and body
```
BasThomas: Awesome you thought about the header docs! If we drop the body requirement though, this should be updated.

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -205,6 +213,10 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
         bodyField.inputAccessoryView = actions
     }
 
+    func updateSubmitButtonState() {
```
BasThomas: Would you like to write a test for this? :)

Classes/New Issue/NewIssueTableViewController.swift
wayni208: How about we just throw up an "Are you sure you want to submit this issue with no description?" confirmation before send? 

#2615

Classes/New Issue/NewIssueTableViewController.swift
wayni208: Our titleText variable, higher up in the code, checks inside the field for empty and whitespace. 

```suggestion
private var titleText: String? {
     guard let raw = titleField.text?.trimmingCharacters(in: .whitespacesAndNewlines) else { return nil }
     if raw.isEmpty { return nil }
     return raw
}
```

We could just go back to this original action code for the titleField.

```suggestion
@IBAction func titleFieldEditingChanged(_ sender: Any) {
      navigationItem.rightBarButtonItem?.isEnabled = titleText != nil
 }
```

wayni208: Higher up our titleText variable is checking for empty and whitespace.

```swift
private var titleText: String? {
        guard let raw = titleField.text?.trimmingCharacters(in: .whitespacesAndNewlines) else { return nil }
        if raw.isEmpty { return nil }
        return raw
}
```

I think the best option would be to revert to the original. If we are doing away with the UITextViewDelegate and not checking bodyField to turn on the Submit button we can get rid of that function completely. 

```swift
@IBAction func titleFieldEditingChanged(_ sender: Any) {
        navigationItem.rightBarButtonItem?.isEnabled = titleText != nil
}
```

Classes/New Issue/NewIssueTableViewController.swift
wayni208: We can do away with that function as I mentioned. But if we don't, I sure will!

Classes/New Issue/NewIssueTableViewController.swift
wayni208: Before I resolve this and because I had made the change before you replied @BasThomas , are we good on the message as I've written it? Here are both:

```swift
        guard let titleText = titleText else {
            Squawk.showIssueError(message: NSLocalizedString("An issue title is required. Please add a title and try again.", comment: "Invalid title when sending new issue"), view: bodyField)
            return
        }
```

Also
```swift
cancelAction_onCancel(
            texts: [titleText, bodyText],
            title: NSLocalizedString("Unsaved Changes", comment: "New Issue - Cancel w/ Unsaved Changes Title"),
            message: NSLocalizedString("Are you sure you want to discard this new issue? Your message will be lost.",
                                       comment: "New Issue - Cancel w/ Unsaved Changes Message")
        )
```

Classes/New Issue/NewIssueTableViewController.swift
wayni208: @BasThomas @Huddie 

I've made a few changes and reverted back to the original behavior where it would turn on the submit button once the titleField had changes. I've also added a confirmation dialogue notifying the user they have no description text and asking if they still want to submit or return to then form to add one. 

I had to do a funky deal to get the dialogue correct. I split the onSend function in two because of the @objc function call in the accessibility button function and I needed another function to call on if the user decided to go ahead with the submit. I'm pretty certain this could have been done an easier way so please take a look and give me some direction if there is a better way to implement this.

Classes/New Issue/NewIssueTableViewController.swift
wayni208: I'm concerned about the having split that onSend function in two and if I need to write a test for it. Also concerned about writing the test. I couldn't see anywhere where we are currently testing this portion of the code. I'm unfamiliar with tests in general so if you could give me some direction on this I would appreciate it.

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -140,8 +140,26 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
...
+
+        submitAlert.addAction(UIAlertAction(title: "Submit", style: .default, handler: { (action: UIAlertAction!) in
+            self.finishSend()
+        }))
+
+        submitAlert.addAction(UIAlertAction(title: "Cancel", style: .cancel, handler: { (action: UIAlertAction!) in
+            submitAlert .dismiss(animated: true, completion: nil)
+        }))
+
+        present(submitAlert, animated: true, completion: nil)
```
Huddie: If I recall correctly from @BasThomas, no need for the completion part here. You can leave it `animated: true)` and it will default completion to `nil`

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -140,8 +140,26 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
...
+            self.finishSend()
+        } else {
+        let submitAlert = UIAlertController(title: "Please Provide Description", message: "Are you certain you want to submit this issue without a description?", preferredStyle: UIAlertControllerStyle.alert)
+
+        submitAlert.addAction(UIAlertAction(title: "Submit", style: .default, handler: { (action: UIAlertAction!) in
+            self.finishSend()
+        }))
+
+        submitAlert.addAction(UIAlertAction(title: "Cancel", style: .cancel, handler: { (action: UIAlertAction!) in
+            submitAlert .dismiss(animated: true, completion: nil)
```
Huddie: See below about completion being unnecessary

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -140,8 +140,26 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
 
     /// Attempts to sends the current forms information to GitHub, on success will redirect the user to the new issue
     @objc func onSend() {
+        if let bodyText = bodyText {
+            self.finishSend()
+        } else {
+        let submitAlert = UIAlertController(title: "Please Provide Description", message: "Are you certain you want to submit this issue without a description?", preferredStyle: UIAlertControllerStyle.alert)
+
+        submitAlert.addAction(UIAlertAction(title: "Submit", style: .default, handler: { (action: UIAlertAction!) in
```
Huddie: I think you can just do `handler: { _ in ...
}`

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -140,8 +140,26 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
 
     /// Attempts to sends the current forms information to GitHub, on success will redirect the user to the new issue
     @objc func onSend() {
+        if let bodyText = bodyText {
+            self.finishSend()
+        } else {
+        let submitAlert = UIAlertController(title: "Please Provide Description", message: "Are you certain you want to submit this issue without a description?", preferredStyle: UIAlertControllerStyle.alert)
```
Huddie: nit: indent

Huddie: You can just do `.alert` it will infer the type

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -140,8 +140,26 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
 
     /// Attempts to sends the current forms information to GitHub, on success will redirect the user to the new issue
     @objc func onSend() {
+        if let bodyText = bodyText {
+            self.finishSend()
```
Huddie: Is `self` necessary here?

Classes/New Issue/NewIssueTableViewController.swift
wayni208: Instead of all the crazy ellipses? That would be cool.

Classes/New Issue/NewIssueTableViewController.swift
wayni208: No. Just in the closure. Fixed

Classes/New Issue/NewIssueTableViewController.swift
wayni208: Nope. I still count all those crazy ellipses. But I was able to clean everything up overall.

Classes/New Issue/NewIssueTableViewController.swift
BasThomas: Sounds good!

Classes/New Issue/NewIssueTableViewController.swift
BasThomas: To be honest I'm still in favor of allowing an empty description.

Classes/New Issue/NewIssueTableViewController.swift
BasThomas: Ah, I missed that. Go ahead.

Classes/New Issue/NewIssueTableViewController.swift
BasThomas: I'll get back to you on this one once I have a bit more time. Let's leave it for now; it's OK :)

Classes/New Issue/NewIssueTableViewController.swift
BasThomas: Correct; `completion` has a default expression of `nil` here. :)

Classes/New Issue/NewIssueTableViewController.swift
BasThomas: Can you check again? You should be able to change this:

```diff
- { (action: UIAlertAction!) in
+ { _ in
```

Classes/New Issue/NewIssueTableViewController.swift
```diff
@@ -140,8 +140,22 @@ final class NewIssueTableViewController: UITableViewController, UITextFieldDeleg
 
     /// Attempts to sends the current forms information to GitHub, on success will redirect the user to the new issue
     @objc func onSend() {
+        if let bodyText = bodyText {
+            finishSend()
+        } else {
+            let submitAlert = UIAlertController(title: "Please Provide Description", message: "Are you certain you want to submit this issue without a description?", preferredStyle: .alert)
+
+                submitAlert.addAction(UIAlertAction(title: "Submit", style: .default, handler: { _ in self.finishSend() }))
```
BasThomas: The indentation here is still a bit off.

Classes/New Issue/NewIssueTableViewController.swift
wayni208: @BasThomas as in not popping an alert before sending an empty description?

BasThomas: Exactly!

PR #2685, opened 2019-03-11
Emojis!

### LINKED ISSUE #1616: Support emoji in labels
GitHub [recently released](https://blog.github.com/2018-02-22-label-improvements-emoji-descriptions-and-more/) a feature allowing you to add emoji to label names. This allows you to do `bug 🐛` instead of the boring `bug`.

GItHawk doesn’t yet recognize emoji in labels and just renders it as a string. It looks like this:
a914c35c-6302-4604-9bd7-e23261202107

Huddie:
:bug: :smile:
Yay! Closes #1616

Simulator Screen Shot - iPhone XR - 2019-03-11 at 14 33 24

### DIFF
--- a/Classes/Models/RepositoryLabel.swift
+++ b/Classes/Models/RepositoryLabel.swift
@@ -12,11 +12,11 @@ import IGListKit
 final class RepositoryLabel: ListDiffable, Hashable, Equatable, ListSwiftDiffable {
 
     let color: String
-    let name: String
+    var name: String = ""
 
     init(color: String, name: String) {
         self.color = color
-        self.name = name
+        self.name = name.replacingGithubEmoji
     }
 
     // MARK: ListDiffable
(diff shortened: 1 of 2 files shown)


Huddie:
Note: We already have a method that converts `:bug:` to :bug: as well as a test. So I do not include a test in this PR

Sherlouk:
Well that is simpler than I expected 😂

Huddie:
@Sherlouk agreed! 

Classes/Models/RepositoryLabel.swift
```diff
@@ -12,11 +12,11 @@ import IGListKit
 final class RepositoryLabel: ListDiffable, Hashable, Equatable, ListSwiftDiffable {
 
     let color: String
-    let name: String
+    var name: String = ""
```
BasThomas: Any reason you changed this?

Huddie: If let you can’t mutate. Meaning I can’t switch `:bug:` no?

BasThomas: You're only setting it once in the initializer though, isn't it?

Huddie: Oh yes! I originally had a separate fix (I created my
Own method that did the mapping. And I needed `self` but that doesn’t apply anymore. I’ll PR a quick fix to this soon. Thank you for pointing that out

PR #2689, opened 2019-03-13
Fix notification bug

### LINKED ISSUE #2298: Push notification doesn’t mark as read
Anyone else?

Huddie:
Closes #2298

@BasThomas let me know if this makes sense. 

Issue:
Tapping notification to open issue doesn't mark the notification read

Solution:
Notification ID is passed apart of the request when creating local notification. Before routing to the Issue, extract the ID and mark the notification as read. 

We are marking notifications are read before actually displaying the notification but this no different than how it works when tapping on a note. in the inbox.  

Possible enhancement:
We have some redundant code here. The `markNotificationRead` method shows up here and in `NotificationModelController` suggestions are welcome as to where the best place to put it is. Keep in mind if we refactor both methods we will need to pass in the `GithubClient` which isn't great.

### DIFF
--- a/Classes/Systems/AppRouter/AppController.swift
+++ b/Classes/Systems/AppRouter/AppController.swift
@@ -33,6 +33,10 @@ RouterPropsSource {
         sessionManager.addListener(listener: self)
     }
 
+    func getAppClient() -> GithubClient? {
+        return self.appClient
+    }
+    
     func appDidFinishLaunching(with window: UIWindow?) {
         guard let controller = window?.rootViewController as? AppSplitViewController else {
             fatalError("App must be setup with a split view controller")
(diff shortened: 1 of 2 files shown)


Classes/Systems/AppRouter/AppController.swift
```diff
@@ -33,6 +33,10 @@ RouterPropsSource {
         sessionManager.addListener(listener: self)
     }
 
+    func getAppClient() -> GithubClient? {
```
BasThomas: This feels a bit weird to me... think it would be nice if we can keep the reading logic in one place.

Huddie: No problem so how should I get the app client in the extension?

Huddie: @BasThomas

BasThomas: Hm, that's a question I don't know I can answer :(

Huddie: I can make it public or create a getter, or I could move the extension into the main class

BasThomas: I feel like option 3 would be the best for now.

PR #2698, opened 2019-03-20
Preserve emojis order when adding

### LINKED ISSUE #2403: Adding reaction goes to end of list
So if a comment has a reaction of 🎉 and then I add 👍 githawk adds it to the end of the list. If you then refresh the issue it flips the order since 👍 is before 🎉 in githubs order.

Very minor but we should insert respecting the order of reactions!

<details>
<pre>
Version 1.23.0 (1538244553)
Device: iPhone11,2 (iOS 12.1)
TestFlight: true
</pre>
</details>

viktorasl:
* reactions order in app is defined by array in app (not received order from API)
* reactions picker & UI list share the same data structure

Closes #2403.

When inserting reaction to the middle last cell held UI state applied by `pullOut` and that's why that had to be extracted to a separate method which has to be called by `configure` method.

Not sure if private `Array` extension is correct [implementation](https://github.com/GitHawkApp/GitHawk/pull/2698/files#diff-37a418fe95a2aab45fe0f263b55a58c0R36) in this case.

### DIFF
--- a/Classes/Issues/Comments/Reactions/IssueCommentReactionViewModel.swift
+++ b/Classes/Issues/Comments/Reactions/IssueCommentReactionViewModel.swift
@@ -34,11 +34,10 @@ final class IssueCommentReactionViewModel: ListDiffable {
 
     init(models: [ReactionViewModel]) {
         let reactions = IssueCommentReactionViewModel.allReactions
-        let sortedModels = models.sorted { (m1, m2) -> Bool in
-            let m1Idx = reactions.firstIndex(of: m1.content) ?? 0
-            let m2Idx = reactions.firstIndex(of: m2.content) ?? 0
-            return m1Idx < m2Idx
+        let indexOfModel: (ReactionViewModel) -> Int = {
+            reactions.firstIndex(of: $0.content) ?? 0
         }
+        let sortedModels = models.sorted { indexOfModel($0) < indexOfModel($1) }
         self.models = sortedModels
 
         var map = [ReactionContent: ReactionViewModel]()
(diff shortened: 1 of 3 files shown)


Classes/Issues/Comments/Reactions/IssueCommentReactionViewModel.swift
```diff
@@ -17,16 +17,33 @@ extension ReactionContent: Hashable {
...
+    ]
+
     let models: [ReactionViewModel]
     private let map: [ReactionContent: ReactionViewModel]
     private let flatState: String
 
     init(models: [ReactionViewModel]) {
-        self.models = models
+        let reactions = IssueCommentReactionViewModel.allReactions
+        let sortedModels = models.sorted { (m1, m2) -> Bool in
```
jdisho: nit: 
instead of `m1`, `m2`, why not `lhs`, `rhs`?
additionally for `m1Idx`, `m2Idx`, why not `lhsContent` and `rhsContent`

viktorasl: `lhs` & `rhs` totally makes sense!
Not sure about `xContent` as it's an index of content in predefined array for order.

jdisho: @viktorasl what is the status of this? 😊

viktorasl: Oh sorry, I thought I've fixed this. On it.

viktorasl: So sorry it took so long. Fixed now.

YBHQ:
~~~~

PR #2713, opened 2019-04-01
Only mark new features as fetched after call goes through

BasThomas:
It seems like we only try to check for new features _once_ per version. That's cool, as to limit unnecessary network requests, but would silently not show any new features if that call would initially fail (eg. no network).

It now checks if it either succeeds (200) or can't find the features (404), and then communicate that it was fetched.

### DIFF
--- a/Classes/Notifications/NewFeaturesController.swift
+++ b/Classes/Notifications/NewFeaturesController.swift
@@ -64,17 +64,25 @@ final class NewFeaturesController {
         guard let url = fetchURL,
             (testing == true || hasFetchedLatest == false)
             else { return }
-        hasFetchedLatest = true
 
-        let task = session.dataTask(with: url) { (data, response, _) in
-            if let response = response as? HTTPURLResponse,
-                response.statusCode == 200,
-                let data = data,
-                let string = String(data: data, encoding: .utf8) {
-                DispatchQueue.main.async { [weak self] in
-                    self?.latestMarkdown = string
-                    success()
-                }
+        let task = session.dataTask(with: url) { data, response, error in
+            guard
+                let httpResponse = response as? HTTPURLResponse,
+                [200, 404].contains(httpResponse.statusCode),
+                let data = data else {
+                    let message = """
+                    "Expected a successfull or failing status code, but errored
+                    with \(error?.localizedDescription ?? "no error") for
+                    \(response?.description ?? "no response")
+                    """
+                    return assertionFailure(message)
+            }
+            let string = String(decoding: data, as: UTF8.self)
+            DispatchQueue.main.async { [weak self] in
+                guard let strongSelf = self else { return }
+                strongSelf.hasFetchedLatest = true
+                strongSelf.latestMarkdown = string
+                success()
             }
         }
         task.resume()


Classes/Notifications/NewFeaturesController.swift
```diff
@@ -64,17 +64,25 @@ final class NewFeaturesController {
...
+                    let message = """
+                    "Expected a successfull or failing status code, but errored
+                    with \(error?.localizedDescription ?? "no error") for
+                    \(response?.description ?? "no response")
+                    """
+                    return assertionFailure(message)
+            }
+            let string = String(decoding: data, as: UTF8.self)
+            DispatchQueue.main.async { [weak self] in
+                guard let strongSelf = self else { return }
```
jdisho: nit: I think there is no need to write `strongSelf`. `self` would work.

Classes/Notifications/NewFeaturesController.swift
```diff
@@ -64,17 +64,25 @@ final class NewFeaturesController {
...
-                }
+        let task = session.dataTask(with: url) { data, response, error in
+            guard
+                let httpResponse = response as? HTTPURLResponse,
+                [200, 404].contains(httpResponse.statusCode),
+                let data = data else {
+                    let message = """
+                    "Expected a successfull or failing status code, but errored
+                    with \(error?.localizedDescription ?? "no error") for
+                    \(response?.description ?? "no response")
```
jdisho: Did you mean `httpResponse` (instead of `response`)?

Classes/Notifications/NewFeaturesController.swift
```diff
@@ -64,17 +64,25 @@ final class NewFeaturesController {
...
+                    """
+                    return assertionFailure(message)
+            }
+            let string = String(decoding: data, as: UTF8.self)
+            DispatchQueue.main.async { [weak self] in
+                guard let strongSelf = self else { return }
+                strongSelf.hasFetchedLatest = true
+                strongSelf.latestMarkdown = string
```
jdisho: why not?

```swift 
strongSelf.latestMarkdown = String(decoding: data, as: UTF8.self)
```

Classes/Notifications/NewFeaturesController.swift
```diff
@@ -64,17 +64,25 @@ final class NewFeaturesController {
...
-                let data = data,
-                let string = String(data: data, encoding: .utf8) {
-                DispatchQueue.main.async { [weak self] in
-                    self?.latestMarkdown = string
-                    success()
-                }
+        let task = session.dataTask(with: url) { data, response, error in
+            guard
+                let httpResponse = response as? HTTPURLResponse,
+                [200, 404].contains(httpResponse.statusCode),
```
jdisho: I am not sure if we need to take into consideration `404`. If the status is `404` and rest of the guard passed, means that `hasFetchedLatest` will be `true`. Which, in IMO, shouldn't be. 🤔

jdisho: I think we need to care only about this range `200..<300`.

jdisho (commented):
Hey @BasThomas, thanks for the PR.
I have couple of suggestions/questions.

(Probably outside this PR scope, but wouldn't make sense for `fetch` to have an error handler)?

adilaid:
>It seems like we only try to check for new features _once_ per version. That's cool, as to limit unnecessary network requests, but would silently not show any new features if that call would initially fail (eg. no network). ...

@BasThomas 

PR #2733, opened 2019-04-25
Migrate to Xcode 10.2.x

viktorasl:
To fix compatibility issues Apollo dependency has to be upgraded to >=0.10.0. After that a bit more migration has to be done.


oz90210:
>To fix compatibility issues Apollo dependency has to be upgraded to >=0.10.0. After that a bit more migration has to be done.

@viktorasl

viktorasl:
One other thing: I would like to extract Apollo build step script to a shell script file. One line script in pbxproj is a bit annoying.

Classes/Issues/Issue+IssueType.swift
```diff
@@ -11,6 +11,18 @@ import IGListKit
 
 extension IssueOrPullRequestQuery.Data.Repository.IssueOrPullRequest.AsIssue: IssueType {
 
+    var id: String {
```
BasThomas: These are needed because of what? 🤔

BasThomas: also `isLocked` would be neat, but no blocker of course :)

BasThomas:
Thanks for keeping this up-to-date @viktorasl!

Classes/Issues/Issue+IssueType.swift
viktorasl: New Apollo version generated a bit different interface. I’ll take a look if it can generate the old one (I don’t have much experience w/ GraphQL and Apollo, but it’s a high time to get more)

viktorasl: According to https://www.apollographql.com/docs/ios/fragments:
>  fields included through fragments are not exposed directly, but require you to access the data through the fragment explicitly

probably it was possible to access fields directly for single field fragments:
```
fragment nodeFields on Node {
  id
}
```

viktorasl:
Btw I see no reason to not update to currently latest Apollo version, thus updated to the latest version.

Issue #2646, opened 2019-02-19
Allow individually deleting bookmarks


BasThomas:
See also #914; we should add haptics too.

rluftw:
I'm very new here and I'd love to work on this feature. I don't see anyone working on this in #914. Mind if I take this?

PR #2741, opened 2019-05-03
Enable individual bookmark deletes

rluftw:
Hi all! This is my first time contributing to this project. Huge fan of the app!

This PR enables a user to remove bookmarks individually instead of clearing them all out. It uses the same pattern to remove rows as `SearchRecentSectionController`.

Feature Request: #2646 

ScreenRecording 2019-05-03 at 5 49 05 PM

### DIFF
--- a/Classes/Bookmark/BookmarkViewController.swift
+++ b/Classes/Bookmark/BookmarkViewController.swift
@@ -10,6 +10,10 @@ import Foundation
 import IGListKit
 import Squawk
 
+protocol BookmarkSectionControllerDelegate: AnyObject {
+    func didSwipeToDelete(at indexPath: IndexPath)
+}
+
 protocol BookmarkViewControllerClient {
     func fetch(graphQLIDs: [String], completion: @escaping (Result<[BookmarkModelType]>) -> Void)
 }
--- a/Classes/Bookmark/BookmarkViewController.swift
+++ b/Classes/Bookmark/BookmarkViewController.swift
@@ -23,7 +23,8 @@ final class BookmarkViewController: BaseListViewController<String>,
 BaseListViewControllerDataSource,
 BaseListViewControllerEmptyDataSource,
 BookmarkIDCloudStoreListener,
-BookmarkHeaderSectionControllerDelegate {
+BookmarkHeaderSectionControllerDelegate,
+BookmarkSectionControllerDelegate {
 
     typealias Client = BookmarkViewControllerClient & BookmarkCloudMigratorClient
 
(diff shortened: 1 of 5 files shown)


BasThomas:
This looks awesome, thanks for taking it on @rluftw! 🎉

BasThomas:
And thanks for the kind words!

jdisho:
Should we merge this? @BasThomas

Classes/Bookmark/BookmarkViewController.swift
```diff
@@ -194,4 +199,10 @@ BookmarkHeaderSectionControllerDelegate {
         cloudStore.clear()
     }
 
+    // MARK: BookmarkSectionControllerDelegate
+
+    func didSwipeToDelete(at indexPath: IndexPath) {
+        let graphQLID = cloudStore.ids[indexPath.section-1]
```
BasThomas: Hm, is this guaranteed to always be the correct `id`? I'm a bit worried what would happen during a merge or any other change to the store.

Could we undo the delete when the removal would fail? cc @heshamsalman

BasThomas:
Sorry @jdisho, lost track of this one after I marked it as awaiting review :)

Classes/Bookmark/BookmarkViewController.swift
heshamsalman: @BasThomas 🤔 

Initial thought is that there could be edge-cases here where we end up deleting the wrong bookmark, but it's _probably_ fine. 

My only concern is the scenario where a sync w changes triggered by another devices occurs _just before_ this method fires.

rluftw: @BasThomas While playing around with this, I observed that the selected `id` was consistently correct. I'm not familiar with the migration flow, but what I see is during a merge, the new `id`'s are appended onto the `storage` which I don't believe would effect the deletion unless the deleted `id` was the same? 🤔 

@heshamsalman That's a great point. I'm not sure how this would be remedied? maybe there's some state we can check?

BasThomas: What about verifying with something other than the `id`? Not sure if that's trivial / worth it.

Otherwise we'll ship it like this for now, your call @heshamsalman 😊

rluftw: I can display an alert that shows the repo/issue name for the user to confirm? The tradeoff here is that the user would need to go through an additional step to delete. Not sure how that'll be UX wise?

Here's a gif of that change:
ScreenRecording 2019-05-31 at 10 25 01 AM

heshamsalman: The UX on that isn't ideal, but I think there could be a case made for it. 
I'm for merging this as-is; the likelihood of a storage conflict feels very low to me. 

Thoughts @BasThomas?

BasThomas: Well the confirmation doesn't really do anything to prevent the potential merge conflicts, does it?

BasThomas:
Thanks so much @rluftw! 🎉

PR #2762, opened 2019-05-27
Fix logout crash

nambatee:
This is a fix for https://github.com/GitHawkApp/GitHawk/issues/2743 and https://github.com/GitHawkApp/GitHawk/issues/2643.

Reading through the issues, It looks like the crash only occurred on iPads and that's where I was able to reproduce it. It's the notorious `popoverPresentationController` that's always too easy to miss. 🧐


BasThomas:
Oooh, nice catch! This API drives me crazy. 

PR #2766, opened 2019-06-01
Fix pushing Issues VC when user taps on username in referenced issues

### LINKED ISSUE #2260: Name and underlying button getting called
This is a bug where if you click on someone’s name who referenced an issue/PR.. (Huddie referenced ...) the result is:
1) Your redirected to that users GitHub Profile (Proper outcome)
2) GitHawk segues to the issue/PR which was referenced  (side effect)

How to reproduce:
- Head to #2237
- Scroll to the bottom where it reads ashleyng referenced n hours ago
- Tap ashleyng (redirected)
- Tap the back arrow (your in #2255 now)

nambatee:
This fixes #2260. 🤞 

When user taps `IssueReferencedCell`, 2 competing delegates can be notified - 

1. `UICollectionViewDelegate`
2. `MarkdownStyledTextViewDelegate`

When both delegates are notified (for example when user taps username in a referenced issue), `UICollectionViewDelegate` (`IssueReferencedSectionController`) will push a new `IssuesViewController` :

https://github.com/GitHawkApp/GitHawk/blob/f28b119009e96a1a6dd53ae0e90b7beefc58833c/Classes/Issues/Referenced/IssueReferencedSectionController.swift#L39

and `MarkdownStyledTextViewDelegate` (`IssuesViewController`) will present a profile via `SFSafariViewController` :

https://github.com/GitHawkApp/GitHawk/blob/80d1255401f9d5b53472dd34a3543d3acdf6da1c/Classes/Utility/UIViewController%2BRouting.swift#L21

Since `IssueReferencedSectionController` implements `UICollectionViewDelegate`, I made it conform to the competing delegate (`MarkdownStyledTextViewDelegate`) so that we can figure out which delegate should proceed - if we're notified that username was tapped, we'll present the profile VC and set a flag that will prevent the other delegate from pushing the Issues VC.

Let me know if there's a better way to do this. 🤔

### DIFF
--- a/Classes/Issues/Referenced/IssueReferencedModel.swift
+++ b/Classes/Issues/Referenced/IssueReferencedModel.swift
@@ -53,12 +53,6 @@ final class IssueReferencedModel: ListDiffable {
         self.title = title
         self.actor = actor
 
-        let issueDetailsModel = IssueDetailsModel(
-        owner: actor,
-        repo: repo,
-        number: number
-        )
-
         let builder = StyledTextBuilder(styledText: StyledText(
             style: Styles.Text.secondary.with(foreground: Styles.Colors.Gray.medium.color)
         ))
@@ -75,9 +69,9 @@ final class IssueReferencedModel: ListDiffable {
             .restore()
             .add(text: "\n")
             .save()
-            .add(styledText: StyledText(text: title, style: Styles.Text.secondaryBold.with(attributes: [MarkdownAttribute.issue: issueDetailsModel]) ))
+            .add(styledText: StyledText(text: title, style: Styles.Text.secondaryBold))
             .restore()
-            .add(text: " #\(number)", attributes: [MarkdownAttribute.issue: issueDetailsModel])
+            .add(text: " #\(number)")
 
         self.string = StyledTextRenderer(
             string: builder.build(),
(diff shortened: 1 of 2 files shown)


nambatee:
Good point @jdisho 🤔 Maybe we could use the result of `handle(attribute: DetectedMarkdownAttribute) -> Bool` to set `shouldShowIssueOnItemSelection`? That handles all the cases we care about - `url`, `email`, `username` and `commit` :

https://github.com/GitHawkApp/GitHawk/blob/80d1255401f9d5b53472dd34a3543d3acdf6da1c/Classes/Utility/UIViewController%2BRouting.swift#L14

So something like

```
    func didTap(cell: MarkdownStyledTextView, attribute: DetectedMarkdownAttribute) {
        guard let viewController = viewController else {
            return
        }
        shouldShowIssueOnItemSelection = !viewController.handle(attribute: attribute)
    }
```

If the attribute can be handled, `handle` will do the work & return `true`, if not, we'll allow `didSelectItem` to proceed

jdisho:
Nice one @nambatee! 
I think we can continue with this suggestion 👍

nambatee:
Thanks @jdisho! 😄  Pushed it in b79d0aa

jdisho:
Awesome! Thank you @nambatee

Issue #2718, opened 2019-04-05
Show issue/pr count when viewing a repo


tooomm:
**Is your feature request related to a problem? Please describe.**
You have no idea how many issue's and pr's are listed in each category or get an idea how active, maintained, big, old... a repo is.

**Describe the solution you'd like**
Show the count of issue's/pr's in the top bar when viewing a repo:
`   Overview   |   Issues (132)  |   Pull Requests (2)   |   Code   `

LamourBt:
I'll give this one a try.

LamourBt:
Here is the current state of that feature

<img width="369" alt="Screen Shot 2019-06-03 at 10 42 31 AM" src="https://user-images.githubusercontent.com/14245117/58810643-579a0b00-85ec-11e9-9b67-361c6f0ac760.png">

As you can see the title size should be changed here. 

<img width="373" alt="Screen Shot 2019-06-03 at 10 40 24 AM" src="https://user-images.githubusercontent.com/14245117/58810550-2588a900-85ec-11e9-8545-1b4345a304e6.png">

Previously the title's font was set to 18, changing it to 10 made it worked.

we can bump up the font size if we handle value greater than 1000 like 1K+, 1000000 like 1M+. etc...

Any ideas, thoughts?

Thank you.

jdisho:
Thank you @LamourBt, but it seems too small for me 😕 
Is't is possible to increase the width based on the text size?

LamourBt:
@jdisho just solved it 
<img width="372" alt="Screen Shot 2019-06-03 at 4 05 01 PM" src="https://user-images.githubusercontent.com/14245117/58830910-677c1400-8619-11e9-9609-1fd3596986b9.png">

jdisho:
Awesome! Feel free to create a PR 😊 @LamourBt 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jhjohn123:
Stop can’t help u anymore not a thief tried to do good not helping u guys
do bad things I a good guy that works hard

PR #2767, opened 2019-06-03
Feature/show issues and pr count

LamourBt:
This feature is based on this ticket #2718

### DIFF
--- a/Classes/Repository/RepositoryClient.swift
+++ b/Classes/Repository/RepositoryClient.swift
@@ -15,6 +15,8 @@ protocol RepositoryQuery {
     // generated queries should share the same init
     func summaryTypes(from data: GraphQLSelectionSet) -> [RepositoryIssueSummaryType]
     func nextPageToken(from data: GraphQLSelectionSet) -> String?
+    func numberOfItems(from data: GraphQLSelectionSet) -> Int
+
 }
 
 extension RepoSearchPagesQuery: RepositoryQuery {
@@ -28,6 +30,12 @@ extension RepoSearchPagesQuery: RepositoryQuery {
         guard let results = data as? RepoSearchPagesQuery.Data,
             results.search.pageInfo.hasNextPage else { return nil }
         return results.search.pageInfo.endCursor
+    }
+    
+    func numberOfItems(from data: GraphQLSelectionSet) -> Int {
+        guard let data = data as? RepoSearchPagesQuery.Data else { return 0 }
+        // issueCount is used for both issues and pull requests.
+        return data.search.issueCount
     }
 
 }
@@ -88,7 +96,7 @@ func createSummaryModel(
     data: GraphQLSelectionSet,
     contentSizeCategory: UIContentSizeCategory,
     containerWidth: CGFloat
-    ) -> (models: [RepositoryIssueSummaryModel], nextPage: String?) {
+    ) -> (models: [RepositoryIssueSummaryModel], nextPage: String?, numberOfItems: Int) {
     let nextPage = query.nextPageToken(from: data)
     let models = query.summaryTypes(from: data).compactMap { node in
         return createSummaryModel(
@@ -97,7 +105,8 @@ func createSummaryModel(
             containerWidth: containerWidth
         )
     }.sorted {$0.created > $1.created }
-    return (models, nextPage)
+    let numberOfItems = query.numberOfItems(from: data)
+    return (models, nextPage, numberOfItems)
 }
 
 final class RepositoryClient {
@@ -116,6 +125,7 @@ final class RepositoryClient {
     struct RepositoryPayload {
         let models: [RepositoryIssueSummaryModel]
         let nextPage: String?
+        let numberOfItems: Int
     }
 
     private func loadPage<T: GraphQLQuery>(
@@ -141,7 +151,8 @@ final class RepositoryClient {
                     DispatchQueue.main.async {
                         completion(.success(RepositoryPayload(
                             models: summary.models,
-                            nextPage: summary.nextPage
+                            nextPage: summary.nextPage,
+                            numberOfItems: summary.numberOfItems
                         )))
                     }
                 }
--- a/Classes/Repository/RepositoryClient.swift
+++ b/Classes/Repository/RepositoryClient.swift
@@ -15,7 +15,7 @@ protocol RepositoryQuery {
     // generated queries should share the same init
     func summaryTypes(from data: GraphQLSelectionSet) -> [RepositoryIssueSummaryType]
     func nextPageToken(from data: GraphQLSelectionSet) -> String?
-    func numberOfItems(from data: GraphQLSelectionSet) -> Int
+    func numberOfItems(from data: GraphQLSelectionSet) -> Int?
 
 }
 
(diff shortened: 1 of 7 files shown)


Classes/Repository/RepositoryIssuesViewController.swift
```diff
@@ -94,6 +99,19 @@ IndicatorInfoProvider {
             }
         }
     }
+    func composeIndicatorTitle(with itemCount: Int) -> String? {
+        var newTitle: String
+        switch type {
+        case .issues: newTitle = "Issues (\(itemCount))"
+        case .pullRequests: newTitle = "Pull Requests (\(itemCount))"
+        }
+        guard itemCount > 0 else { return nil }
```
jdisho: I think it makes more sense for to `guard` agains `itemCount` in top of the function body.

Classes/Repository/RepositoryIssuesViewController.swift
```diff
@@ -94,6 +99,19 @@ IndicatorInfoProvider {
             }
         }
     }
+    func composeIndicatorTitle(with itemCount: Int) -> String? {
+        var newTitle: String
+        switch type {
+        case .issues: newTitle = "Issues (\(itemCount))"
```
jdisho: Shouldn't "Issues" and "Pull Requests" be localised?

Classes/Repository/RepositoryClient.swift
```diff
@@ -29,6 +31,12 @@ extension RepoSearchPagesQuery: RepositoryQuery {
             results.search.pageInfo.hasNextPage else { return nil }
         return results.search.pageInfo.endCursor
     }
+    
+    func numberOfItems(from data: GraphQLSelectionSet) -> Int {
+        guard let data = data as? RepoSearchPagesQuery.Data else { return 0 }
```
jdisho: I think it is better to `return nil` rather then `0`. If data is `nil`, doesn't necessary mean that the number of items is 0 and it wouldn't be accurate to show 0.

Classes/Repository/RepositoryIssuesViewController.swift
LamourBt: I did localize the output from `composeIndicatorTitle` method before assigning it to the title  
  ```self?.title = NSLocalizedString(updatedTitle, comment: "")```

Classes/Repository/RepositoryIssuesViewController.swift
```diff
@@ -85,6 +85,12 @@ IndicatorInfoProvider {
             case .error:
                 self?.error(animated: trueUnlessReduceMotionEnabled)
             case .success(let payload):
+                if let itemCount = payload.numberOfItems,
```
BasThomas: It might be worth creating a `strongSelf` here (`guard let strongSelf = self else { return }`) so we don't have to deal with the optional `self`. What do you think?

Classes/Repository/RepositoryIssuesViewController.swift
```diff
@@ -94,6 +100,18 @@ IndicatorInfoProvider {
             }
         }
     }
+    func composeIndicatorTitle(with itemCount: Int) -> String {
+        var newTitle: String
```
BasThomas: This can be a constant (`let`), as it is inferred to be initialized only once. :)

Classes/Repository/RepositoryIssuesViewController.swift
```diff
@@ -94,6 +100,18 @@ IndicatorInfoProvider {
             }
         }
     }
```
BasThomas: Nit: new line (can we add this to Swiftlint?)

Classes/Repository/RepositoryIssuesViewController.swift
```diff
@@ -94,6 +100,18 @@ IndicatorInfoProvider {
...
+    func composeIndicatorTitle(with itemCount: Int) -> String {
+        var newTitle: String
+        switch type {
+        case .issues: newTitle = "Issues (\(itemCount))"
+        case .pullRequests: newTitle = "Pull Requests (\(itemCount))"
+        }
+        return newTitle
+    }
```
BasThomas: And another newline here ☺️

package-lock.json
```diff
@@ -2,90 +2,36 @@
   "requires": true,
```
BasThomas: Did you intend to update some dependencies here?

Classes/Repository/RepositoryIssuesViewController.swift
BasThomas: I think it would be safer and give a better overview if we'd do so here, instead of where it is used.

Classes/Repository/RepositoryIssuesViewController.swift
```diff
@@ -94,6 +100,18 @@ IndicatorInfoProvider {
             }
         }
     }
+    func composeIndicatorTitle(with itemCount: Int) -> String {
```
BasThomas: I'm wondering if we should consider moving this out of the `ViewController`'s logic... and write a test for it? 😇

Classes/Repository/RepositoryIssuesViewController.swift
LamourBt: I was following the linter in that one lol

package-lock.json
LamourBt: I didn't, I just follow the setup.md

Classes/Repository/RepositoryIssuesViewController.swift
jdisho: Why not just `guard let self = self else { return }`? 😄

Classes/Repository/RepositoryIssuesViewController.swift
jdisho: 👍

Classes/Repository/RepositoryIssuesViewController.swift
BasThomas: 🤔 ok :)

Classes/Repository/RepositoryIssuesViewController.swift
BasThomas: Also possible 😊

Classes/Repository/RepositoryIssuesViewController.swift
```diff
@@ -81,20 +81,40 @@ IndicatorInfoProvider {
...
         ) { [weak self] (result: Result<RepositoryClient.RepositoryPayload>) in
+            guard let `self` = self else { return }
             switch result {
             case .error:
-                self?.error(animated: trueUnlessReduceMotionEnabled)
+                self.error(animated: trueUnlessReduceMotionEnabled)
             case .success(let payload):
+                if let itemCount = payload.numberOfItems,
+                      itemCount > 0 {
+                    self.title = self.composeIndicatorTitle(with: itemCount)
```
BasThomas: Hm, seeing this now which confuses me — this updates the view controller's title, not the pager tab, does it? I can't seem to build the app at the moment, so can't check it myself.

LamourBt: Hmm, yes the indicator is using the viewController title, I used the same pattern used before check the last method within this class. Or maybe I misunderstood your question.

LamourBt: I'm currently using Xcode 10.2 and the app builds normally, am I supposed to use a specific Xcode version?

BasThomas:
Merging; we can address the two comments later if we want to. They are no blocker for this awesomeness!

PR #2770, opened 2019-06-06
Fix #2769

### LINKED ISSUE #2769: Secure vulnerability notification is showing a issue detail view with failure
When we receive a github secure vulnerability issue alert, the cell view handles it like an issue and call the issue detail screen when it taps that. Because of the fact that it isn't a issue, the detail screen shows an error. 
Questions:
It should redirect to our repo the way github web does?
We should have a design for a cell specific for that kind of issue?

**To Reproduce**
1. Receive a github security vulnerability issue on feed.
2. Taps it.
3. See the detail issue fails to load.

**Expected behavior**
When tap the vulnerability issue, we should redirect it to the repo. Not a issue detail.

**Screenshots**
IMG_0732
IMG_0733

**Smartphone (please complete the following information):**
 - Device:  iPhone8]
 - OS: iOS 12
 - Version 1.27.2

LucianoPAlmeida:
Fix #2769
* Show repo when the notification is a security vulnerability type.
* And also since it is not an issue, maybe when should not show the number 

so from:
Screen Shot 2019-06-06 at 00 10 09
to:
Screen Shot 2019-06-06 at 00 12 16


Classes/Notifications/NotificationSectionController.swift
```diff
@@ -81,6 +81,14 @@ final class NotificationSectionController: ListSwiftSectionController<Notificati
 
         BadgeNotifications.clear(for: model)
 
+        if model.type == .securityVulnerability {
```
jdisho: This seems like a good case for a `guard` 😄

LucianoPAlmeida: Sure thing :))

BasThomas:
Awesome, thanks @LucianoPAlmeida (and for reviewing/merging, @jdisho)!

j-f1:
Shouldn’t this open the URL of the vulnerability in a Safari View Controller so the user can view the report itself?

LucianoPAlmeida:
@j-f1 that's a good point 👍 
What do you think @BasThomas @jdisho

PR #2936, opened 2021-02-04
UX Enhancement

gokulnair2001:
## Haptic Feedback
* So while using your app I found it to be really impressive but the feature it lacked the most was in terms of User Experience. I saw in your app haptic feedback was implemented but it was not as proper as required. So I thought why not make it lil bit perfect.
* So I added haptic feedback wherever it was really required, like: the switch control buttons, info button, signup button, token button etc. Being a user I sincerely felt, this was a drawback for your app, because you had the code of haptic feedback already written just it was to be implemented wherever required. 
* So I made the changes according to a user what I felt. 
* Hope so you would like and accept my PR.
* If there are an changes to be done please inform me.
* I got some more really interesting features for GitHawk, as soon as you would accept my PR I can start on that features too.

### DIFF
--- a/Classes/Image Upload/ImageUploadTableViewController.swift
+++ b/Classes/Image Upload/ImageUploadTableViewController.swift
@@ -213,6 +213,7 @@ final class ImageUploadTableViewController: UITableViewController,
             initialPhoto: photo,
             delegate: nil
         ))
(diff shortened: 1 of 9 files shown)


gokulnair2001:
Can I know why all checks are failed?.

PR #2938, opened 2021-03-23
Add UNAuthorizationStatus.ephemeral ios14 case

svoip:


YBHQ:
~~~~

